PR #2, opened 2015-06-18
Added logo to the top of the page and some README tips.

rushtong:


src/static/index.html
```diff
@@ -3,7 +3,7 @@
     <meta charset="utf-8">
     <meta http-equiv="Content-type" content="text/html;charset=utf-8">
     <meta name="viewport" content="initial-scale=1,width=device-width">
-    <title>FireCloud | Braod Institute</title>
+    <title>FireCloud | Broad Institute</title>
```
dmohs: Thank you for that :)

dmohs:
My only suggestion is to reduce the size of the logo image to twice the specified size (you specified a height of 36 pixels, so I'd size it to a height of 72 pixels). This is a reasonable compromise between keeping the download size small and looking okay on high-resolution (retina) screens.

Great comments on Intellij usage.

Great PR. Good to go otherwise.

rushtong:
Will update the image size. This version comes straight from the communications website - I didn't do any re-touching.

PR #3, opened 2015-06-18
Created initial footer component

MattPutnam:
Possibly a bit overengineered, but I was reminded of an annoyance at my previous job--every year there would be a stupid bug report to fix all of the copyright years in about 80 places.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -3,15 +3,20 @@
    [dmohs.react :as react]
    ))
 
+(defn footer []
+  (let [thisyear (.getFullYear (js/Date.))
+        startyear 2015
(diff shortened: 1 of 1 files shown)


rushtong:
Showoff!

rushtong:
:+1: Merging.

dmohs:
Nicely done. One question for you and the team: For unicode characters, I tend to just put the character directly in my file. In this case, I would paste in © from somewhere rather than use the escape sequence ("\u00A9"). Any thoughts for a standard style for us?

rushtong:
In general, I try to use html entities. I probably should have called that out. Not sure if there are clojure syntax conflicts with & - it would be good to have an informed choice.

dmohs:
There are no problems in Clojure, but adding these in React is a pain in the neck. If we need to do this, I'll be happy to provide an example.

PR #5, opened 2015-06-22
Enable hot reloading using Figwheel. Add some scripts for documentation.

dmohs:


MattPutnam:
Hmm, I'm having a minor problem where if I trigger a hot reload while not logged in, the login button disappears, so I have to refresh the page.

dmohs:
I do not believe this is a hot reload issue. I believe the button also disappears if you log-out. We will need to address this, but I don't think we need to for this PR.

PR #6, opened 2015-06-22
Skeleton workspace and methods repository pages

dmohs:
Basic navigation structure and tables. Follows PixelMedia design for Workspaces page. Mocked server calls.

The table is custom-built. If we want to use a third-party table, I suggest we discuss and implement that during the next sprint.


MattPutnam:
:+1: Even if it's not perfect we can fix it up as we go along.

PR #7, opened 2015-06-24
Create release build

dmohs:
This commit creates two scripts: The first compiles in advanced mode so that the entire project is compressed into a single file. The second removes build artifacts that are not necessary for deployed code. README updated.

Task: https://broadinstitute.atlassian.net/browse/DSDEEPB-670

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,6 +14,7 @@ WORKDIR /usr/firecloud-ui
 RUN ./script/dev/build-once.sh
 RUN lein test
 
+RUN ./script/release/build-release.sh
 RUN cp -R /usr/firecloud-ui/target/* /var/www/html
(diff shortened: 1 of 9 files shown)


davidangb:
@dmohs and I talked in person; I have two main points of feedback:
* the dev/build-once.sh and release/build-once.sh scripts contain a lot of shared code; is there a way to refactor the commonality?
* building a release requires two scripts. David Mohs separated these to allow debugging problems in a release (once we remove the build artifacts, it's harder to debug). However, I don't like having to call two scripts. One possible solution is to add another script e.g. release.sh which is the one we call to build a release, and simply delegates to the two underlying scripts.

dmohs:
Comments addressed. This is ready for another review.

davidangb:
loving this - looks good!

PR #12, opened 2015-06-29
Created a modal dialog for creating a workspace

MattPutnam:
The dialog is poorly styled and is hard-coded to this page.  I'll leave better styling to more qualified people, and genericizing the dialog is covered in DSDEEPB-629.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
@@ -125,23 +125,45 @@
         [:div {:style {:clear "both"}}]])}))
 
 
+(defn- render-overlay [state refs]
+  [:div {:style {:display (if-not (:overlay-shown? @state) "none")
+                 :position "absolute"
+                 :top 0 :left 0 :width "100%" :height "100%"
+                 :background "rgba(82, 129, 197, 0.4)" :zIndex "99999"
+                 :textAlign "center"}}
+   [:div {:style {:width 300 :margin "100px auto"
+                  :backgroundColor "white" :border "1px solid black"
+                  :padding 15
+                  :textAlign "center"}}
+    [:div {} "Enter a name for the new Workspace:"]
+    [:div {:style {:padding "10px"}} [:input {:ref "wsName" :type "text"}]]
+    (let [clear-overlay (fn []
+                          (set! (.-value (.getDOMNode (@refs "wsName"))) "")
+                          (swap! state assoc :overlay-shown? false))]
+      [:div {}
+       [:button {:onClick #(let [n (.-value (.getDOMNode (@refs "wsName")))]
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -129,19 +129,38 @@
   {:render
    (fn [{:keys [state refs]}]
      [:div {}
+      ;; Begin overlay
```
dmohs: I would prefer to avoid this style of comments. I think the distinction is valuable, so my preference would be to pull this out into a function, e.g.:
```
(defn render-overlay [state refs]
  [:div {:style {:visibility (if-not ...
  ...)
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -129,19 +129,38 @@
...
+      [:div {:style {:visibility (if-not (:overlay-shown? @state) "hidden")
```
dmohs: Any reason for `{:visibility "hidden"}` instead of `{:display "none"}`? I generally prefer the latter unless the former is required.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -129,19 +129,38 @@
...
+                     :position "absolute"
+                     :top "0px" :left "0px" :width "100%" :height "100%"
```
dmohs: Items like `:top "0px"` can be shortened to `:top 0`. Most pixel values can be expressed this way. It's just a documented feature of React. I suggest we follow this to ensure our code is as compact as possible (which increases the amount of context that is visible on a single screen).

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -129,19 +129,38 @@
...
+                     :text-align "center"}}
+       [:div {:style {:width "300px" :margin "100px auto"
+                      :background-color "white" :border "1px solid black"
+                      :padding "15px"
+                      :text-align "center"}}
+        [:div {} "Enter a name for the new Workspace:"]
+        [:div {:style {:padding "10px"}} [:input {:ref "wsName" :type "text"}]]
+        (let [clear-overlay (fn []
+                              (swap! state assoc :overlay-shown? false)
+                              (set! (.-value (.getDOMNode (@refs "wsName"))) ""))]
```
dmohs: If you decide to go with `:display "none"` above, you'll need to move this statement up one line so you're not setting the value of a DOM node that is no longer present in the document.

dmohs:
Some very minor comments. Overall an excellent PR. Nicely done.

PR #15, opened 2015-07-07
Filter implementation for the methods list

MattPutnam:
Not necessarily 100% done, but it works for the mock data currently present.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
@@ -37,6 +37,13 @@
                        (:methods props))})])])})
 
 
+(defn- contains-text [text fields]
+  (fn [method]
+    (some #(not= -1 (.indexOf (method %) text)) fields)))
+
+(defn- filter-methods [methods fields text]
+  (filter (contains-text text fields) methods))
+
 (defn- create-mock-methods []
   (map
     (fn [i]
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
```diff
@@ -49,12 +53,21 @@
...
   {:render
-   (fn [{:keys [state]}]
+   (fn [{:keys [state refs]}]
      [:div {:style {:padding "1em"}}
       [:h2 {} "Method Repository"]
       [:div {}
        (cond
-         (:methods-loaded? @state) [MethodsList {:methods (:methods @state)}]
+         (:methods-loaded? @state) [:div {}
+                                    (let [do-filter #(swap! state assoc :filtered-methods
```
dmohs: I suggest a slight restructuring here. Define a function `filter-methods` that takes a list of methods and a filter value and returns the filtered list. Then, you would create the `MethodsList` component like
```
[MethodsList {:methods (filter-methods (:methods @state) (:filter-text @state))}]
```
In the places you call `do-filter`, instead do `(swap! state assoc :filter-text (.-value (.getDOMNode (@refs "input"))))`

src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
```diff
@@ -49,12 +53,21 @@
...
+                                                       (filter (contains-text (.-value (.getDOMNode (@refs "input")))
+                                                                 '("namespace" "name" "synopsis")) (:methods @state)))]
+                                      [:div {:style {:paddingBottom "1em" :paddingLeft "4em"}}
+                                       [:input {:style common/input-text-style :type "text" :ref "input" :placeholder "Filter"
```
dmohs: Nit: line length

dmohs:
Minor comments. One suggested refactoring then I think it's good to go.

PR #16, opened 2015-07-08
Adding keyboard control to the Create Workspace dialog

MattPutnam:
* Made common/Button an <a> instead of a <div> with an onClick.  This automatically makes it part of the focus cycle, including styling to highlight when focused.
* Added onKeyDown functions to make common/Button and the Create Workspace 'cancel' button respond to [space] or [enter] when focused.  TODO: a utility function to reduce repetition
* ESC key functions the same as 'cancel'.  However, the focus must be on one of the dialog's components for this to work; another TODO is to focus the workspace name field when the dialog is shown, which would reduce this problem.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -34,16 +34,19 @@
 (react/defc Button
   {:get-default-props
    (fn []
-     {:color :blue})
+     {:color (:button-blue colors)})
    :render
    (fn [{:keys [props]}]
-     [:span {:style {:display "inline-block"
-                     :backgroundColor (case (:color props)
-                                        (:button-blue colors))
-                     :color "white" :fontWeight 500
-                     :borderRadius 2 :padding "0.7em 1em"
-                     :cursor "pointer"}
-             :onClick (fn [e] ((:onClick props) e))}
+     [:a {:style {:display "inline-block"
+                  :backgroundColor (:color props)
+                  :color "white" :fontWeight 500
+                  :borderRadius 2 :padding "0.7em 1em"
+                  :textDecoration "none"}
+          :href "javascript:;"
+          :onClick (fn [e] ((:onClick props) e))
+          :onKeyDown (fn [e]
+                       (let [k (.-keyCode e)]
+                         (when (or (= 32 k) (= 13 k)) ((:onClick props) e))))}
       (:text props)
       (when (= (:style props) :add)
         [:span {:style {:display "inline-block" :height "1em" :width "1em" :marginLeft "1em"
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -37,13 +37,17 @@
...
    (fn [{:keys [props]}]
-     [:span {:style {:display "inline-block"
-                     :backgroundColor (case (:color props)
-                                        (:button-blue colors))
-                     :color "white" :fontWeight 500
-                     :borderRadius 2 :padding "0.7em 1em"
-                     :cursor "pointer"}
-             :onClick (fn [e] ((:onClick props) e))}
+     [:a {:style {:display "inline-block"
+                  :backgroundColor (case (:color props)
```
dmohs: I see you only reformatted this, but I think this `case` should be a `when`.

src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -37,13 +37,17 @@
...
+                                     (:button-blue colors))
+                  :color "white" :fontWeight 500
+                  :borderRadius 2 :padding "0.7em 1em"
+                  :textDecoration "none"}
+          :href "javascript:;"
```
dmohs: Just FYI: I pretty much add this {:href "javascript:;"} boilerplate to every single :a element in my own codebases. When you're thinking about factoring out common functionality, you might consider this case.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -153,36 +153,42 @@
...
-                           (set! (.-value (.getDOMNode (@refs "wsDesc"))) "")
-                           (set! (.-value (.getDOMNode (@refs "shareWith"))) "")
-                           (swap! state assoc :overlay-shown? false))]
+  (let [clear-overlay (fn []
+                        (set! (.-value (.getDOMNode (@refs "wsName"))) "")
+                        (set! (.-value (.getDOMNode (@refs "wsDesc"))) "")
+                        (set! (.-value (.getDOMNode (@refs "shareWith"))) "")
+                        (swap! state assoc :overlay-shown? false))]
+    [:div {:style (modal-background state)
+           :onKeyDown (fn [e] (when (= 27 (.-keyCode e)) (clear-overlay)))}
```
dmohs: Consider something like this for clarity (in every case where you do this comparison):
```
(when (= (:esc common/key-codes) (.-keyCode e))
  ...
```

dmohs:
LGTM (looks good to me). A couple of minor suggestions.

PR #17, opened 2015-07-10
addeed some commentary based on a code review

eddiebroad:
during the review, if you see "???" or if a comment seems vague or altogether incorrect please let me know?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,7 +1,11 @@
+;; define a root namesapce 'org.broadinstitute'
 (def root-ns "org.broadinstitute")
+
+;; set target as the relative build directory
 (def build-dir-relative "target")
 
 
+;; defproject settings for lein
 (defproject org.broadinstitute/firecloud-ui "0.0.1"
   :dependencies
   [
--- a/script/common/build.sh
+++ b/script/common/build.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 set -euo pipefail
 IFS=$'\n\t'
 
--- a/script/dev/build-once.sh
+++ b/script/dev/build-once.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -x
 set -euo pipefail
 IFS=$'\n\t'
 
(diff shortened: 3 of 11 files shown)


dmohs:
Most of the comments in this PR describe what the code is doing. In general, the code should describe itself, while comments should be used for describing why the approach was taken, in cases where that is necessary. I would like to see all of the "what" comments removed. Otherwise, these will create a maintenance burden as we will have to remember to update the comment whenever the code changes.

For certain utility functions, it might make sense to give it a doc string (as you have done). In that case, it should look like this:
```
(defn function-name
  "Doc string goes here
and can be multi-line."
  [args]
  (do-stuff))
```
These should be used sparingly, since most functions should be self-describing. If that isn't the case, it is better to refactor/rename the function rather than fix the problem with a doc string.

There are a few places where you have added vertical whitespace to the files. We want to avoid this. In general, we want a developer to see as much context as possible on a single screen, so we prefer less vertical whitespace as long as it is easy to differentiate things visually.

rushtong:
I agree with David, for the most part, except to note that having as much context as possible visible isn't ideal for me. The *right* context needs to be visible, and having too much context impedes my understanding quite a bit. IDEs are great for this in that they can collapse irrelevant code. That being said, more than 2 lines of white space is just redundant.

rushtong:
@eddiebroad - What is the status of this? Do you wish to keep this PR open and rework things or can it be closed?

eddiebroad:
I want to delete this now

On Tue, Jul 21, 2015 at 7:41 AM, Greg Rushton <notifications@github.com>
wrote:

> @eddiebroad <https://github.com/eddiebroad> - What is the status of this?
> Do you wish to keep this PR open and rework things or can it be closed?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/broadinstitute/firecloud-ui/pull/17#issuecomment-123277764>
> .
>



-- 
Edward A. Salinas
Senior Software Engineer
Broad Institute of MIT and Harvard
75 Ames Street, Rm #4020
Cambridge, MA 02142
Ph: 617-714-7905

PR #23, opened 2015-07-14
Updates to the tab bar

MattPutnam:
The tab bar now takes a component for each tab and automatically switches to it.  There is also an optional 'onTabSelected' function that can be used to do any updating needed.  This will be used to set the filters for the main workspace list, and select between views for the workspace details page.


dmohs:
Looks great. Minor comments as discussed in-person.

PR #24, opened 2015-07-14
Initial icons using Font Icons.

dmohs:


dmohs:
Matt said this looks okay, so merging.

PR #25, opened 2015-07-15
Add icons and colors for workspace status.

MattPutnam:
Live data is amended to add status=Complete field, fake data randomly selects statuses.

There is some screwiness with the icon component layout, someone can fix this later.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -1,4 +1,6 @@
-(ns org.broadinstitute.firecloud-ui.common)
+(ns org.broadinstitute.firecloud-ui.common
+  (:require
+    [org.broadinstitute.firecloud-ui.utils :as utils]))
 
 
 (def keymap
@@ -14,3 +16,8 @@
 (defn clear! [refs & ids]
   (doseq [id ids]
     (set! (.-value (.getDOMNode (@refs id))) "")))
+
+(defn center [props & children]
+  [:div (utils/deep-merge props {:style {:position "absolute" :top "50%" :left "50%"
+                                         :transform "translate(-50%, -50%)"}})
+   children])
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -63,7 +63,7 @@
                                    :backgroundColor "white"}}])])})]
     {:get-initial-state
      (fn [{:keys [props]}]
-       {:active-tab 0})
+       {:active-tab-index 0})
      :render
      (fn [{:keys [props state]}]
        [:div {}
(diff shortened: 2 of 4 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -14,3 +16,8 @@
 (defn clear! [refs & ids]
   (doseq [id ids]
     (set! (.-value (.getDOMNode (@refs id))) "")))
+
+(defn center [props child]
```
dmohs: Since this is a general utility, you may want to make it handle arbitrary children, e.g. `(defn center [props & children] ...`

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -80,4 +80,51 @@
                               (when-let [f (:onTabSelected tab)] (f e)))}])
            (:items props))
          [:div {:style {:clear "both"}}]]
-        [:div {} (:component (nth (:items props) (:active-tab @state)))]])}))
+        [:div {} (:component (nth (:items props) (or (:active-tab @state) 0)))]])}))
```
dmohs: Nit: `:active-tab-index` might be clearer. Also, it seems suspicious that you have to check this for `nil` given that it has an initial state set to 0.

MattPutnam: Ah, the "or 0" thing was because the state was getting screwed up on hot reloading.  I'll take it out.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -217,16 +222,18 @@
...
-    [:div {:style {:borderRadius 5 :backgroundColor (:success-green style/colors) :color "#fff"
-                   :fontSize "125%" :fontWeight 400 :padding 25 :textAlign "center"}}
-     [:span {:style {:display "inline-block" :verticalAlign "middle"
-                     :width "2.5em" :height "2.5em" :position "relative"
-                     :backgroundColor "#fff" :borderRadius "100%"}}
-      [:span {:style {:fontFamily "fontIcons" :color (:success-green style/colors) :fontSize "90%"
-                      :position "absolute"
-                      :top "50%" :left "50%" :transform "translate(-50%, -50%)"}} ""]]
-     [:span {:style {:marginLeft "1.5ex"}} "Complete"]]
+                  :top 0 :left 0 :width 290 :marginRight 40}}
```
dmohs: Might be worth adding a TODO here for making this a dynamic width so it takes up less space on smaller screens.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -317,7 +326,10 @@
        "/workspaces"
        {:on-done (fn [{:keys [success? xhr]}]
                    (if success?
-                     (let [workspaces (utils/parse-json-string (.-responseText xhr))]
+                     (let [raw-workspaces (utils/parse-json-string (.-responseText xhr))
+                           ;; The following is just to make the icons work with the
+                           ;; current live data, which uses this missing field
+                           workspaces (map #(assoc % "status" "Complete") raw-workspaces)]
```
dmohs: I think this file is getting too long. Maybe a TODO at the top to refactor into multiple files?

dmohs: I think this modification should be conditional on `utils/use-live-data`. Otherwise, I still can't see the nice icons unless I also modify the code here. This is hacky, and I'm not even sure if Rawls will ever return status like this, but let's at least make it easy to expose them for now.

dmohs:
Looks good. Minor comments.

PR #27, opened 2015-07-16
Filter buttons working

MattPutnam:
Filter buttons are togglable, filter based on status, and display the counts of each category

Also shoehorned in a minor fix to scrolling--entering a workspace now scrolls to the top

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
@@ -66,11 +66,19 @@
       (:data props)])})
 
 
+(defn- filter-workspaces [f workspaces]
+  (case f
+    :all workspaces
+    :complete  (filter (fn [ws] (= "Complete"  (ws "status"))) workspaces)
+    :running   (filter (fn [ws] (= "Running"   (ws "status"))) workspaces)
+    :exception (filter (fn [ws] (= "Exception" (ws "status"))) workspaces)))
+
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -66,11 +66,19 @@
       (:data props)])})
 
 
+(defn- filter-workspaces [f workspaces]
+  (case f
+    :all workspaces
+    :complete  (filter (fn [ws] (= "Complete"  (ws "status"))) workspaces)
```
dmohs: I always turn this alignment style off in my IDE. Do you think it adds readability?

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -322,22 +344,23 @@
         (:error-message @state) [:div {:style {:color "red"}}
                                   "FireCloud service returned error: " (:error-message @state)]
         :else [comps/Spinner {:text "Loading workspaces..."}])])
+   :get-initial-state
```
dmohs: I generally add this before :render. I basically put these in the same order they are called in the lifecycle.

dmohs:
Looks good. I rebased on this and we still do need the scroll-to, but I've refactored it a bit in my branch. I'll rebase and add that after we merge this one.

PR #29, opened 2015-07-17
Page to import entities from a JSON file

helgridly:


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -57,6 +58,9 @@
   {:render
    (fn [{:keys [props]}]
      [:div {:style {:textAlign "right"}}
+      [TopNavBarLink {:name "Import Data"
```
helgridly: So if you're adding something to the top nav, you have to do so in three places: here, line 77, and line 96. Can we make a list of "things to put in the top nav" and reference it from these three places instead? 'cos I missed one first time round.

dmohs: That's a good idea. Something like:
```clojure
(def top-nav-bar-items
[{:key :workspaces
  :name "Workspaces"
  :component workspaces/Page}
  ...
]
```
should work. Feel free to go for it or just add a TODO for us to tackle.

helgridly: I'll do it in a separate PR.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -93,3 +93,10 @@
     (fn [x1 x2] (if (and (map? x1) (map? x2)) (deep-merge x1 x2) x2))
     maps))
 
+
+(defn generate-form-data [params]
+  "Create a blob of multipart/form-data from the provided map."
```
dmohs: Annoyingly, the docs for `defn` have the doc string directly after the name and before the parameters. We should probably follow this even though it sucks.

src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
```diff
@@ -0,0 +1,36 @@
...
+  (utils/ajax-orch
+    (str "/workspaces/" workspaceNamespace "/" workspaceName "/importEntitiesJSON")
+    {:method :post
+     :encType "multipart/form-data"
+     :on-done :default
+     :data (utils/generate-form-data {:entities entityFile})}))
+
+
+(react/defc Page
+            {:render
```
dmohs: Nit: This should be a normal (2-space) indent. I'll try to figure out how to make IntelliJ behave.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
helgridly: Yeah, I'd read this but I copied off something in utils for consistency.

src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
```diff
@@ -0,0 +1,36 @@
+(ns org.broadinstitute.firecloud-ui.page.import-data
+  (:require
+    [clojure.string]
+    [dmohs.react :as react]
+    [org.broadinstitute.firecloud-ui.common :as common]
+    [org.broadinstitute.firecloud-ui.common.components :as comps]
+    [org.broadinstitute.firecloud-ui.common.style :as style]
+    [org.broadinstitute.firecloud-ui.common.table :as table]
```
dmohs: We don't have optimize imports. Sorry. No need to clean these up, but do so if you feel like it.

dmohs:
Good to merge as-is or tweak and merge.

src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
dmohs: Preferences -> Editor -> Code Style -> Clojure -> General tab -> "Default to Only Indent" should be checked.

PR #33, opened 2015-07-21
view entities of a single type in the workspace data tab (temp soln)

MatthewBemis:
this is a temporary solution to fulfill [DSDEEPB-746](https://broadinstitute.atlassian.net/browse/DSDEEPB-746). in the future, the ajax call will be done in an area where it can be reloaded based on which workspace is selected and which entity type is selected. this solution solely demonstrates the ability to pull in workspace entities from rawls and display those entities in a formatted table

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
@@ -121,6 +121,33 @@
                           "Me"])
                     filtered-workspaces)})])]))})
 
+(react/defc EntitiesList
+  {:render
+   (fn [{:keys [props]}]
+     [:div {:style {:padding "0 4em"}}
+      (if (zero? (count (:entities props)))
+        [:div {:style {:textAlign "center" :backgroundColor (:background-gray style/colors)
+                       :padding "1em 0" :borderRadius 8}}
+         "No entities to display."]
+        [table/Table
+         (let [cell-style {:flexBasis "8ex" :flexGrow 1 :whiteSpace "nowrap" :overflow "hidden"
+                           :borderLeft (str "1px solid " (:line-gray style/colors))}
+               header-label (fn [text & [padding]]
+                              [:span {:style {:paddingLeft (or padding "1em")}}
+                               [:span {:style {:fontSize "90%"}} text]])]
+           {:columns [{:label (header-label "Entity Type")
+                       :style (merge cell-style {:borderLeft "none"})}
+                      {:label (header-label "Entity Name")
+                       :style cell-style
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -36,6 +36,21 @@
...
+                [:form {:action "http://httpbin.org/post" :encType "multipart/form-data" :method "post"}
+                 [:div {:style {:padding "22px 48px 40px" :boxSizing "inherit" }}
+                  (style/create-form-label "Workspace Name")
+                  (style/create-text-field {:style {:width "33%"} :name "workspaceName"})
+                  (style/create-form-label "Workspace Namespace")
+                  (style/create-text-field {:style {:width "33%"} :name "workspaceNamespace"})[:br]
+                  [:input {:type "file" :name "entities"}]
+                  [:input {:type "submit" :value "Upload"}]]
+                 ]]
+               )})
```
MattPutnam: Nit: indentation for this function, and the close )]} should all go on the end of line 50.  Same in several other places.

src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
```diff
@@ -31,4 +31,4 @@
                       :onClick (fn [e] (submit-entities
                                          (-> (@refs "workspaceName") .getDOMNode .-value)
                                          (-> (@refs "workspaceNamespace") .getDOMNode .-value)
-                                         (-> (@refs "entities") .getDOMNode .-files (aget 0))))}]]])})
+                                         (-> (@refs "entities") .getDOMNode .-files (aget 0))))}]]])})
```
MattPutnam: Let's re-add this endline to avoid cluttering the diff

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces.cljs
```diff
@@ -153,6 +180,43 @@
           [Button {:text (build-text "Exception" :exception) :state state :filter :exception}]
           [:div {:style {:clear "both"}}]]))}))
 
+(react/defc EntityFilterButtons
```
MattPutnam: We should really factor this out into a common component.  I realize we're trying to close out a sprint, so we can do this later.

MattPutnam:
A few minor formatting things, and I think this is good to go.

MatthewBemis:
PR comments addressed. I decided to remove the upload component because it's only being used in one place

MatthewBemis:
PR comments addressed

PR #42, opened 2015-07-23
removed comments and whitespace

eddiebroad:
require more tweaking/viewing is needed thought.


eddiebroad:
at this point, it seems like all that needs to be done is aesthetic review?

PR #45, opened 2015-07-23
Refactoring the filter buttons, and making them work on ws data tab

MattPutnam:


MattPutnam:
This is just refactoring, so nothing should change (except for the filter buttons working on the workspace->data tab).

I assume you're talking about the filter-entities function?  The previous ones didn't do anything, they had just been copy/pasted in and weren't being called..  The new code is actually functional.

PR #49, opened 2015-07-27
Overhaul to my new table.  Much easier to see each column.

MattPutnam:
There are a few unused things (an index in the table component, and a ":resizable? false" in the usage of it), but these are intentional placeholders for future work.


eddiebroad:
I was able to successfully compile the branch and view the corresponding UI and saw no differences.

PR #50, opened 2015-07-27
Column resizing works!

MattPutnam:
There are a few outstanding issues but I'm going to leave this for now

* I'd like there to be a visual cue when you're dragging--the standard gray line showing where the column would be resized to
* There's a bug where if you drag outside the table, release the mouse button, then re-enter, it still thinks you're dragging and will resize when you click.  I can't figure out the right way to cancel the gesture when you go outside the table.


eddiebroad:
nice!  I will review this PR momentarily

eddiebroad:
I build/tested and experienced resizing.  looks good

PR #51, opened 2015-07-27
some UI for viewing method configurations to import

eddiebroad:
* I added a new tab in the workspaces tab for importing method configurations. this perhaps next be moved to appear as a modal div from a button appearing in the other "Method Configurations" tab
*  I added code that retrieves from a checkbox the status of the checkbox
* this is solely in the UI layer, nothing happens in any other layer

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_method_confs_import.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_method_confs_import.cljs
@@ -1,0 +1,78 @@
+(ns org.broadinstitute.firecloud-ui.page.workspace.workspace-method-confs-import
+  (:require
+    [dmohs.react :as react]
+    [org.broadinstitute.firecloud-ui.common.components :as comps]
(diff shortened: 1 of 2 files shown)


eddiebroad:
I accidentally named this with the ID  506..... but everywhere you see here 506, it should instead be 878.

MattPutnam:
As we discussed after the standup, this isn't the UI we want.  We eventually want two places to do this: a modal "dialog" (see the "Create Workspace" button) from the Method Configs tab that shows all available configs to select from, and an ability to add from the Method Repository page.

eddiebroad:
Okay Matt.  I had though to rearrange the GUI elements to appear as modal via the "Method Configurations" tab but wasn't sure if I should do so before this PR or not.  I wasn't sure if it's a good idea or not to issue a PR then follow it with another subtask (involving rearranging the GUI elements) with another PR for it.

eddiebroad:
I have created a local branch and am working to get a modal "Select Method Configuration To Import Into Your Workspace" box  to show up...

PR #55, opened 2015-07-30
Made table column resizing continuous

MattPutnam:
Also made the drag handle a bit bigger.  Still has the issue with dragging off the component.


dmohs:
I could suggest tweaks, but this already works as good or better than any third-party table component I've used. It's excellent. Let's revisit after we have full functionality (or at least method submission/monitoring). I'll let someone else merge, since it sounds like there was discussion around this I didn't hear.

eddiebroad:
nice fluidly-moving columns

PR #56, opened 2015-07-30
Made paginator position configurable, defaulting to below.

MattPutnam:
Also added a minimum column width while I was at it.

The rebase for this branch got kinda screwed up, so please double-check for regressions.


dmohs:
:+1: Looks good. There is a bit of duplicated code around creating the paginator for each table, but since we have time in this sprint for refactoring, and I plan to hit that piece first, I'll see what I can do there.

PR #59, opened 2015-08-04
1) Propagation of workspace-information to import MC screen ; 2) GUI mods for actual import

eddiebroad:
Removed checkboxes from import MC window and replaced them with gear buttons.

These gear buttons ar intended to import a single/given/unique MC at a time
Removed the "Import Selected ..." big button.  The only way to leave the screen is through the X.

Propagate workspace information to the MC import screen

Minor text tweak for TODO alert.
The import button tells us a TODO

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -66,18 +66,18 @@
                                                      (:header-darkgray style/colors)}}
                                         children])
                  cell (fn [children] [:span {:style {:paddingLeft 16}} children])]
-             {:columns [{:header-component [:div {:style {:padding "13px 0 12px 12px"
-                                                          :backgroundColor (:header-darkgray style/colors)}}
-                                            [:input {:type "checkbox" :ref "imcallcheck"}]]
-                         :starting-width 42 :resizable? false
+             {:columns [{:header-component (header "Name") :starting-width 200
                          :cell-renderer (fn [row-num conf]
-                                          [:div {:style {:paddingLeft 12}}
-                                           [:input {:type "checkbox" :ref (str "imc_" row-num)}]])}
-                        {:header-component (header "Name") :starting-width 200
-                         :cell-renderer (fn [row-num conf]
-                                          (cell [:a {:href "javascript:;"
-                                                     :style {:color (:button-blue style/colors)
-                                                             :textDecoration "none"}} (conf "name")]))}
+                                          (cell [:a {:onClick
+                                                     (fn [] (js/alert
+                                                              (str "TODO : process/prompt for "
+                                                                "import of this method configuration "
+                                                                conf " into the workspace →"
+                                                                (get props :workspace) )))
+                                                     :href "javascript:;"
+                                                     :style {:color
+                                                              (:button-blue style/colors) :textDecoration "none"}}
+                                                 (conf "name")]))}
                         {:header-component (header "Namespace") :starting-width 200
                          :cell-renderer (fn [row-num conf] (cell (conf "namespace")))}
                         {:header-component (header "Type") :starting-width 100
@@ -133,7 +133,7 @@
    :width "90%"})
 
 
-(defn render-import-overlay [state]
+(defn render-import-overlay [state  workspace ]
   (let [clear-import-overlay #(swap! state assoc :import-overlay-shown? false)]
     (when (:import-overlay-shown? @state)
       [:div {:style modal-import-background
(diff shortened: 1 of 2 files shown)


MattPutnam:
We don't need the first column at all.  There's already a hyperlink on the name column, just use that.

eddiebroad:
Matt, I thought to replace the checkbox with a button with a gear-icon (it seem to imply "Import this Method Configuration") because I thought the hyperlink might be slightly ambiguous as to what clicking it would do.  If we use the hyperlink for the import action, then I had thought of adding a little instructional text at the top something like "Click a method configuration to import it".  However, by removing the first column, then there are less columns and perhaps the screen might be less "cluttered"  I wonder what you think?

eddiebroad:
I will go ahead and in a moment, refactor to cause the currently "empty" hyperlink to have the functionality as the button.

MattPutnam:
Well, whatever the decision, something needs to happen.  Either use the link, or have a button column and remove the link.  If we want a column with an "add" button, then the gear icon is definitely the wrong icon.  There's a "plus" icon that would make more sense.  And rather than making it a Button, just add it as a text component and make the whole table cell clickable.

eddiebroad:
I had forgot about the "+" icon.  that does make more sense.
I wonder why you prefer a clickable table cell over a clickable button.

MattPutnam:
Hmm, the button might be okay, it just needs to be centered/aligned better.  components/Button doesn't have any fine-grained control right now, so it might be unsuited for this.

eddiebroad:
one thing about using the "empty/javascript" hyperlink is that by using that, then this table has consistency with the other table of listing MCs in a workspace (the table which is the "entry point" to editing an MC).

MattPutnam:
That's not a consistency that we need.  If it's not clickable it shouldn't look like it is.

eddiebroad:
I went the consistency route with the push at 11:25 AM today.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -148,8 +149,11 @@
                        :borderBottom (str "1px solid " (:line-gray style/colors))
                        :padding "20px 48px 18px"}}
          [:div {:style {:fontSize 24 :align "center" :textAlign "center" :paddingBottom "0.5em"}}
-          "Select Method Configurations For Import"]
-         [ImportWorkspaceMethodsConfigurationsList]
+          "Select A Method Configuration For Import"]
+         [ImportWorkspaceMethodsConfigurationsList
```
dmohs: 153-157 all on one line.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -66,18 +66,19 @@
...
-                                                             :textDecoration "none"}} (conf "name")]))}
+                                          (cell [:a {:onClick
+                                                     (fn [] (js/alert
+                                                              (str "TODO : process/prompt for "
+                                                                "import of this method configuration "
+                                                                conf " into the workspace →"
+                                                                (get props :workspace) )))
+                                                     :href "javascript:;"
+                                                     :style {:color
+                                                              (:button-blue style/colors)
```
dmohs: Merge with previous line.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -148,8 +149,11 @@
...
+          "Select A Method Configuration For Import"]
+         [ImportWorkspaceMethodsConfigurationsList
+          {
+           :workspace workspace
+           }
+          ]
          [:div {:style {:paddingTop "0.5em"}}
-          [comps/Button {:style :add :text "Add selected to workspace"
-                         :onClick #(swap! state assoc :import-overlay-shown? false)}]]]]])))
+          ]]]])))
```
dmohs: Merge with previous line.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
eddiebroad: @dmohs  Regarding this note "Merge with previous line." about line 80, I specifically added the newline in line 79 after "(:button-blue style/colors)" so that the line wouldn't break the
rule "When feasible, avoid making lines longer than 100 characters."  (  from here https://broadinstitute.atlassian.net/wiki/display/DSDE/Code+Style+Conventions ) .  I did so because In my judgement and determination I considered the insertion of the line to be "feasible"

eddiebroad:
I pushed a commit at 1:47 PM removing whitespace coupled with merging lines

eddiebroad:
David assigned this to me 3 or so hours ago and I addressed the whitespace concerns so I  am going to complete this PR and merge.

eddiebroad:
"Close and Comment" doesn't mean "Close, merge, and comment!"

PR #60, opened 2015-08-04
added entitylist call per workspace / supported TSV uploading from UI

MatthewBemis:
the important part of this PR is the support for TSV uploading, but I also removed the hard-coded listentities call so it's now per workspace which will be nice to have for the blueteam demo

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
@@ -22,22 +22,22 @@
 
 (defn- submit-entities [state workspaceName workspaceNamespace entityFile]
   (utils/ajax-orch
-    (str "/workspaces/" workspaceNamespace "/" workspaceName "/importEntitiesJSON")
+    (str "/workspaces/" workspaceNamespace "/" workspaceName "/importEntities")
     {:method :post
      :encType "multipart/form-data"
-     :on-done #(swap! state assoc :entities-loaded? true :uploadResults (utils/parse-json-string (-> (:xhr %) .-responseText)))
+     :on-done #(swap! state assoc :entities-loaded? true :message (-> (:xhr %) .-responseText))
      :data (utils/generate-form-data {:entities entityFile})}))
 
 
 (react/defc Page
   {:render
    (fn [{:keys [state refs]}]
-     [:div {}
-      [:div {:style {:padding "22px 48px 40px" :boxSizing "inherit" }}
-       (style/create-form-label "Workspace Name")
-       (style/create-text-field {:style {:width "33%"} :name "workspaceName" :ref "workspaceName"})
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/import_data.cljs
```diff
@@ -47,29 +47,8 @@
...
-                          [(entity "entityName")
-                           (entity "entityType")
-                           (str (entity "succeeded"))
-                           (entity "message")])
-                     (:uploadResults @state))})]])])})
+        [:div {:style {:paddingTop "22px"}}
+         [:div {:style {:fontSize "125%" :fontWeight 500}} "Import Results" [:br]
+          [:div {:style {:fontSize "75%" :fontWeight 100}} (:message @state)]
+
+          ]])])})
```
MattPutnam: Nitpick: merge lines 52-54

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_data.cljs
```diff
@@ -13,6 +13,14 @@
     :aliquot (filter (fn [entity] (= "aliquot" (entity "entityType"))) entities)
     entities))
 
+(defn- create-mock-entities []
+  (map
+    (fn [i]
+      {:entityType (rand-nth ["sample"])
```
MattPutnam: rand-nth with one item is kinda silly.  Are there other types that we can throw in for mock?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_data.cljs
```diff
@@ -57,8 +65,25 @@
...
                                         (make-button "All" :all)]}]
-        [:div {:style {:padding "2em 0"}} [EntitiesList {:entities (:entities props)
-                                                         :active-filter (:active-filter @state)}]]]))})
 
-(defn render-workspace-data [entities]
-  [WorkspaceData {:entities entities}])
+        [:div {:style {:padding "2em 0"}}
+         (cond
+           (:entities-loaded? @state) [EntitiesList {:entities (:entities @state)
+                                                                               :parent-state state}]
```
MattPutnam: Maybe there was a bad merge here at some point?  EntitiesList no longer takes a parent-state property, and should take {:active-filter (:active-filter @state)} instead.

MattPutnam:
Just a few small things, really.  Otherwise :+1:

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_data.cljs
MatthewBemis: awesome, that explains why the filtering broke! thanks for catching that

PR #65, opened 2015-08-06
Use new table for method repo.

dmohs:


rushtong:
:+1: Love the removal of so much inline styling.

PR #67, opened 2015-08-07
Use new table for config import.

dmohs:


MattPutnam:
We lost column resizing.  Other than that :+1:

PR #68, opened 2015-08-07
Added column sorting to the new table

MattPutnam:
Sorting demonstrated in the workspace method configs tab.

A standing concern is that the assorted render functions are getting too complicated.


eddiebroad:
Nice!

dmohs:
:+1: Agree with your concerns. We will continue to refactor.

PR #69, opened 2015-08-07
Fix hot-reloading after uncaught exception.

dmohs:
This is fixed in v0.2.6 of the wrapper, so just bumping the version.


eddiebroad:
thanks for the upgrade.  i approve of this commit/merge after compiling and testing!

PR #71, opened 2015-08-07
Track window scrolling to keep edit/save/cancel buttons always visible

MattPutnam:


eddiebroad:
I compiled and tested this GUI and as I added pre-requisites the "Save" and "Cancel Editing" buttons did not become invisible as I scrolled down.  Nice!

eddiebroad:
If you ask me I have no objection to the merging.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -168,7 +171,17 @@
...
-      :editing? false})
+      :editing? false
+      :sidebar-visible? true})
    :render
    (fn [{:keys [state refs]}]
-     (render-display state refs))})
+     (render-display state refs))
+   :component-did-mount
+   (fn [{:keys [state refs]}]
+     (set! js/window.onscroll
```
dmohs: Prefer:
```
(.addEventListener js/window "scroll" (fn [] ...))
```
to avoid clobbering other listeners.

dmohs:
One comment, but this looks good.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
eddiebroad: that's a good point, especially if there are multiple listeners which may be the case as the codebase matures.

MattPutnam: How do I remove the listener then?

dmohs: Good question! It's as simple as applying this totally godawful pattern:
```
:component-did-mount
(fn [{:keys [this]}]
  (set! (.-onScrollHandler this)
    (fn [] (let [visible...
  (.addEventListener "scroll" (.-onScrollHandler this)))
:component-will-unmount
(fn [{:keys [this]}]
  (.removeEventListener "scroll" (.-onScrollHandler this)))
```
I am so, so sorry.

MattPutnam:
Is this ok to merge?

PR #77, opened 2015-08-12
Hack to make it easier to re-auth with Rawls when necessary.

dmohs:


abaumann:
Yeah just temporary until google auth stuff is worked out (in one of the upcoming sprints) - this is mostly because during demos or just testing things you might need a new token and dont need to change tabs to do so (which is especially disruptive for demos).  Great addition!

MattPutnam:
:+1:

PR #78, opened 2015-08-12
change the URL path to match the orchestration endpoint path

eddiebroad:
indicated in src/main/scala/org/broadinstitute/dsde/firecloud/service/MethodsService.scala


dmohs:
:+1: Yay this works for me!

PR #82, opened 2015-08-12
Reworking how things work after a Rawls update.

MattPutnam:
NOTE: This code is dependent on PR#50 in orchestration.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_method_confs.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_method_confs.cljs
@@ -28,6 +28,8 @@
        :methodStoreConfig {:methodConfigNamespace (str "msc_ns_" (inc i))
                            :methodConfigName (str "msc_n_" (inc i))
                            :methodConfigVersion (str "msc_v_" (inc i))}
+       ;; Real data doesn't have the following fields, but for mock data we carry the same
+       ;; objects around, so initialize them here for convenience
        :inputs {"Input 1" "[some value]"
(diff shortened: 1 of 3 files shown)


eddiebroad:
From this commit, should there be seen any difference from the user perspective (in the browser)?  I didn't see any difference from that perspective.

eddiebroad:
PR 50 of the orchestration was happily merged around 1:58PM this afternoon!

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/workspace_method_confs.cljs
```diff
@@ -28,6 +28,8 @@
        :methodStoreConfig {:methodConfigNamespace (str "msc_ns_" (inc i))
                            :methodConfigName (str "msc_n_" (inc i))
                            :methodConfigVersion (str "msc_v_" (inc i))}
+       ;; Real data doesn't have the following fields, but for mock data we carry the same
+       ;; objects around, so initialize them here for convenience
```
dmohs: Nice comment.

dmohs:
:+1:

MattPutnam:
@eddiebroad The behavior is slightly different.  The method config edit page now has to load the config from Rawls, instead of just using the same object from the table in the method config tab.

eddiebroad:
@MattPutnam  (2-3 min before this comment) it looks like lines 220-228 do that? is that right?

MattPutnam:
Right.

eddiebroad:
:+1:

eddiebroad:
i don't see why this shouldn't be merged

PR #84, opened 2015-08-13
chet the PO requested these changes

eddiebroad:
for these 6 fields to be in the view for selecting an MC to import


MattPutnam:
It would be nice to parse the date into something more readable, but otherwise :+1:

eddiebroad:
@MattPutnam  is there a particular column which deserves additional parsing?  My thinking is that the synopsis field column content might eventually get some text area or be bigger ... any such change might depends on the size of the contained text.

MattPutnam:
The created date column.  And actually, I didn't notice this at first, but the label would be "Created Date", not "Create Date".  Still :+1: as-is though.

eddiebroad:
This didn't show up via the hot-related, but when I reloaded from the begining I see an extra column on the right....I'm not sure if that's from this PR or if it's from the table v2?  Do you see it?

PR #85, opened 2015-08-13
Added filterer to tables

MattPutnam:
Filterer is controlled by `:filterable?` property, defaults to `true`.

Columns filter by string value, or a specified :`filter-by` function taking the column value.  Filtering can be disabled for a column by setting `:filter-by :none`.

Filtering has been added to all tables.  The only changes needed to them was to specify a custom filterer for columns that were not `str`-able.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -1,11 +1,13 @@
 (ns org.broadinstitute.firecloud-ui.common.table
   (:require
-    clojure.string
+    [clojure.string :refer [trim]]
     [dmohs.react :as react]
     [inflections.core :refer [pluralize]]
     [org.broadinstitute.firecloud-ui.common :as common]
+    [org.broadinstitute.firecloud-ui.common.components :as comps]
     [org.broadinstitute.firecloud-ui.common.style :as style]
     [org.broadinstitute.firecloud-ui.common.icons :as icons]
+    [org.broadinstitute.firecloud-ui.utils :as utils]
     ))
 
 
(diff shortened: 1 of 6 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -26,6 +26,10 @@
   ([s what start-index] (.indexOf s what start-index)))
 
 
+(defn contains-ignore-case [s what]
+  (<= 0 (str-index-of (.toLowerCase s) (.toLowerCase what))))
```
dmohs: Might be worth using `clojure.string/lower-case`.

eddiebroad:
overall :+1:

src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -170,19 +172,30 @@
...
      (let [paginator-above (= :above (:paginator props))
            paginator-below (= :below (:paginator props))
            paginator [Paginator {:ref "paginator"
-                                 :num-rows (count (:data props))
+                                 :num-rows (count (:filtered-data @state))
                                  :onChange #(react/call :handle-pagination-change this)}]]
        [:div {}
+        (when (:filterable? props)
+          (let [apply-filter #(swap! state assoc :filtered-data
+                               (react/call :filter-data this (trim (.-value (.getDOMNode (@refs "filter-field"))))))]
```
dmohs: `(react/call :filter-data this (-> (@refs "filter-field") .getDOMNode .-value trim))` would be another option if you like.

dmohs:
:+1: Really nice to have the filtering everywhere. Good to merge.

src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
eddiebroad: more clear now

eddiebroad: it's more clear now

PR #87, opened 2015-08-13
First pass at selecting an entity

MattPutnam:
This isn't totally complete, I'm still waiting on the service call to get all of the entities in a map.  As it stands, the table doesn't give indications that it's loading and there is no error reporting if one of the 'get-entitites-of-type' calls fails, but this code is going to get replaced so I'm leaving it for now.

I'd like to get this in because there are new features that I want to use to finish off DSDEEPB-24.


dmohs:
:+1:

PR #89, opened 2015-08-14
Move workspace details into separate file.

dmohs:


eddiebroad:
Having compiled and viewed the result too in the UI I can see no reason not to merge

dmohs:
Thanks Eddie!

PR #90, opened 2015-08-14
Editing a method configuration now works.

MattPutnam:
TODO: make a unified orchestration call to do the full update at once


eddiebroad:
@MattPutnam  does it work when "use-live-data" is set to true? is there a good way to experience the "editing"? does the back-end persistence work do you know Matt?

MattPutnam:
Yes, this should all work.

eddiebroad:
wonderful!

dmohs:
:+1: Code looks good. I like the rename shim for now. Merging since Rawls is down and we can't test.

PR #92, opened 2015-08-14
Fixed table refresh issue

MattPutnam:


src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -274,4 +275,8 @@
          (:data props))))
    :handle-pagination-change
    (fn [{:keys [this refs]}]
-     (react/call :set-rows (@refs "body") (react/call :get-sliced-data this)))})
+     (react/call :set-rows (@refs "body") (react/call :get-sliced-data this)))
+   :component-will-receive-props
+   (fn [{:keys [this state next-props refs]}]
+     (swap! state assoc :filtered-data
```
dmohs: I fixed this too. My code looks like:
```clojure
:component-will-receive-props
(fn [{:keys [props next-props state]}]
  (when-not (apply = (map :data [props next-props]))
    (swap! state assoc :filtered-data (:data next-props))))
```

dmohs:
:+1: As discussed on HipChat.

PR #94, opened 2015-08-14
More table futzing

MattPutnam:
Left as separate commits so it's easier to see the changes.

* Fixed the paginator page range to always show (at most) 5 numbers, so next and prev buttons don't move out from under the mouse.
* Changed the column resize cursor
* Double-click a column resizing tab to reset the column to its initial width
* Table filtering now checks if the row contains all of the space-separated words in the search text
* Refactor of header/body cell rendering code


eddiebroad:
:+1:

PR #98, opened 2015-08-17
Gave filter a visual cue to when it's applied

MattPutnam:
I would appreciate feedback on color choice

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -182,6 +182,31 @@
         (:rows @state))])})
 
 
+(react/defc Filterer
+  {:get-initial-state
+   (fn [] {:initial true :synced true})
+   :render
+   (fn [{:keys [state this]}]
+     [:span {}
+      (style/create-text-field {:ref "filter-field" :placeholder "Filter"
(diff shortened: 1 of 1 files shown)


rushtong:
I think, if we want to indicate that a list is filtered using colors, then perhaps we should leave the filter box shaded as long as there is a filter in effect.

It would also be more idiomatic to add a line of text to the result set description at the bottom that says, instead of 
```
1 - 6 of 6 results
```
something like 
```
Showing 1 - 6 of 12 filtered results
```

rushtong:
Datatables (http://datatables.net/examples/basic_init/zero_configuration.html) does something like this:
`Showing 1 to 2 of 2 entries (filtered from 57 total entries)`

MattPutnam:
Good idea about the text.  I'll see if I can get that working.

The idea of the text field background color is to indicate when the table is actually reflecting the filter text.  An existing problem is that when the contents of the table change out from under it, it leaves the filterer in a weird state.  This way it can re-highlight that it isn't in sync and needs to be reapplied.

rushtong:
I think the filter box shading is nice, and if it helps clarify the sync state, then that's good too. A change to the textual description would be more apparent than the filter box shading.

rushtong:
:+1: - Love the new color and text!

PR #102, opened 2015-08-17
Add radio buttons to select entities

MattPutnam:
@davidangb the radio buttons leave the selected entity in a state variable called `:selected-entity`


davidangb:
:+1:

PR #103, opened 2015-08-18
first pass at submitting analyses; work in progress

davidangb:
This is a very early PR, since I know everybody is waiting on this: it adds a button to submit an analysis. However ...

*there is no feedback to the end user yet*, beyond browser console logging. So it's pretty incomplete, but hopefully I can get early code review, and others can get access to this functionality for their own testing.

Still left to do:
* success/error messaging to the user; something primitive, but we need something.
* allow the user to submit an empty expression, which then will use the rawls/cromwell defaults
* better canned response for development, when not using live data

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
@@ -143,93 +143,110 @@
        "Launch Analysis"]])])
 
 (defn- render-launch-overlay [state refs workspace config]
-  [comps/ModalDialog
-   {:show-when (:submitting? @state)
-    :dismiss-self #(swap! state assoc :submitting? false)
-    :width "80%"
-    :content
-    (react/create-element
-      [:div {}
-       [:div {:style {:backgroundColor "#fff"
-                      :borderBottom (str "1px solid " (:line-gray style/colors))
-                      :padding "20px 48px 18px"
-                      :fontSize "137%" :fontWeight 400 :lineHeight 1}}
-        "Select Entity"]
-       [:div {:style {:position "absolute" :top 4 :right 4}}
-        [comps/Button {:icon :x :onClick #(swap! state assoc :submitting? false)}]]
-       [:div {:style {:padding "22px 48px 40px" :backgroundColor (:background-gray style/colors)}}
-        (style/create-form-label "Select Entity Type")
-        (style/create-select
-          {:style {:width "50%" :minWidth 50 :maxWidth 200} :ref "filter"
-           :onChange #(let [value (-> (@refs "filter") .getDOMNode .-value)]
-                       (when-not (= value "Select...")
-                         (utils/call-ajax-orch
-                           (paths/list-all-entities-path workspace value)
-                           {:on-success (fn [{:keys [parsed-response]}]
-                                          (swap! state assoc :entities parsed-response
-                                            :selected-entity (first parsed-response)))
-                            :on-failure (fn [{:keys [status-text]}]
-                                          (utils/rlog "Error: " status-text))
-                            :mock-data [{"entityType" value
-                                         "name" "A mock entity"
-                                         "attributes" {}}]})))}
-          (:entity-types @state))
-        (style/create-form-label "Select Entity")
-        (if (zero? (count (:entities @state)))
-          (style/create-message-well "No entities to display.")
-          [:div {:style {:backgroundColor "#fff" :border (str "1px solid " (:line-gray style/colors))
-                         :padding "1em" :marginBottom "0.5em"}}
-           [table/Table
-            {:columns [{:header "" :starting-width 40 :resizable? false
-                        :content-renderer (fn [row-index data]
-                                            [:input {:type "radio"
-                                                     :checked (identical? data (:selected-entity @state))
-                                                     :onChange #(swap! state assoc :selected-entity data)}])}
-                       {:header "Entity Type" :starting-width 100}
-                       {:header "Entity Name" :starting-width 100}
-                       {:header "Attributes" :starting-width 400}]
-             :data (map (fn [m]
-                          [m
-                           (m "entityType")
-                           (m "name")
-                           (m "attributes")])
-                     (:entities @state))}]])
-        (style/create-form-label "Define Expression")
-        (style/create-text-field {:ref "expressionname" :defaultValue "this"})]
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
                            (m "attributes")])
                      (:entities @state))}]])
         (style/create-form-label "Define Expression")
-        (style/create-text-field {})]])}])
+        (style/create-text-field {:ref "expressionname" :defaultValue "this"})]
+
+       [:div {:style {:fontSize "106%" :lineHeight 1 :textAlign "center"}}
```
dmohs: If you use Matt's Button class (in components), you'll get nice button styling for free.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
...
+        [:div {:style   {:padding         "0.7em 0" :cursor "pointer"
+                         :backgroundColor (:button-blue style/colors)
+                         :color "#fff" :borderRadius 4
+                         :border          (str "1px solid " (:line-gray style/colors))}
+
+               ;; TODO: what should we show in the UI after submitting?
```
dmohs: I think a `js/alert` would work for a first pass.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
...
+                                                           :entityType                   ((:selected-entity @state) "entityType")
+                                                           :entityName                   ((:selected-entity @state) "name")
+                                                           :expression                   (-> (@refs "expressionname") .getDOMNode .-value)
+                                                           })
+                               :headers {"Content-Type" "application/json"}
+                               :on-done (fn [{:keys [success? xhr]}]
+                                            (if success?
+                                              (utils/rlog "Submission id: " ((utils/parse-json-string (.-responseText xhr)) "submissionId"))
+                                              (utils/rlog "Error: " (.-responseText xhr))))
+                               :canned-response {:responseText (utils/->json-string
```
dmohs: Seems like all you really need is: `{"submissionId" "my-test-submission"}`

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
...
+                         :backgroundColor (:button-blue style/colors)
+                         :color "#fff" :borderRadius 4
+                         :border          (str "1px solid " (:line-gray style/colors))}
+
+               ;; TODO: what should we show in the UI after submitting?
+               ;; TODO: if no expression entered, do not send one in the ajax request
+               ;; TODO: disable/enable submit button based on the state of entity selection & expression value
+               ;; TODO: diable submit button after submitting
+               ;; TODO: better canned response
+               :onClick (fn [{:keys []}]
```
dmohs: This function only takes one argument: the click event. This should be either `(fn [e]...)` or `(fn [] ...)`

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
...
+                            (utils/ajax-orch
+                              (paths/submit-method-path workspace)
+                              {:method :post
+                               :data (utils/->json-string {:methodConfigurationNamespace (config "namespace")
+                                                           :methodConfigurationName      (config "name")
```
dmohs: Nit: Prefer to not line these up. Easy to adjust in IntelliJ (if you didn't do this by hand).

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
```diff
@@ -195,7 +195,41 @@
...
+        (style/create-text-field {:ref "expressionname" :defaultValue "this"})]
+
+       [:div {:style {:fontSize "106%" :lineHeight 1 :textAlign "center"}}
+        [:div {:style   {:padding         "0.7em 0" :cursor "pointer"
+                         :backgroundColor (:button-blue style/colors)
+                         :color "#fff" :borderRadius 4
+                         :border          (str "1px solid " (:line-gray style/colors))}
+
+               ;; TODO: what should we show in the UI after submitting?
+               ;; TODO: if no expression entered, do not send one in the ajax request
```
dmohs: You'll probably want to do:
```clojure
(let [expression (clojure.string/trim (-> (@refs "expressionname") .getDOMNode .-value))]
  ...
  :data .. (merge {:methodConfigurationNamespace...}
                 (when-not (blank? expression) {:expression expression}))
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
rushtong: I have to say that I strongly dislike js alerts. We should look into a more user-friendly pattern.

dmohs: I'm with you @rushtong. I think we will eventually want to put the user on the "submission status" page for the returned submission ID. We just need some kind of placeholder until that page is created.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
davidangb: hmm ... I happen to like the full-width button! But it seems nobody else does :)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
davidangb: I've implemented a pretty hacky placeholder without an alert ... will update the PR momentarily.

davidangb:
ready for re-review, I hope!

dmohs:
:+1: Let's get this in so we can demo. We can follow-up with cleanups.

PR #104, opened 2015-08-18
have an "onImport" function to be called when

eddiebroad:
an import succeeds.  Causes you to go back to
the MC listing in a workspace and for it to be
refreshed too.

Most of the commit/diff here consists of "relaying" a reference
to the handler from its actual body to where it's actually called

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -82,11 +82,11 @@
                                   :data (utils/->json-string post-data)
                                   :on-done (fn [{:keys [success? xhr]}]
                                              (if success?
-                                               (utils/rlog "SUCCESS in import ! trigger re-render of workspace MC list here?")
+                                               ((:on-import props))
                                                (js/alert (str "Error in import : "  (.-statusText xhr)))))})))}]]))})
 
 
-(defn render-import-modal [state props refs]
+(defn render-import-modal [state props refs on-import]
   [comps/ModalDialog
    {:width 750
     :content (react/create-element
@@ -110,6 +110,7 @@
                  [ModalImportOptionsAndButton {:init-Name (get-in @state [:selected-conf "name"])
                                                :init-Namespace (get-in @state [:selected-conf "namespace"])
                                                :init-SnapshotId (get-in @state [:selected-conf "snapshotId"])
+                                               :on-import on-import
                                                :parental-state state
                                                :workspace (get props :workspace)}]]])
     :show-when true}])
@@ -120,7 +121,7 @@
    (fn [{:keys [state props refs]}]
      [:div {}
       (when (:show-import-mc-modal? @state)
-        (render-import-modal state props refs))
+        (render-import-modal state props refs (:on-import props)))
       (cond
         (:loaded-import-confs? @state)
         (if (zero? (count (:method-confs @state)))
@@ -179,7 +180,7 @@
    :width "90%"})
 
 
-(defn render-import-overlay [state  workspace ]
+(defn render-import-overlay [state  workspace on-import]
   (let [clear-import-overlay #(swap! state assoc :import-overlay-shown? false)]
     (when (:import-overlay-shown? @state)
       [:div {:style modal-import-background
@@ -195,5 +196,5 @@
                        :padding "20px 48px 18px"}}
          [:div {:style {:fontSize 24 :align "center" :textAlign "center" :paddingBottom "0.5em"}}
           "Select A Method Configuration For Import"]
-         [ImportWorkspaceMethodsConfigurationsList {:workspace workspace}]
+         [ImportWorkspaceMethodsConfigurationsList {:workspace workspace :on-import on-import}]
          [:div {:style {:paddingTop "0.5em"}}]]]])))
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs.cljs
@@ -43,7 +43,9 @@
   {:render
    (fn [{:keys [props state]}]
      [:div {}
-      (importmc/render-import-overlay state (:workspace props) )
+      (importmc/render-import-overlay state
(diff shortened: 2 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -82,11 +82,11 @@
                                   :data (utils/->json-string post-data)
                                   :on-done (fn [{:keys [success? xhr]}]
                                              (if success?
-                                               (utils/rlog "SUCCESS in import ! trigger re-render of workspace MC list here?")
+                                               ((:onImportFunc props))
```
MattPutnam: I would prefer the name `:on-import`

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -120,7 +121,7 @@
    (fn [{:keys [state props refs]}]
      [:div {}
       (when (:show-import-mc-modal? @state)
-        (render-import-modal state props refs))
+        (render-import-modal state props refs (:onImportFunc props)))
```
MattPutnam: Just leave this in props, no need to pull it out and pass it around as an extra parameter.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
eddiebroad: I will go ahead and add a commit to suit the preference.

Are there any other concerns?  Is the PR otherwise good to merge?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs.cljs
```diff
@@ -99,9 +101,12 @@
                                             (react/call :load-method-configs this)
                                             (swap! state update-in [:method-confs]
                                               assoc (:selected-index @state) new-conf)))}]
-         (:method-confs-loaded? @state) [WorkspaceMethodsConfigurationsList {:workspace (:workspace props)
-                                                                             :method-confs (:method-confs @state)
-                                                                             :parent-state state}]
+         (:method-confs-loaded? @state) [WorkspaceMethodsConfigurationsList
+                                          {:onImport
+                                             (fn [] (react/call :load-method-configs this))
```
MattPutnam: This is right for the real data, but it's going to clobber mock data.  I think this callback function should take the imported config as a parameter, and then in the case of mock data, this code can just `conj` it onto the mock configs list.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -179,7 +180,7 @@
    :width "90%"})
 
 
-(defn render-import-overlay [state  workspace ]
+(defn render-import-overlay [state  workspace onImportFunc]
```
MattPutnam: Nit: there's an extra space here.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -120,7 +121,7 @@
    (fn [{:keys [state props refs]}]
      [:div {}
       (when (:show-import-mc-modal? @state)
-        (render-import-modal state props refs))
+        (render-import-modal state props refs (:on-import props)))
```
MattPutnam: Looks like my previous comments got blasted by the force-push.  This doesn't need to be pulled out of props here, you can leave it in and then extract it on line 113.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs.cljs
```diff
@@ -99,9 +101,12 @@
                                             (react/call :load-method-configs this)
                                             (swap! state update-in [:method-confs]
                                               assoc (:selected-index @state) new-conf)))}]
-         (:method-confs-loaded? @state) [WorkspaceMethodsConfigurationsList {:workspace (:workspace props)
-                                                                             :method-confs (:method-confs @state)
-                                                                             :parent-state state}]
+         (:method-confs-loaded? @state) [WorkspaceMethodsConfigurationsList
+                                          {:on-import
+                                             (fn [] (react/call :load-method-configs this))
```
MattPutnam: Make this function take the imported config as a parameter, and then if we're using mock data we can just update the list of mock configs.  Otherwise calling load-method-configs is going to blast everything.

eddiebroad: I purposely wanted to re-load all the method configs.  What do you mean by "blast everything"?  I wanted to trigger updating of the entire thing.  Are you saying that it's better to just add an MC to the list than to reload the list of MCs?  What is your motivation?

MattPutnam: I'm talking about mock.  If you call `:load-method-configs` again, it will just re-generate a list of mock data.

eddiebroad: I guess since it was mock data, that actually importing anything wasn't terribly important.  I had never thought to add a mock MC to a list of mock MCs in a mock workspace.  I thought that refreshing the list was the important thing to do

MattPutnam: Mock doesn't mean we don't care about things being consistent and sensible.

dmohs: I think our codebase should be mostly unaware that mock data exists, so if the only reason to include a parameter is to make it look better when using mock data, my preference would be to not include it. Otherwise, I might look at this code later and say, "Why does the config get passed in? That seems less safe than refreshing the list to make sure the import actually went through."

If you decide it is worth doing, then I would at least ask for a comment calling out that the parameter is only used for mock data/demo purposes. My hope, though, is that we will never have to use mock data for demos again. Maybe that's naively optimistic :)

PR #105, opened 2015-08-18
Used new get-all-entities call

MattPutnam:
No more loading every time you select an entity type from the combo box


dmohs:
API endpoints don't seem to be working so I can't test this, but code looks good. :+1:

PR #112, opened 2015-08-18
Method configs refactor...

dmohs:
- only make AJAX call when listing configs
- hot reloading works better (more to come)

Make sure to ignore whitespace to reduce noise. Tested by-hand through launching an analysis.


MattPutnam:
:+1:

PR #113, opened 2015-08-19
DO NOT MERGE: Upgrade clojure, clojurescript, lein, figwheel to latest versions

davidangb:
This all seems to work for me just fine. I can't see any functional problems in manual testing. We may find some over time.

Please do not merge this until after the multi-team demo, to avoid risk of regression!


MattPutnam:
:+1:

PR #117, opened 2015-08-20
First of several config editor refactoring...

dmohs:
- moves launch overlay to separate file
- uses built-in gensym instead of our own code

Pretty straightforward code move. Tested manually.


MattPutnam:
:+1: Works for me.  Interesting that I can find very little about gensym via google.

PR #118, opened 2015-08-20
Show/hide columns

MattPutnam:
This is the first half of the hide/reorder table columns story.  As such, much of the naming reflects the broader scope (things labeled "reorder") and there is some preparatory work for reordering that doesn't directly benefit show/hide.

Tables can now allow showing/hiding of columns via a `:reorderable-columns?` property, defaulting to true.  Individual columns can then have hiding (and later, reordering) disabled via `:reorderable?` on the column.  See the launch analysis dialog for a place where it's necessary to disable this.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -3,7 +3,8 @@
     [dmohs.react :as react]
     [org.broadinstitute.firecloud-ui.common :as common]
     [org.broadinstitute.firecloud-ui.common.icons :as icons]
-    [org.broadinstitute.firecloud-ui.common.style :as style]))
+    [org.broadinstitute.firecloud-ui.common.style :as style]
+    ))
 
 
 (react/defc Spinner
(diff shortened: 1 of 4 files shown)


eddiebroad:
nice!

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -3,7 +3,8 @@
     [dmohs.react :as react]
     [org.broadinstitute.firecloud-ui.common :as common]
     [org.broadinstitute.firecloud-ui.common.icons :as icons]
-    [org.broadinstitute.firecloud-ui.common.style :as style]))
+    [org.broadinstitute.firecloud-ui.common.style :as style]
+    ))
```
dmohs: This is the style I prefer for imports (having the close parens on a separate line), so that adding a new import (fairly common) results in a one-line diff rather than two.

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -113,6 +114,24 @@
         [:div {} (:component (nth (:items props) (:active-tab-index @state)))]])}))
 
 
+(react/defc AnchoredDialog
```
dmohs: I have used this concept elsewhere, but I added an onClick handler to the window. This pattern (adding an element behind the box to catch the click) results in a much simpler and cleaner implementation. Really nice.

dmohs:
Really nice. Works great. On the next iteration, add a label to enclose each checkbox and its label text so that clicking on the column name also checks/un-checks the box.

PR #120, opened 2015-08-21
Fix hot reloading bug...

dmohs:
See this commit: https://github.com/dmohs/react-cljs/commit/3c2f255ee607391547a289546c9da919e940c3eb

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -6,7 +6,7 @@
   :dependencies
   [
    ;; React library and cljs bindings for React.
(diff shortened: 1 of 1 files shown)


eddiebroad:
I compiled and used it with the hot-reloader and experienced this.  I see no reason not to merge

eddiebroad:
If there is a problem upon/after merging we can deal with that

eddiebroad:
:+1:

PR #125, opened 2015-08-21
Moar hot reloading fixes plus some cleanups

dmohs:
The AJAX call to load workspace details now only happens on the summary page. This prevents the reload from clobbering state.


MattPutnam:
It doesn't look like utils/ws->id is being used anywhere.

dmohs:
@MattPutnam Nice catch! Thank you. I had to use it all over the place during the refactor, but after it was done, not needed.

MattPutnam:
:+1:

PR #126, opened 2015-08-21
Allow column resize drag to work outside the table's bounds

MattPutnam:
This was a wild ride.  Making a checkpoint for a PR now so there isn't ever too much craziness at once.

I keep naming these branches things like "column drag" thinking I'm finally going to get around to working on drag-to-reorder, but I keep hitting things that need to get done first :/

This PR allows the column resizing gesture to work outside of the table itself.  To accomplish this, a mouse listener is added to the window rather than using onMouseMove on the table component itself.  This caused text to be selected wildly, so a new utility for disabling and restoring the text selection properties was created.


eddiebroad:
if it was a wild ride, does that it was a drag???

eddiebroad:
I compiled and reviewed the code and see no reason why it shouldn't be merged

PR #128, opened 2015-08-24
Replace the logo image with FireCloud sized similarly to the original…

rushtong:
… logo.

Fix for https://broadinstitute.atlassian.net/browse/DSDEEPB-1118

screen shot 2015-08-24 at 3 50 31 pm


davidangb:
code looks fine, nice small change. I wish it looked better visually, but there's not much to do right now. How about making it bold and using the link-blue color? line 35 has an example of how to reference common colors.

rushtong:
@davidangb - I updated as per feedback. I think the link color looks a little washed out and the brighter blue button color might be a better choice.

davidangb:
:+1: alright, great - make it button-blue if you think that looks better, feel free to merge on your own!

PR #130, opened 2015-08-24
Data tab now shows all entities

MattPutnam:
Mostly a copy/paste job from the submit analysis dialog.  Not exactly what was asked for but it's what we want going forward, I think.


dmohs:
:+1:

PR #132, opened 2015-08-25
Modal work, hot reloading fix, table fix

dmohs:
Remove usages of :show-when in ModalDialog. Instead, it is only rendered when it should appear.

Eliminate scrolling of body when modal is displayed.

Remove derived state from table (instead, calculate the derived data during render).

Hot reloading fix.

Left as separate commits for easier review.


MattPutnam:
There's a problem with the table logic.  Look at the workspaces list--something is confused between the filterer and the paginator

dmohs:
Thanks for catching that @MattPutnam. Ready for another look.

MattPutnam:
:+1:

PR #135, opened 2015-08-25
audit table stuff, also DSDEEPB-1119

MattPutnam:
Added as much table sorting and filtering as made sense throughout the application.  Also @dmohs your refactoring broke sorting and I somehow missed that when I reviewed it, but it was a quick fix.

Specifying sorting is getting kinda boilerplatey, but I have ideas for future enhancement:  I'd like to be able to export tables to CSV/TSV, which will require a way to render fancy cells to text.  Then filter and sort can work off of those by default.  Way out of scope for this though.


eddiebroad:
I git checkout'ed the code, compiled it, and  i experienced the new tables.  :+1:

PR #136, opened 2015-08-25
Unify dialogs

dmohs:
- use gray for background instead of "highlight" blue
- better scrolling behavior throughout
- ESC key works from anywhere on page


MattPutnam:
:+1: Nice, the scrolling is way better.  Only thing is that the dialog can push off the bottom of the page:

screen shot 2015-08-25 at 3 47 04 pm

But I don't suppose there's really anything we can do about that.

dmohs:
You might be able to recalculate the hight of the background after rendering the dialog, but it could be messy. Possible future optimization.

PR #137, opened 2015-08-25
Submissions table with selection

dmohs:


eddiebroad:
:+1:

eddiebroad:
DSDEEPB-1078 can be upcoming code/project to work on...

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/monitor_tab.cljs
```diff
@@ -0,0 +1,133 @@
...
+       :submissionEntity {:entityType "sample"
+                          :entityName (str "sample_" (inc i))}
+       :submitter "abaumann@broadinstitute.org"})
+    (range (rand-int 50))))
+
+
+(defn- render-submissions-table [submissions on-submission-clicked]
+  [table/Table
+   {:columns
+    [{:header "Date" :starting-width 200 :sort-by :value
```
MattPutnam: This needs a custom :sort-by and :filter-by, otherwise it tries to sort and filter by the string representation of the entire object.

dmohs: I think we should import MomentJs to do real date formatting, but that should be addressed later. For now, sorting by the string value is actually correct because of the nature of the ISO format (unless I'm missing something).

MattPutnam: It's more than that.  The column value is the entire map.  For example, try filtering on "submi", and notice that everything matches.  It's hitting the "submissionID" key.  Also note that sorting is really slow.

dmohs: Thanks for catching that. You were right on. Added sort-by and filter-by functions.

PR #138, opened 2015-08-26
Fix delete button, also cleanup

MattPutnam:
Made the new delete button not show when editing, and while I was at it I converted all of the display-nones to whens.


dmohs:
:+1: Much better.

PR #139, opened 2015-08-26
icons now properly contained

MattPutnam:
Very little visual change on this one.

Previously, the custom icons in components.cljs were not contained properly and wanted to center in their parent component, requiring extra external effort to override this.  So there's actually *more* code around some usages of them now, but now it's easier to use them across the board.


eddiebroad:
:+1:  for easier

PR #142, opened 2015-08-27
Submissions detail page

MattPutnam:
Also had to do even more work on the icons.  Now we're needing a bunch of `:verticalAlign "middle"` to get stuff lined up.


dmohs:
:+1: Code looks great. Went through the UI quite a bit. All I can say is that is looks absolutely phenomenal.

PR #143, opened 2015-08-27
code cleanup

MattPutnam:
This page had been pretty neglected :(

Also, there are several more columns from Agora that we could be displaying.


eddiebroad:
yes, the code had been neglected a bit now, it's good it's getting some attention

eddiebroad:
the code is tidier indeed after this commit :+1:

dmohs:
Agree with Eddie. Nice.

PR #144, opened 2015-08-28
Import Method Configuration Refactor

rushtong:
For DSDEEPB-1072 

Update import method configuration so that it redirects to the after the configuration is successfully imported.

Notes:
* We need to treat config similarly to the way we do workspace - it should be an object with name and namespace.


dmohs:
:+1:

PR #146, opened 2015-08-28
Use Moment.js on submissions stuff

MattPutnam:
Added Moment.js, and used some of its functionality to better generate random times for mock data, and to render absolute and relative times in the submissions list and details page.

Note for future use:  Including Moment.js causes the compiled javascript to include a global "moment" function.  This makes it a runtime error to try to bind a Moment to a variable called "moment".


dmohs:
:+1: Love the random recent time function.

PR #147, opened 2015-08-28
progress on UI front for *aborting* a submission/job

eddiebroad:
cause reload of the details if success in *aborting* the submission

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
@@ -122,9 +122,38 @@
                (:workflows props))}])})
 
 
+(react/defc AbortButton
+  {:render (fn [{:keys [state this]}]
+             (if (:aborting-submission? @state)
+               [comps/Blocker {:banner "Aborting submission ..."}]
+               [:div {:style {:padding "0.7em 0" :marginTop "0.5em" :color "#fff" :cursor "pointer"
+                              :backgroundColor (:exception-red style/colors) :borderRadius 5}
+                      :onClick #(when (js/confirm "Are you sure?")
+                                 (react/call :abort-submission this))}
+                (icons/font-icon {:style {:fontSize "160%" :marginRight 14 :verticalAlign "middle"}}
+                  :status-warning-triangle)
+                [:span {:style {:fontSize "125%" :verticalAlign "middle"}} "Abort"]]))
+   :abort-submission (fn [{:keys [props state]}]
+                       (swap! state assoc :aborting-submission? true)
+                       (let [path (paths/get-abort-submission-path
+                                    (:workspace-id props)
+                                    (:submission-id props))
+                             canned-response {:status 200 :delay-ms (rand-int 2000)}
+                             on-done (fn [{:keys [success? status-text]}]
+                                       (swap! state assoc :aborting-submission? false)
+                                       (if success?
+                                         ((:on-abort props))
+                                         (js/alert (str "Error in aborting the job : " status-text))))]
+                         (utils/ajax-orch path
+                           {:headers {"Content-Type" "application/json"}
+                            :method "DELETE"
+                            :canned-response canned-response
+                            :on-done on-done})))})
+
+
 (react/defc Page
   {:render
-   (fn [{:keys [state]}]
+   (fn [{:keys [state props this]}]
      (let [server-response (:server-response @state)
            {:keys [submission error-message]} server-response]
        (cond
@@ -141,12 +170,12 @@
                             :verticalAlign "middle"}}
              (submission "status")]]
            (when (= "Submitted" (submission "status"))
-             [:div {:style {:padding "0.7em 0" :marginTop "0.5em" :color "#fff" :cursor "pointer"
-                            :backgroundColor (:exception-red style/colors) :borderRadius 5}
-                    :onClick #(js/console.log "Insert abort logic here")}
-              (icons/font-icon {:style {:fontSize "160%" :marginRight 14 :verticalAlign "middle"}}
-                :status-warning-triangle)
-              [:span {:style {:fontSize "125%" :verticalAlign "middle"}} "Abort"]])]
+             [AbortButton
+              {:on-abort (fn []
+                          (swap! state assoc :server-response nil)
+                          (react/call :load-details this))
+               :workspace-id (:workspace-id props)
+               :submission-id (submission "submissionId")}])]
           [:div {:style {:float "left"}}
            (style/create-section-header "Method Configuration")
            (style/create-paragraph
@@ -176,7 +205,7 @@
           [WorkflowsTable {:workflows (submission "workflows")}]
           [:div {:style {:padding "3em 0 0.5em 0"}} "Failed to Start:"]
           [WorkflowFailuresTable {:workflows (submission "notstarted")}]])))
-   :component-did-mount
+   :load-details
    (fn [{:keys [props state]}]
      (let [url (paths/submission-details (:workspace-id props) (:submission-id props))
            on-done (fn [{:keys [success? status-text get-parsed-response]}]
@@ -187,4 +216,5 @@
                                             (create-mock-submission-details (:submission-id props)))
                             :status 200
                             :delay-ms (rand-int 2000)}]
-       (utils/ajax-orch url {:on-done on-done :canned-response canned-response})))})
+       (utils/ajax-orch url {:on-done on-done :canned-response canned-response})))
+   :component-did-mount (fn [{:keys [this]}] (react/call :load-details this))})
--- a/src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
@@ -45,5 +45,8 @@
 (defn submission-details [workspace-id submission-id]
   (str (submissions-list workspace-id) "/" submission-id))
 
+(defn get-abort-submission-path [workspace-id submission-id]
+  (submission-details workspace-id submission-id))
+
 (defn rm-method-configuration-path [workspace-id config]
   (update-method-config-path workspace-id config))


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,50 @@
                    (:workflows props))}])))})
 
 
+(react/defc AbortSubmission
```
MattPutnam: This doesn't make sense as a logical unit.  I don't think the abort logic needs to be wrapped in its own React class, but if you want to do it that way, then it makes more sense to make an AbortButton class and not also encompass the status display.

eddiebroad: In the sentence "This doesn't make sense as a logical unit." what do you refer to by "this"?

Maybe it should be named "AbortButton" instead of "AbortSubmission"?  I had thought to name it "AbortButton", but said  "it's a div, not a button", so then thought "AbortDiv" should be a name...but that might be confusing so then I thought "AbortSubmission" which is maybe confusing too.

By "status display" do you refer to the div whose backgroundcolor calls the "color-for-submission"?

MattPutnam: By "this" I mean the AbortSubmission component.  It's not just an abort button, it's the status label and the abort button.

eddiebroad: i think i see what you mean.....one possiblility is to move the "status" (reffering to the  div whose backgroundcolor calls the "color-for-submission") *outside* of the "AbortSubmission" as you mention.

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
```diff
@@ -45,5 +45,8 @@
 (defn submission-details [workspace-id submission-id]
   (str (submissions-list workspace-id) "/" submission-id))
 
+(defn submission-kill [workspace-id submission-id]
```
rushtong: To be more consistent with the other path method names and more closely align to the named functionality, this should be renamed to `get-abort-submission-path`

eddiebroad: Is it redundant to add "-path" as a suffix because it's already in a file named paths? (and by this logic all the -path suffixes above are also redundant)

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
```diff
@@ -45,5 +45,8 @@
...
+  (submission-details workspace-id submission-id))
```
rushtong: Looking at this in context - it is the exact same path as `submission-details`. Do we really want two methods that do the exact same thing? The path is the same whether you are using a `GET` or a `DELETE` ... why do we need two code paths for the single case?

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
rushtong: Yes, it probably is redundant. That redundancy should be addressed as a separate refactoring task though - not in individual commits.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,43 @@
...
+                       (js/alert (str "Error in stopping the job : " status-text))))]
+       (utils/ajax-orch path
+         {:headers headers
+          :method "DELETE"
+          :canned-response canned-response
+          :on-done on-done})))
+   :render (fn [{:keys [props state this]}]
+             (if (:stopping-submission? @state)
+               [comps/Blocker {:banner "Stopping submission ..."}]
+                (when (= "Submitted" ((:submission props) "status"))
```
MattPutnam: I think this logic should be moved to where AbortButton is created, on line 181.  Don't even create an AbortButton instance unless the status is "Submitted".

rushtong:
It's best to wait for folks to go through the PR in detail before squashing and merging.

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
```diff
@@ -45,5 +45,8 @@
 (defn submission-details [workspace-id submission-id]
   (str (submissions-list workspace-id) "/" submission-id))
 
+(defn get-abort-submission-path [workspace-id submission-id]
+  (submission-details workspace-id submission-id))
+
```
rushtong: Re-adding a previous comment that was lost in the squash:
Looking at this in context - it is the exact same path as submission-details. Do we really want two methods that do the exact same thing? The path is the same whether you are using a GET or a DELETE ... why do we need two code paths for the single case?

MattPutnam: I first created this file and started us down this pattern.  My thinking was that it's a way of documenting which endpoint you're hitting specifically.  At the same time though, I didn't chain them together like this.  My thought was that it's essentially a coincidence that getting a thing and deleting the thing use the same URL.  Then again I'm not that experienced with REST so if this is an idiomatic enough thing I guess it makes sense to merge them.

I want to do a refactor to unify the paths and the mock data.  So when making an AJAX call, you'd just specify the endpoint and it would get the path and the mock data together.  That would go along with this idea of separate functions.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,43 @@
                    (:workflows props))}])))})
 
 
+(react/defc AbortButton
+  {:kill-submission
```
MattPutnam: At this point there's "abort", "kill", and "stop" all floating around.  Let's consolidate on one term.  Rawls uses "abort" for the endpoint, so I think we should go with that.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,43 @@
...
+   (fn [{:keys [props state this]}]
```
rushtong: In discussions with David Mohs, we are of the opinion that we should be avoiding having state passed into child react components. Instead, they should be passed only what they need as arguments and state-handling call-backs.

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
eddiebroad: this function calls the other directly.
however another name shows it can serve dual purposes. (DELETE and GET as you mention).
However, "submission-details" when called,might possibly be confusing in a DELETE contact, so that was the motivation to make something, with a different (maybe less confusing) name to call the other function ( a few lines ahead)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -191,4 +224,6 @@
                                             (create-mock-submission-details (:submission-id props)))
                             :status 200
                             :delay-ms (rand-int 2000)}]
-       (utils/ajax-orch url {:on-done on-done :canned-response canned-response})))})
+       (utils/ajax-orch url {:on-done on-done :canned-response canned-response})))
+   :component-did-mount (fn [{:keys [this]}] (react/call :load-details this))
+   })
```
rushtong: Nit - hanging brace and paren

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
MattPutnam: This isn't a case of that.  The `state` variable here is the component's own state.

rushtong: Thank you - still getting a handle on the various scopes of `state`

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
eddiebroad: i suppose the source of this issue is the overloading of the underlying REST api path with multiple methods ("GET" and "DELETE")?

MattPutnam:
General comment: this branch name is way too long.

rushtong:
I defer to Matt's judgement on the state of this, but I'm finished reviewing.

src/cljs/org/broadinstitute/firecloud_ui/paths.cljs
rushtong: Idiomatic REST does call for the same path for both GET/DELETE and similarly for PUT/POST/PATCH. The difference is just the method.

MattPutnam: Yeah, it's kind of a mess right now.  I think it's out of scope for this PR.  Maybe this should be something we look at refactoring in the next sprint.

rushtong: :+1:

eddiebroad: I'm just EDDIE and I've used AJAX before, but not REST extensively (I've made both front-end AJAX handler via php and back-end AJAX).  Maybe this means I am not "well RESTed"????

I just did this because it made sense to me to compensate for the effect of overloading (GET and DELETE) onto one path.

rushtong: http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api

This example from that page is more correct than what I posted above:
```
GET /tickets - Retrieves a list of tickets
GET /tickets/12 - Retrieves a specific ticket
POST /tickets - Creates a new ticket
PUT /tickets/12 - Updates ticket #12
PATCH /tickets/12 - Partially updates ticket #12
DELETE /tickets/12 - Deletes ticket #12
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,41 @@
...
+           workspace-id (:workspace-id props)
+           path (paths/get-abort-submission-path workspace-id submission-id)
+           headers {"Content-Type" "application/json"}
+           canned-response {:responseText ""
+                            :status 200 :delay-ms (rand-int 2000)}
+           on-done (fn [{:keys [success? status-text]}]
+                     (swap! state assoc :stopping-submission? false)
+                     (if success?
+                       ((:on-kill props))
+                       (js/alert (str "Error in stopping the job : " status-text))))]
```
MattPutnam: Ok, so now the AbortButton only uses the submission's ID, so it can take that instead of the whole submission object.  I still think you're over-using `let`: We don't need to bind submission-id and workspace-id since we're just using them in once place.  Similarly, binding the path and headers doesn't add any value.  On line 137 you can just get rid of the `:responseText` key.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,41 @@
...
+          :method "DELETE"
+          :canned-response canned-response
+          :on-done on-done})))
+   :render (fn [{:keys [props state this]}]
+             (if (:stopping-submission? @state)
+               [comps/Blocker {:banner "Stopping submission ..."}]
+               [:div {:style {:padding "0.7em 0" :marginTop "0.5em" :color "#fff" :cursor "pointer"
+                              :backgroundColor (:exception-red style/colors) :borderRadius 5}
+                              :onClick (fn [e]
+                                         (react/call :kill-submission this))}
```
MattPutnam: Save a line (and some characters, and an unused `e`) by using the `#` function macro.  Also, note that you're not using `this` in `:kill-submission` or `props` in `:render`.

MattPutnam: Also, indentation: I thought you had your onClick inside style at first.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -145,12 +176,12 @@
...
-                    :onClick #(js/console.log "Insert abort logic here")}
-              (icons/font-icon {:style {:fontSize "160%" :marginRight 14 :verticalAlign "middle"}}
-                :status-warning-triangle)
-              [:span {:style {:fontSize "125%" :verticalAlign "middle"}} "Abort"]])]
+             [AbortButton
+              {:on-abort (fn []
+                          (swap! state assoc :server-response nil)
+                          (react/call :load-details this))
+               :workspace-id (:workspace-id props)
+               :submission submission}])]
```
dmohs: I think you can get by with passing only the submission ID rather than the entire submission object/map.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -126,9 +126,40 @@
                    (:workflows props))}])))})
 
 
+(react/defc AbortButton
+  {:abort-submission
```
dmohs: We don't have visibility modifiers for these, unfortunately. So, the convention I've been using is: public methods go first, followed by component lifecycle methods (e.g. :render), followed by private methods. Please move this method after render for consistency with that pattern.

dmohs:
:+1: Looks good. Two tiny tweaks. Also enjoyed the conversation about REST URLs that came up here.

PR #148, opened 2015-08-28
Fix call to launch analysis

dmohs:


rushtong:
:+1: Yep ... that did the trick.

PR #149, opened 2015-08-28
Move handling of empty case inside table component

MattPutnam:


eddiebroad:
therer seems to be a default empty message "There are no rows to display." (new table.cljs line 325)  . it is possible to  override this with a custom empty message (for example "No methods to display" line 15 inmethod_repo.cljs).  nice!

dmohs:
:+1: Great idea.

PR #150, opened 2015-08-31
Final (for now) launch refactoring

dmohs:
Works with mock data. Requests seem good, but can't find good test data.


eddiebroad:
@dmohs  do you like the layout and placement of the blue launch button at the bottom of the  "Launch Analysis..." window?

dmohs:
I think it could be too big to obviously be a button, but changing it is out of scope for this refactoring.

eddiebroad:
I see no reason why it shouldn't be merged

PR #151, opened 2015-08-31
Fixed null reference error with empty tables

MattPutnam:
On the method configs tab with no method configs, trying to click the import button trips here because there is no filterer.  This sort of seems like a hack though...

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -384,10 +384,6 @@
      (let [rows (react/call :get-filtered-data this)]
        (react/call :set-rows (@refs "body") (react/call :get-body-rows this rows))
        (react/call :set-num-rows-visible (@refs "paginator") (count rows))))
-   :component-will-receive-props
-   (fn [{:keys [props state refs]}]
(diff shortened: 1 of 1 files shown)


eddiebroad:
I load the code both before and after this commit.  I saw that the null-reference-error indeed was fixed by the commit so that the error doesn't come up after the commit.

eddiebroad:
Why do you say "This sort of seems like a hack though..."?

MattPutnam:
Ideally, checks like this shouldn't have to be made, and it may be indicative of something broken in the design.

eddiebroad:
What is the significance of "(@refs "filterer")" in this context?

dmohs:
Are you sure this is even needed? If you remove the call altogether, what happens?

MattPutnam:
Hmm, good question.  It was definitely being used at one point, but it looks like refactoring may have eliminated the need for it.

dmohs:
If it isn't used/necessary, my preference is to remove it. GitHub remembers if we need it again.

MattPutnam:
Aha!  It's needed for the workspaces list.  The list changes if you add a new workspace, or every time one of the filtering buttons is clicked.

eddiebroad:
it seems that this should be merged!

PR #153, opened 2015-09-01
Some table bug fixes

MattPutnam:
The scope of this has changed quite a bit...

While poking around with this issue, I found some bugs:
* The workflows list table was not refreshing with the filter buttons.  This turned out to be another faulty react optimization, so I gave it the current filter as a key.
* There was some bad logic with one of the submission details tables.  That has been fixed.

All in all, the component-will-receive-props logic in the table is no longer needed.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -384,10 +384,6 @@
      (let [rows (react/call :get-filtered-data this)]
        (react/call :set-rows (@refs "body") (react/call :get-body-rows this rows))
        (react/call :set-num-rows-visible (@refs "paginator") (count rows))))
-   :component-will-receive-props
-   (fn [{:keys [props state refs]}]
-     (swap! state assoc :ordered-columns (create-ordered-columns (:columns props)))
-     (react/call :make-desynced (@refs "filterer")))
    :component-did-mount
    (fn [{:keys [this state]}]
      (set! (.-onMouseMoveHandler this)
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -384,10 +384,6 @@
      (let [rows (react/call :get-filtered-data this)]
        (react/call :set-rows (@refs "body") (react/call :get-body-rows this rows))
        (react/call :set-num-rows-visible (@refs "paginator") (count rows))))
-   :component-will-receive-props
-   (fn [{:keys [props state refs]}]
-     (swap! state assoc :ordered-columns (create-ordered-columns (:columns props)))
-     (react/call :make-desynced (@refs "filterer")))
```
dmohs: Maybe remove this method if it is no longer used.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
```diff
@@ -70,14 +70,8 @@
...
-       :columns [{:header "Data Entity" :starting-width 200 :sort-by :value
-                  :filter-by (fn [entity]
-                               (str (entity "entityType") " " (entity "entityName")))}
-                 {:header "Last Changed" :starting-width 280 :sort-by :value
-                  :content-renderer (fn [row-index date]
-                                      (let [m (js/moment date)]
-                                        (str (.format m "L [at] LTS") " ("
-                                          (.fromNow m) ")")))}
+       :columns [{:header "Data Entity" :starting-width 200 :sort-by :value}
+                 {:header "Last Changed" :starting-width 280 :sort-by :value}
```
dmohs: Will this sort chronologically?

dmohs:
:+1: with minor comments.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/submission_details.cljs
MattPutnam: Ack, you're right, it won't.  I'll revert.

PR #155, opened 2015-09-01
Focus work for dialogs

MattPutnam:
Dialogs can now specify three more properties:

* `:get-first-element-dom-node` -- When specified, this component is focused (and selected if it's a text component) when the dialog is shown.
* `:get-last-element-dom-node` -- When additionally specified, key handlers are added to the first and last items to constrain focus in the dialog.
* `:cycle-focus?` (defaults to false) -- When both first and last are specified, this controls whether or not the focus should cycle.  When true, tabbing from the last component focuses the first, and shift-tabbing from the first focuses the last.  When false, focus traversal is simply blocked on each end.


eddiebroad:
nice.  I loaded it up and experienced a NON-cycle focusing.

dmohs:
:+1: Love this.

PR #157, opened 2015-09-03
Fixed UI code to conform to new Rawls response.

MattPutnam:
Also fixed mock data, create workspace, and added a blocker during workspace creation.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -11,7 +11,7 @@
 (react/defc Spinner
   {:render
    (fn [{:keys [props]}]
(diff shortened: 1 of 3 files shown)


eddiebroad:
Have you tested this code running with RAWLS? If yes, did it run correctly?

MattPutnam:
It was working earlier today.  Looks like they're currently switching the domain names so this is broken again.

eddiebroad:
re: "It was working earlier today."  in that case, I think it can and perhaps should be merged.  Do you agree?

MattPutnam:
Sure.  We'll have to change some URLs but that's a separate issue.

eddiebroad:
re: "Sure" , okay, I plan to merge.  Do you explicitly confirm that you have no objection to that?

MattPutnam:
Yes

PR #158, opened 2015-09-03
back button addition in import a selected MC window

eddiebroad:
to return to main list of MCs

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -104,8 +104,10 @@
                      "Destination Namespace"
                      (style/create-text-field {:defaultValue selected-conf-namespace :ref "destinationNamespace"})}]
           (create-formatted-label-textfield k v))
-        (render-import-button props refs)]))})
-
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -151,6 +153,13 @@
       (cond
         (:show-import-mc-modal? @state) [ConfigurationImportForm {:selected-method-config (:selected-method-config @state)
                                                                   :workspace-id (:workspace-id props)
+                                                                  :on-back
+                                                                  (fn []
+                                                                    (swap! state dissoc
+                                                                      :show-import-mc-modal?
+                                                                      :selected-method-config)
+                                                                    (swap! state assoc
+                                                                      :loaded-import-confs? true))
```
MattPutnam: We don't need to much with `:loaded-import-confs?` in order to do this window switch.  If you get rid of this swap and also the `:loaded-import-confs? false` a few lines below, this still works.  Really, what we're switching on is `:show-import-mc-modal?` (which is poorly named now); whether or not the import confs have been loaded doesn't change.

MattPutnam:
One thing but otherwise :+1:

PR #159, opened 2015-09-03
Cleanup, related to DSDEEPB-889

MattPutnam:


eddiebroad:
Have you been able to test/run this with live data?  Has RAWLS/AGORA been up recently?

MattPutnam:
No, I haven't been able to test with live data.  Nothing here is really dependent on real data though.

eddiebroad:
okay

PR #162, opened 2015-09-04
enter the method-config table from repo for browsing

eddiebroad:
This commit/PR is largely for modifying the user-facing UI.
If this PR is merged, then a next/todo PR/work is to further tweak
the "Import" window (which would be called the 
 "import and browse" window I guess) for browsing
 MCs and not just importing them.
Here "tweak" means additional logic so that the table won't display
MCs for any method except for those of the selected method.
It also means to *perhaps* modify the code so that the user can import an MC from
having entered the screen via the method-repo.
Additional tweak may be necessary in any event especially because AGORA is updating
the method-config schema

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
@@ -5,23 +5,41 @@
    [org.broadinstitute.firecloud-ui.common.style :as style]
    [org.broadinstitute.firecloud-ui.common.table :as table]
    [org.broadinstitute.firecloud-ui.paths :refer [get-methods-path]]
+   [org.broadinstitute.firecloud-ui.page.workspace.method-config-importer :as importmc]
    [org.broadinstitute.firecloud-ui.utils :as utils]
    ))
 
 
 (react/defc MethodsList
   {:render
-   (fn [{:keys [props]}]
+   (fn [{:keys [props state]}]
+     (if (:show-import-overlay? @state)
+       [comps/Dialog
+        {:blocking? true
+         :width "80%"
+         :dismiss-self #(swap! state dissoc :show-import-overlay?)
+         :content  (importmc/render-import-overlay
+                     nil ; nil workspace because entrypoint is  from method repo
+                     #(swap! state dissoc :show-import-overlay?) ; on-close
+                     nil ; on-import nil because entry from method re
+                     (:selected-method @state))}]
      [table/Table
       {:empty-message "No methods to display."
        :columns [{:header "Namespace" :starting-width 150 :sort-by :value}
-                 {:header "Name" :starting-width 150 :sort-by :value}
+                 {:header "Name" :starting-width 150 :sort-by :value
+                  :content-renderer (fn [row-index msm]
+                                      [:a {:style {:color (:button-blue style/colors)
+                                                   :textDecoration "none"}
+                                           :href "javascript:;"
+                                           :onClick #(swap! state assoc :show-import-overlay? true
+                                                      :selected-method (str msm))}
+                                       (str msm)])}
                  {:header "Synopsis" :starting-width 500 :sort-by :value}]
        :data (map (fn [m]
                     [(m "namespace")
                      (m "name")
                      (m "synopsis")])
-               (:methods props))}])})
+               (:methods props))}]))})
 
 
 (defn- create-mock-methods []
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -76,21 +76,29 @@
            on-import (:on-import props)]
        (assert selected-config (str "Missing a selected configuration: " props))
        (assert on-import (str "Missing on-import handler: " props))
-       (assert workspace-id (str "Missing workspace-id: " props))
        [:div {}
-        (create-formatted-header "Import Method Configuration")
+        [:div {:style {:position "absolute" :left 2 :top 2}}
+         [comps/Button {:icon :angle-left
+                        :onClick #((:on-back props))}]]
+        (create-formatted-header
+          (if (:workspace-id props)
+            "Import Method Configuration"
+            "Browse Method Configuration"))
         (for [[k v] {"Name" selected-conf-name
                      "Namespace" selected-conf-namespace
                      "Snapshot Id" selected-conf-snapshot-id}]
           (create-formatted-label-text k v))
-        (for [[k v] {"Destination Name"
-                     (style/create-text-field {:defaultValue selected-conf-name :ref "destinationName"})
-                     "Destination Namespace"
-                     (style/create-text-field {:defaultValue selected-conf-namespace :ref "destinationNamespace"})}]
-          (create-formatted-label-textfield k v))
-        (render-import-button props refs)
-        [:span {:style {:marginLeft "0.5em"}}
-         [comps/Button {:text "Back" :onClick #((:on-back props))}]]]))})
+        (when (:workspace-id props)
+          [:div {}
+           (for [[k v] {"Destination Name"
+                        (style/create-text-field {:defaultValue selected-conf-name
+                                                  :ref "destinationName"})
+                        "Destination Namespace"
+                        (style/create-text-field {:defaultValue selected-conf-namespace
+                                                  :ref "destinationNamespace"})}]
+             (create-formatted-label-textfield k v))
+           (render-import-button props refs)])
+        [:span {:style {:marginLeft (when (:workspace-id props) "0.5em")}}]]))})
 
 (react/defc ConfigurationsTable
   {:render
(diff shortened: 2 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -150,12 +161,16 @@
                       (swap! state assoc :error-message status-text))
         :mock-data (create-mock-methodconfs-import)}))})
 
-(defn render-import-overlay [workspace-id on-close on-import]
+(defn render-import-overlay [workspace-id on-close on-import selected-method]
```
MattPutnam: There's now the wrong number of arguments being passed in from method_configs_tab

src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
```diff
@@ -5,23 +5,42 @@
...
    ))
 
 
 (react/defc MethodsList
   {:render
-   (fn [{:keys [props]}]
+   (fn [{:keys [props state]}]
+     (if (:show-import-overlay? @state)
+       [comps/Dialog
+        {:blocker? true
```
MattPutnam: Remove this line, this isn't a property of comps/Dialog

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -76,21 +76,25 @@
            on-import (:on-import props)]
        (assert selected-config (str "Missing a selected configuration: " props))
        (assert on-import (str "Missing on-import handler: " props))
-       (assert workspace-id (str "Missing workspace-id: " props))
        [:div {}
-        (create-formatted-header "Import Method Configuration")
+        (create-formatted-header
+          (if (:workspace-id props)
+          "Import Method Configuration"
+          "Browse Method Configuration"))
```
MattPutnam: Nit: indentation

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -76,21 +76,25 @@
...
         (for [[k v] {"Name" selected-conf-name
                      "Namespace" selected-conf-namespace
                      "Snapshot Id" selected-conf-snapshot-id}]
           (create-formatted-label-text k v))
+        (when (:workspace-id props)
```
MattPutnam: Indentation (the chunk of code inside this `when`)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -76,21 +76,25 @@
...
         (for [[k v] {"Destination Name"
                      (style/create-text-field {:defaultValue selected-conf-name :ref "destinationName"})
                      "Destination Namespace"
                      (style/create-text-field {:defaultValue selected-conf-namespace :ref "destinationNamespace"})}]
           (create-formatted-label-textfield k v))
-        (render-import-button props refs)
-        [:span {:style {:marginLeft "0.5em"}}
-         [comps/Button {:text "Back" :onClick #((:on-back props))}]]]))})
+          (render-import-button props refs))
+        [:span {:style {:marginLeft (if (:workspace-id props)"0.5em" "0.0em")}}
```
MattPutnam: You can just say `(when (:workspace-id props) "0.5em")`

Also, I really don't like the placement of this back button.  It's completely out of place where it is.  Maybe change it to an icon button with the "<" icon, and place it in the top left corner?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -76,21 +76,25 @@
...
+         [comps/Button {:text "Back" :onClick #((:on-back props))}]]
+        ]))})
```
MattPutnam: Merge with previous line

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -100,7 +104,10 @@
        (assert on-config-selected (str "Missing an on-selected-config handler: " props))
        (assert method-configs (str "Missing a list of method configs: " props))
        [:div {}
-        (create-formatted-header "Select A Method Configuration For Import")
+        (create-formatted-header
+          (if (:selected-method props)
+            (str "Browse Method Configurations for " (:selected-method props)  )
+          "Select A Method Configuration For Import"))
```
MattPutnam: Indentation

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -150,12 +161,16 @@
...
      [:div {:style {:position "absolute" :right 2 :top 2}}
       [comps/Button {:icon :x :onClick #(on-close)}]]
      [:div {:style {:backgroundColor "#fff" :borderBottom (str "1px solid " (:line-gray style/colors))
                     :padding "20px 48px 18px"}}
-      [ModalPage {:workspace-id workspace-id :on-close on-close :on-import on-import}]
+      [ModalPage {:workspace-id workspace-id
+                  :on-close on-close
+                  :on-import on-import
+                  :selected-method selected-method
+                  }]
```
MattPutnam: Merge with previous line.

MattPutnam:
General suggestion: Instead of passing 4 separate arguments to `method-config-importer/render-import-overlay`, with some of them being `nil` sometimes, pass them as a map.  `method-config-importer/ModalPage` is already taking them as a props map, and this helps name the arguments so they don't get crossed.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
eddiebroad: One idea is to put "nil" in the position to fix the number.

eddiebroad:
re "General suggestion: Instead of passing 4 separate arguments..."
I had thought to use a map in this way...but I was thinking it's better to prioritize first what is desired (for example the back button you mention) and the implementation (code for a selected UI) of what is desired before its code is refactored.  An objective I imagined this PR having is requirements analysis and discussion for the UI and some discussion and analysis has indeed taken place (such as the "micro-discussion" on the backbutton and its placement.  What do you think?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
eddiebroad: Re : "Maybe change it to an icon button with the "<" icon, and place it in the top left corner?", I will try that out and view it in a developmental GUI environment

eddiebroad: re " I really don't like the placement of this back button. "
What is the "X" button in the upper right is modified to have *two* functions.  The first function, done when viewing the table, is leaving the window altogether.  The second function, when viewing a selected MC (either for browse or for import)  is to have the same functionality as a back button.  The "flip" side of this coin is that the window would necessitate *two* clicks at most to leave it (either import or a two-step back-out).  What do you think?

MattPutnam: I think "go back to the previous screen" and "cancel this interaction" are distinct ideas.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
MattPutnam: Right.  It just needs to be there.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
eddiebroad: re : "I think "go back to the previous screen" and "cancel this interaction" are distinct ideas."  My interpretation of this is that the scenario of a dual-functioning X is "over-over-loading" the button and would be confusing.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -76,21 +76,29 @@
            on-import (:on-import props)]
        (assert selected-config (str "Missing a selected configuration: " props))
        (assert on-import (str "Missing on-import handler: " props))
-       (assert workspace-id (str "Missing workspace-id: " props))
        [:div {}
-        (create-formatted-header "Import Method Configuration")
+        [comps/Button {:icon :angle-left
```
MattPutnam: Use absolute positioning to put it in the corner of the whole dialog, and not push the rest of the content down.

Another spot for it might be 50% of the way down the left side.

eddiebroad:
Re: "Use absolute positioning ...."  I added a div for adjusting the abs. positioning.  Hopefully the button fits snug in the new position.

MattPutnam:
Cool.  Check indentation on that code and get rid of the commented out line, and I say this is good to go.

PR #163, opened 2015-09-08
Overhaul of AJAX calling

MattPutnam:
* Removed mock data from UI code.  This helps shorten the AJAX calls.
* endpoints.cljs makes it easy to inspect our backend calls, making it easier to react to API changes, and to add new endpoints as well.
* Structure allows for future enhancements: (1) Lazily generating mock data that's internally consistent, (2) Validating real data against the mock structure and warning if fields have been added or removed

I also picked up a change to the get-individual-workspace call, so now the summary tab is correct


dmohs:
I like this, but I still am not clear on the benefit of utils/call-ajax-orch rather than, e.g., (endpoints/foo arg1 arg2 on-done)

MattPutnam:
I haven't thought about it that hard, but I suspect that keeping the static data separate from its usage will make it easier to add the future enhancements I mentioned.  Or maybe not, I dunno.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -153,15 +153,14 @@
...
-(defn call-ajax-orch [path arg-map]
-  (ajax-orch path (assoc arg-map
-                    :on-done (fn [{:keys [success? xhr] :as map}]
-                               (if success?
-                                 ((:on-success arg-map) (merge map {:parsed-response (parse-json-string (.-responseText xhr))}))
-                                 ((:on-failure arg-map) (merge map {:status-text (.-statusText xhr)}))))
-                    :canned-response {:status 200 :delay-ms (rand-int 2000)
-                                      :responseText (if-let [mock-data (:mock-data arg-map)]
-                                                      (->json-string mock-data))})))
+(defn call-ajax-orch [{:keys [endpoint] :as arg-map}]
```
dmohs: This function is specific to an endpoint now, so I think it should live in the endpoints namespace.

dmohs:
:+1: Looks good. One suggestion.

PR #164, opened 2015-09-08
filter as you type

MattPutnam:
Still up for debate whether or not we actually want this functionality.  I don't really like it.


eddiebroad:
What does it mean "filter as you type"?  I hit keys, letter-by-letter, but had to press enter in order to observe filtering

MattPutnam:
Clean your build.  Enter doesn't even do anything any more.

dmohs:
:+1: Do you have a sense about how many rows this can support before performance becomes an issue?

MattPutnam:
It seems to get noticeably laggy around 5000 entries on the method configs tab.  Each table is going to be a bit different though.

dmohs:
Thanks for the performance data. Let's discuss this with Kristian before merging.

davidangb:
I spoke to Kristian about this. Let's push this out past FireCloud v1 MVP.

PR #166, opened 2015-09-09
add blocker for import button

eddiebroad:
per DSDEEPB-1160 comment

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -21,7 +21,7 @@
 (defn- create-formatted-header [text]
   [:div {:style {:fontSize 24 :align "center" :textAlign "center" :paddingBottom "0.5em"}} text])
 
-(defn- render-import-button [props refs]
+(defn- render-import-button [props refs state]
   (let [selected-config (:selected-method-config props)
         selected-conf-name (selected-config "name")
         selected-conf-namespace (selected-config "namespace")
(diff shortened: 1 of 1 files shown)


MattPutnam:
I think in this case it's okay to just pass state.  We're already passing props and refs.  This isn't a "shared state" thing like previously discussed, the `render-import-button` function is merely a convenience to keep the code clear.

eddiebroad:
Re: "I think in this case..." I passed and used set! with ":importing?" via that.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -30,24 +30,27 @@
...
-                              "destinationNamespace" dest-conf-namespace
-                              "destinationName" dest-conf-name}
-                    :on-done (fn [{:keys [success? xhr]}]
-                               (if success?
-                                 (on-import {"name" dest-conf-name "namespace" dest-conf-namespace})
-                                 (js/alert (str "Import Error: " (.-responseText xhr)))))}))}]))
+      :onClick (fn []
+                 (let [dest-conf-name (-> (@refs "destinationName") .getDOMNode .-value)
+                       dest-conf-namespace (-> (@refs "destinationNamespace") .getDOMNode .-value)]
+                   ((fn [] (swap! state assoc :importing? true)))
```
MattPutnam: Why are you creating a function and then immediately calling it?

eddiebroad: to achieve the desired function with the passed-in state.
Would it be preferred to use a  let binding here?

MattPutnam: No... just write `(swap! ...)`

eddiebroad:
Re : "No... just write (swap! ...)"  Now, it's called more directly with less "wrapper" overhead.

PR #167, opened 2015-09-09
Added filter buttons to workflows table

MattPutnam:
This should complete everything in DSDEEPB-841 that is currently available.


dmohs:
:+1: Code looks good. UI looks good. I could only test with mock data, but I will assume it works with real data.

PR #168, opened 2015-09-09
Use root entity type in launch submission

MattPutnam:


dmohs:
FYI: I think we have been told to not merge this one.

MattPutnam:
Yeah, I saw the discussion.

Still, maybe we can pre-select the root entity type in the combo.

PR #169, opened 2015-09-10
View and edit workspace ACL from UI

MattPutnam:
Note: this depends on the matching orchestration PR.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
@@ -62,6 +62,20 @@
                                :lastFailureDate (rand-nth [nil (utils/rand-recent-time)])}
     :owners ["test@broadinstitute.org"]}})
 
+(defn get-workspace-acl [workspace-id]
+  {:path (str "/workspaces/" (ws-path workspace-id) "/acl")
+   :method :get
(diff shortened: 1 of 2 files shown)


MattPutnam:
Actually, now that I look at this closer there's some screwy behavior.  But I want to add this functionality sooner rather than later for dev purposes, so I'll keep this PR open.

eddiebroad:
I loaded this with but then saw "Loading ACL..." and the spinner never went away....  Is that the "screwy" behavior that you're talking about?

MattPutnam:
Good call on ACL vs. Permissions.  Will change to "permissions"

It should be working.  Try cleaning your build?

MattPutnam:
See 2nd commit for updates.  Previously you could edit the text of an existing ACL entry which did weird things, and I wasn't handling the case of adding a new row but leaving it empty.  I'm much happier about this now.

MattPutnam:
Update: Orchestration PR is in so this is good to go

eddiebroad:
@MattPutnam I cleaned my build but still see the spinner and it doesn't go away.  That's with use-live-data=false .....  should I not use mock data?  Has anyone else gotten past the spinner I wonder...

dmohs:
:+1: Code looks good. Works great! So awesome!

PR #172, opened 2015-09-11
adding checking for a code and message which may sometimes come from

eddiebroad:
AGORA

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
@@ -56,6 +56,12 @@
        {:endpoint endpoints/list-methods
         :on-done (fn [{:keys [success? get-parsed-response status-text]}]
                    (if success?
-                     (swap! state assoc :methods (get-parsed-response))
+                     (let [parsed-response-text (get-parsed-response)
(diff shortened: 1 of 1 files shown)


MattPutnam:
Is this really a bug in Agora?  Should Agora be returning a more appropriate response code?

eddiebroad:
@MattPutnam you have a good point.  When I issued/created this PR it was specifically with DSDEEPB-1239 in mind.  I wonder if @davidangb  would say that the subtask dsdeepb-1239 should be something for silver to address or for Agora to address.

dmohs:
Agree with @MattPutnam. Our services have standardized on returning a 400+ HTTP status code on error. Let's close this PR once a bug has been filed against Agora.

davidangb:
we need to isolate whether or not this is a bug in Agora or in orchestration layer. It is possible (I have no proof one way or another) that agora returns a 500 but orchestration layer returns 200.

eddiebroad:
image

eddiebroad:
I posted one case where an agora bug is seemingly manifested

davidangb:
I discussed with Brad - they consider it a bug in agora if they return an http 200 with an error, and they believe they've fixed it (at least in dev). We should not add any special handling, which means we can close this. I'm closing it now.

PR #173, opened 2015-09-11
Fix data import issue

MattPutnam:
It was a problem with the way in which the payload was encoded.  In all other cases we're converting a clojurescript data structure to JSON, so I was doing that by default.  But here we don't want any conversion.  So I added a mechanism to specify how to encode the payload, defaulting to JSON.

Also shoehorning in the "Creating Workflow..." cosmetic bug that Greg noticed since this is a small PR.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
@@ -6,12 +6,19 @@
 
 (defn call-ajax-orch [{:keys [endpoint] :as arg-map}]
   (utils/ajax-orch (:path endpoint)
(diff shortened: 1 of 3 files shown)


MattPutnam:
Correct.  The `nil` case is the default, i.e. no encoding specified.

dmohs:
Another option for this: The caller specifies `:data` for raw data (which is passed through unchanged) or `:payload` for the thing that will get converted to JSON. Could even rename `:payload` to `:cljs-data`. Good to merge as-is or with this suggested change.

MattPutnam:
Ooh, I like that idea better.

dmohs:
Me too :)

PR #175, opened 2015-09-11
for adding workspace fields to import

eddiebroad:
an MC into a workspace when entering from the
Method Repo.

[this is ready for review now]

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo.cljs
@@ -20,7 +20,7 @@
          :content  (importmc/render-import-overlay
                      nil ; nil workspace because entrypoint is  from method repo
                      #(swap! state dissoc :show-import-overlay?) ; on-close
-                     nil ; on-import nil because entry from method re
+                     #(swap! state dissoc :show-import-overlay?) ; on-import because entry from method repo
                      (:selected-method @state))}]
      [table/Table
       {:empty-message "No methods to display."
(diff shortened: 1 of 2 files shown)


davidangb:
When coming from the method repo (i.e. outside of a workspace), I'm having a few issues:
1) The title of the modal says "Export Method Configuration to Workspace," but the submit button still says "Import."
2) If the user leaves the destination workspace name or destination workspace namespace blank, we throw an ugly JS alert error; we could/should protect better against that.
3) If the user populates the destination workspace name/namespace, but those are invalid for whatever reason (e.g. the namespace doesn't exist), we also throw an ugly JS alert error.

eddiebroad:
@davidangb 

Re: "1) The title of the modal says...." , okay, I'll modify the button to say "Export" instead of "Import"
Re: "2) If the user leaves ...." I'll add some basic validation for the two fields and have something like "You must provide a destination."
Re: "3)  If the user populates..." , I can tidy up the exception message when I address #1 and #2.
What do you think?

eddiebroad:
@davidangb Re "3) If the user populates the destination .....", regarding this we propagate errors from RAWLS and AGORA.... should we ask them for better errors messages?

davidangb:
regarding the error messages - I'm not talking about the wording/content. I just don't like using a js alert, when we could/should instead display something in the UI.

eddiebroad:
@davidangb Re: "regarding the error message - ...."  @dmohs mentioned that all the error messages come up as js/alerts.  He and talked about how the js/alert response isn't the most ideal, but that it perhaps *might* suffice for now.  He and I also said that in the long run, we might have a standardized list of error codes and how, the UI might do something like put a red start by a text-field or something similar to more "cleanly" alert users to errors they can fix.

davidangb:
okay, I'm going to merge as-is. What exists is functional; I think we have a ways left to go on the UX side of things, but as discussed we'll break that out in to follow-on stories.

PR #177, opened 2015-09-14
auto-populate entity type combo with root entity type

MattPutnam:
Although we're not restricted to just the root entity type, we can at least pre-select it.
Unable to test with real data at the moment.


dmohs:
:+1:

PR #178, opened 2015-09-14
Add delete workspace functionality

MattPutnam:
* Delete button now appears on summary tab if you have OWNER status.  After delete you get taken back to the workspaces list.
* Generalized the buttons and status labels that appear in the left column of several pages.
* Fixed an issue with ACL mock data


dmohs:
Tested this. Works great. Nice component refactoring too. :+1:

PR #179, opened 2015-09-15
rearrange layout to be more columnar (src on left ; dest on right) in

eddiebroad:
the import/export MC screen.
Add refs to labels.
Add tempred function for making style color green for corresponding label
if error detected during a basic validation

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
@@ -41,10 +41,8 @@
                          (:workspace-id props))
                        dest-conf-name (-> (@refs "destinationName") .getDOMNode .-value)
                        dest-conf-namespace (-> (@refs "destinationNamespace") .getDOMNode .-value)
-                       dest-ws-n-len  (.-length (:name workspace-id))
-                       dest-ws-ns-len (.-length (:namespace workspace-id))
-                       n-basic-valid (if (>= dest-ws-n-len 1) true false)
-                       ns-basic-valid (if (>= dest-ws-ns-len 1) true false)]
+                       n-basic-valid (if (zero? (count (:name workspace-id))) false true)
+                       ns-basic-valid (if (zero? (count (:namespace workspace-id))) false true)]
                    (if-not n-basic-valid
                      (setrefcolor "wnref" refs "red")
                      (if-not ns-basic-valid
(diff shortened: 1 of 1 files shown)


eddiebroad:
this error notification is basic.  It Might work for now, but it seems we need something more polished

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_importer.cljs
```diff
@@ -43,9 +51,9 @@
                        n-basic-valid (if (>= dest-ws-n-len 1) true false)
                        ns-basic-valid (if (>= dest-ws-ns-len 1) true false)]
                    (if-not n-basic-valid
```
MattPutnam: This line can be replaced with `(if (zero? (count (:name workspace-id)))` (and similarly for the other check), and then you can get rid of 4 lines in the `let`.

MattPutnam:
I think this is a reasonable solution for now.  I've been thinking about creating a "validatable text field" component that would replace this (and a lot of others), but haven't gotten around to it.

Format to get rid of some extraneous whitespace and this gets a :+1: (unofficial review) from me.

davidangb:
from a UX perspective, I don't like this approach. The validation messages are time-sensitive; if you attempt to import/export a config, look away for a moment, and the import/export failed, you will have no idea what happened. I'd prefer to keep the js alerts until we have time to think through the proper UX for this.

I'm fine with rearranging the columns.

MattPutnam:
The better thing to do would be to set the color to red indefinitely, and change it back when the user types something into the text field.  This would take a bit more work though.

eddiebroad:
@davidangb I commited code for column rearrangement based on a comment from Chet

eddiebroad:
@MattPutnam Re: "The better thing to do...." I committed for that (05bf9....)

davidangb:
So, I have no real problem with this code ... I just don't think it's worth it. The js alert errors are basic and simple and functional. This change feels like it's a partial step towards a better UI, but it incurs technical debt. For instance, I'd rather see the label coloring be completely dynamic:
* upon first loading, labels would be red because their values are empty
* as soon as a user enters a value, labels become black
* as soon as a user erases the value (backspace/cut/whatever), they become red again

Implementation-wise, this should be a purely dynamic color setting based on the value, not needing to call a separate function.

Bottom line, I think we should postpone these changes until we've had a chance to figure out the right approach.

PR #182, opened 2015-09-16
make the import/export selected MC screen more tabular

eddiebroad:
src on left  and destination on right
motivation for this is based on comments from Chet


eddiebroad:
this PR has no change in error correction/detection.  It's only changes are in the layout of the UI/GUI elements (input fields and labels) via div layout modifications

PR #184, opened 2015-09-17
View and edit workspace attributes

MatthewBemis:
This is the UI work associated with [DSDEEPB-1137](https://broadinstitute.atlassian.net/browse/DSDEEPB-1137) and firecloud-orchestration [PR #77](https://github.com/broadinstitute/firecloud-orchestration/pull/77).

This is functionally complete and ready for review- you can view/edit/add/delete attributes on workspaces.

I'd like some feedback on how to solve the following problems:

~~A small bug that doesn't really come up often:
1) Delete an attribute that already exists (i.e. “Attribute1”)
2) Add new row, and put “Attribute1” into name field
3) Add new row again, and the field with "Attribute1" in it becomes un-editable 
4) Right now, If the user does need to edit it after encountering this, they can just delete that row and enter that info again and it will be fine~~

~~Another bug which is more of an annoyance rather than a problem:
Adding more than 9 rows causes the new rows to spawn in seemingly random locations. I'm assuming that this has to do with the fact that I'm using a map for my attribute data, and the docs seem to say that when adding pairs to a map, they aren't appended at the end but rather added anywhere within the map. Any suggestions?~~

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
@@ -71,6 +71,54 @@
                               (react/call :capture-ui-state this)
                               (react/call :persist-acl this))}]}])
 
+(defn- create-section [& children]
+  [:div {:style {:padding "1em 0 2em 0"}} children])
+
+(defn- reload-attributes [props state on-success]
+  (endpoints/call-ajax-orch
+    {:endpoint (endpoints/get-workspace (:workspace-id props))
+     :on-done  (fn [{:keys [success? get-parsed-response status-text]}]
+                 (if success?
+                   (let [response (get-parsed-response)
+                         ws (response "workspace")
+                         attrs-list (mapv (fn [[k v]] [k v]) (ws "attributes"))]
+                     (swap! state assoc :server-response
+                       {:attrs-list attrs-list} :loaded-attrs? true :saving? true)
+                     (on-success))
+                   (swap! state assoc :server-response {:error-message status-text})))}))
+
+(defn- add-update-attributes [props attrs on-success]
+  (let [workspace-id (:workspace-id props)
+        add-update-ops (for [pair attrs]
+                         {:op "AddUpdateAttribute"
+                          :attributeName (key pair)
+                          :addUpdateAttribute (val pair)})]
+    (endpoints/call-ajax-orch
+      {:endpoint (endpoints/update-workspace-attrs workspace-id)
+       :method :PATCH
+       :payload add-update-ops
+       :headers {"Content-Type" "application/json"}
+       :on-done (fn [{:keys [success? xhr]}]
+                  (if-not success?
+                    (js/alert (str "Exception:\n" (.-statusText xhr)))
+                    (on-success)))
+       :canned-response {:status 200 :delay-ms (rand-int 2000)}})))
+
+(defn- delete-attribute [props attr on-success]
+  (let [workspace-id (:workspace-id props)
+        delete-op [{:op "RemoveAttribute"
+                    :attributeName attr}]]
+    (endpoints/call-ajax-orch
+      {:endpoint (endpoints/update-workspace-attrs workspace-id)
+       :method :PATCH
+       :payload delete-op
+       :headers {"Content-Type" "application/json"}
+       :on-done (fn [{:keys [success? xhr]}]
+                  (if-not success?
+                    (js/alert (str "Exception:\n" (.-statusText xhr)))
+                    (on-success)))
+       :canned-response {:status 200 :delay-ms (rand-int 2000)}})))
+
 (react/defc AclEditor
   {:render
    (fn [{:keys [props state this]}]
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -91,7 +91,7 @@
 ;;
 
 
-(def use-live-data? true)
+(def use-live-data? false)
```
MatthewBemis: I'll change this back before merge :)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -24,177 +26,321 @@
...
+       (when owner?
+         [comps/SidebarButton {:style :light :margin :top :color :exception-red
+                               :text "Delete" :icon :trash-can
+                               :onClick #(when (js/confirm "Are you sure?")
+                                          (swap! state assoc :deleting? true)
+                                          (react/call :delete this))}])])
+
+    (when (and viewing-attributes? (not editing-attributes?))
+      [:div {}
+       (when writer?
```
MatthewBemis: Is writer the correct access level to require for editing workspace attributes? Or is this something for owners?

MatthewBemis:
@dmohs This is just about ready for re-review, but I have one more question. Right now I'm trying to disable textfields for names of attributes that already exist in the workspace (because rawls doesn't directly support renaming). I can't seem to figure out a good way to disable the textfields without creating some other problems.

Right now I grab the attributes from the workspace itself, but this isn't going to be a complete list once attributes are added, because we don't make an ajax call to get the workspace again (I could do that, but it seems unnecessary since we have the information we need in the state). So the only reserved attribute names are the ones that were grabbed during the initial get-workspace call.

So I figured that the better option would be to grab the attributes from the state, since that's updated live. The problem is that we're updating the state with every keystroke, so if somebody were to save an attribute with the name "r", this would cause the textbox to be disabled as soon as the user started to type the first character in "readme". So that's not very functional.

Any suggestions?

MatthewBemis:
@dmohs The refactoring should be complete! All of the concerns that I had with the old implementation are fixed in this new one. It appears to work on live data as well.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -29,6 +29,52 @@
...
+                   (if success? {:workspace (get-parsed-response)}
+                                {:error-message status-text}))
+                 (let [ws ((get-in @state [:server-response :workspace]) "workspace")]
+                   (swap! state assoc :attrs-list (mapv (fn [[k v]] [k v]) (ws "attributes")))))})
+  )
+
+(defn- add-update-attributes [props attrs]
+  (let [workspace-id (:workspace-id props)
+        add-update-ops (for [pair attrs]
+                         {:op                 "AddUpdateAttribute"
```
dmohs: Don't line these up—it's painful to maintain this. Just one space after keyword.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -29,6 +29,52 @@
...
+                          :attributeName      (key pair)
+                          :addUpdateAttribute (val pair)
+                          })]
```
dmohs: Join this with previous line. Please fix all occurrences.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -115,8 +161,133 @@
...
+           [:div {}
+            [comps/SidebarButton {:style   :light :margin :bottom :color :button-blue
+                                  :text    "View summary" :icon :document
+                                  :onClick on-done
+                                  }]
+            (when writer?
+              [comps/SidebarButton {:style   :light :color :button-blue
+                                    :text    "Edit attributes" :icon :pencil
+                                    :onClick #(do (swap! state assoc :editing? true)
+                                                  (if-not (:attrs-list @state) (swap! state assoc :attrs-list attrs-list))
```
dmohs: 100 column limit in our style guide. Please fix all.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -115,8 +161,133 @@
...
+                       :x)]
+                    (common/clear-both)])
+                 (if-not (:attrs-list @state) (reload-attributes props state) (:attrs-list @state)))
+               ;TODO: new textfields should gain typing focus when adding new rows
+               [:div {:style {:display (when-not (:editing? @state) "none")}}
+                [comps/Button {:style   :add :text "Add new"
+                               :onClick #(do (swap! state update-in [:attrs-list] conj ["" ""]))
+                               }]]])
+            (style/create-paragraph [:em {} "There are no attributes to display"]))]]
+        [:div {:style {:clear "both"}}]]))})
```
dmohs: This is such a common patter that we have `(common/clear-both)`.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -126,68 +297,33 @@
...
    :delete
    (fn [{:keys [props state]}]
      (endpoints/call-ajax-orch
        {:endpoint (endpoints/delete-workspace (:workspace-id props))
-        :on-done (fn [{:keys [success? status-text]}]
-                   (swap! state dissoc :deleting?)
-                   (if success?
-                     ((:on-delete props))
-                     (js/alert (str "Error on delete: " status-text))))}))
+        :on-done  (fn [{:keys [success? status-text]}]
```
dmohs: Accidental change?

dmohs:
This is really good. Please just fix to match our style conventions and rebase. I'll take one last look then.

MatthewBemis:
@dmohs comments have been addressed and branch has been rebased and squashed

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -126,59 +311,24 @@
...
-           (style/create-section-header "Tags")
-           (style/create-paragraph (render-tags ["Fake" "Tag" "Placeholders"]))
-           (style/create-section-header "Research Purpose")
-           (style/create-paragraph [:em {} "Research purpose not available yet"])
-           (style/create-section-header "Billing Account")
-           (style/create-paragraph [:em {} "Billing account not available yet"])]
-          [:div {:style {:clear "both"}}]])))
+         (if (:viewing-attributes? @state)
+           (let [ws-response ((get-in @state [:server-response :workspace]) "workspace")]
+             [AttributeViewer {:ws           ws :writer? writer? :on-done #(swap! state dissoc :viewing-attributes?)
```
dmohs: Missed these ones.

dmohs:
UI Review:
1) If I edit and then cancel, my edits are left in place.
2) All AJAX calls should show a spinner while they are in-flight.
3) If I add an attribute, the delete button doesn't work on it while I'm in editing mode.
4) I can save with an empty attribute.

MatthewBemis:
I can't seem to reproduce #3. I'll fix the others in the meantime, though

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -118,8 +165,150 @@
...
+                               {:backgroundColor "#fff"})})
+                   (icons/font-icon
+                     {:style {:paddingLeft "0.5em" :padding "1em 0.7em"
+                              :color "red" :cursor "pointer"
+                              :display (when-not (:editing? @state) "none")}
+                      :onClick (fn [e] (do (when (contains? (:reserved-keys @state) i)
+                                      (delete-attribute props (first a) #(swap! state assoc :deleting? false))
+                                      (swap! state update-in [:reserved-keys] utils/delete i)
+                                      (swap! state assoc :deleting? true))
+                                    (swap! state update-in [:attrs-list] utils/delete i)))}
```
dmohs: Reformat this function and only update state once (not tested):
```Clojure
(fn [e]
  (if (contains? (:reserved-keys @state) i)
    (do
      (swap! state #(assoc :deleting? true))
      (delete-attribute props (first a)
                        #(swap! state #(-> %
                                           (assoc :deleting? false)
                                           (update-in [:reserved-keys] utils/delete i)
                                           (update-in [:attrs-list] utils/delete i)))))
    (swap! state update-in [:attrs-list] utils/delete i)))
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -118,8 +165,150 @@
...
+           :text "Done" :icon :status-done
+           :onClick (fn [e] (do (swap! state assoc :attrs-list
+                                  (filterv (fn [pair] (not (clojure.string/blank?
+                                                             (clojure.string/trim (first pair)))))
+                                    (:attrs-list @state)) :saving? true)
+                                ;TODO: rawls will soon return attributes after update- use that intead of reload
+                                (add-update-attributes props (:attrs-list @state)
+                                  (fn []
+                                    (swap! state assoc :editing? false :saving? false)
+                                    (reload-attributes props state #(swap! state assoc :saving? false))))))}]])])})
```
dmohs: Try this (not tested):
```Clojure
(fn [e]
  (swap! state assoc
         :attrs-list (filterv
                      (fn [pair] (not (clojure.string/blank? (clojure.string/trim (first pair)))))
                      (:attrs-list @state))
         :saving? true)
  ;; TODO: rawls will soon return attributes after update- use that intead of reload
  (add-update-attributes
   props (:attrs-list @state)
   #(reload-attributes props state #(swap! state assoc :editing? false :saving? false))))
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -129,60 +318,26 @@
...
-           (style/create-paragraph (render-tags ["Fake" "Tag" "Placeholders"]))
-           (style/create-section-header "Research Purpose")
-           (style/create-paragraph [:em {} "Research purpose not available yet"])
-           (style/create-section-header "Billing Account")
-           (style/create-paragraph [:em {} "Billing account not available yet"])]
-          (common/clear-both)])))
+         (if (:viewing-attributes? @state)
+           (let [ws-response ((get-in @state [:server-response :workspace]) "workspace")]
+             #_(react/call :load-workspace this)
+             (utils/rlog "rav")
```
dmohs: Remove these two commented lines.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -129,60 +318,26 @@
...
-        :on-done (fn [{:keys [success? get-parsed-response status-text]}]
-                   (swap! state assoc :server-response
-                     (if success? {:workspace (get-parsed-response)}
-                                  {:error-message status-text})))}))
+        :on-done  (fn [{:keys [success? get-parsed-response status-text]}]
+                    (swap! state assoc :server-response
+                      (if success? {:workspace (get-parsed-response)}
+                                   {:error-message status-text}))
+                    (let [ws ((get-in @state [:server-response :workspace]) "workspace")]
+                      (swap! state assoc :attrs-list (mapv (fn [[k v]] [k v]) (ws "attributes")))))}))
```
dmohs: Revert this change?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -118,8 +165,150 @@
...
+                                      (swap! state assoc :deleting? true))
+                                    (swap! state update-in [:attrs-list] utils/delete i)))}
+                     :x)]
+                  (common/clear-both)])
+               (:attrs-list @state))
+             ;TODO: new textfields should gain typing focus when adding new rows
+             [:div {:style {:display (when-not (:editing? @state) "none")}}
+              [comps/Button {:style :add :text "Add new"
+                             :onClick #(swap! state update-in [:attrs-list] conj ["" ""])}]]])
+          [:div {:style {:display (when (or (or (:editing? @state) (not-empty (:attrs-list @state))) (:saving? @state)) "none")}}
```
dmohs: 100-col limit.

dmohs:
:+1: Looks good. Minor suggestions. Go ahead and merge when done.

MatthewBemis:
Comments have been addressed, merging!

PR #187, opened 2015-09-21
add a small banner telling the viewer exactly which

eddiebroad:
workspace they're editing permissions for


eddiebroad:
the image below shows the small refinement viewed

image

eddiebroad:
the commit is intended as a reminder for the user.  Ideally they'd know such information already.
Sometimes a reminder can be useful to the user in case they've forgot and don't have to go back
to the main listing

PR #188, opened 2015-09-22
DSDEEPB-1348: Revise US Government warning notice on FireCloud login page

davidangb:
I was given the task of estimating stories in the backlog ... this one was so small that it was easier to fix than to estimate it and push it through the whole prediction process.

Note that beyond the verbiage change, I have modified the div to be a (hardcoded) 600px. Otherwise, on a large screen, the text is one big long line. I realize that hardcoded 600px is not responsive on devices.


rushtong:
:+1:

PR #189, opened 2015-09-22
DSDEEPB-32: add mailto: link for Support footer

davidangb:
Support link in the footer is now a mailto: to firecloud-help@broadinstitute.org, a newly-created Google Group.

Note that the mailto: link WILL ONLY WORK IF YOUR CHROME IS CONFIGURED WITH AN EMAIL HANDLER. See https://support.google.com/chrome/answer/1382847?hl=en. Try loading gmail, looking for the handler icon, and clicking on it.

I realize this restriction may not be tenable for the long term, but that's the MVP scope for this story, so I'm pushing it through.

We can edit members, permission, etc. of the Google Group in the future.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -27,6 +27,7 @@
               {:render
                (fn [{:keys [props state]}]
                  [:a {:href (:href props)
+                      :target (:target props)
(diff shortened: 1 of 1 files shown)


rushtong:
:+1: 
See http://blog.hubspot.com/marketing/set-gmail-as-browser-default-email-client-ht for allowing chrome/gmail to open mailto links.

PR #190, opened 2015-09-23
add :entityType to mock Configuration list to match AGORAs schema

eddiebroad:
image


davidangb:
I see that the mock data also includes "owner", but that's no longer returned by agora? Also remove owner?

eddiebroad:
@davidangb  Re: "I see that.....also includes "owner"...Also remove..."  Good catch!

PR #194, opened 2015-09-24
Config for NIH authentication

dmohs:
This is not meant to be our final implementation. The purpose of this PR is to reduce our risk by getting a functioning connection to NIH.


dmohs:
:+1: from Matt Nicholson

PR #196, opened 2015-09-24
Table cleanup and auditing of uses

MattPutnam:
* Added some "API documentation" to the table component as it's gotten a bit complicated
* Changed defaults so that columns sort by default, which can be disabled at the table level or on a column via `:sort-by :none`
* Removed row index as a parameter from `:content-renderer`, since it was never used
* Fixed some logic in `submission-details`


dmohs:
:+1: Nice boilerplate removal. I love the doc, but I fear it will get out-of-sync with the code. Created DSDEEPB-1390 to research that point.

PR #198, opened 2015-09-25
General cleanup and commonizing of things

MattPutnam:
Probably easier to look at individual commits

Welcoming suggestions on naming the new form component. That name is following a convention from my previous job, which I actually hate.


dmohs:
Thanks for the individual commits. Easy review. :+1:

PR #201, opened 2015-09-28
include workspace id when renaming a config

davidangb:
After closer inspection, our edit-config functionality (within a workspace) DID have a bug ... but the bug wasn't caused by the change in rawls' save-config API. We were simply failing to include the right info in our payload. This PR changes that.


dmohs:
:+1:

PR #202, opened 2015-09-28
Cleanup of method import/export code

MattPutnam:
Not 100% generalized, but since dealing with methods and configs involves different API calls this is about as far as it can go before it gets impossible to read.

* Created a method/config detail viewer that will hopefully be used for the method config editor later.  It operates on a fully loaded method/config from an Agora call.
* Added scroll-to-center utility for making sure the error message is visible.  It would be nice to add smooth scrolling in the future.
* Generalized the import form side of things to reduce code duplication
* Moved loading the workspaces to the import form


dmohs:
:+1: Very nice refactor. Squash if you want then merge.

PR #206, opened 2015-09-29
Added ellipses and did some column width touchup

MattPutnam:
I think I got all of them.


dmohs:
Nice. Doesn't have to be 100%. We just want to have a convention established.

PR #207, opened 2015-09-29
UI for single entity copy between workspaces

MattPutnam:
Rawls currently has a bug that prevents this from working with live data, but other than that this should be complete.


src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -311,6 +311,13 @@
         columns)))
 
 
+(defn date-column [props]
```
dmohs: Nice!

dmohs:
Looks good. Let's test when Rawls is back up.

PR #208, opened 2015-09-29
First pass at smooth scrolling

MattPutnam:
Needs some improvement, but works for now.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -46,18 +46,27 @@
           (+ start-y (* point (- end-y start-y))))
         (js/setTimeout #(animate start-time end-time start-x start-y end-x end-y) 10)))))
 
-(defn scroll-to [x y]
-  (let [start-time (js/Date.now)]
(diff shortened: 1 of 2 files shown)


rushtong:
This looks very cool! Would it be worth it to add an optional argument to scroll-to for the scroll time (currently defaults to 100)?

MattPutnam:
Good idea.  I'll add that.

src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -27,16 +27,46 @@
...
+      (.scrollTo js/window end-x end-y)
+      (let [point (smooth-step start-time end-time (js/Date.now))]
+        (.scrollTo js/window
+          (+ start-x (* point (- end-x start-x)))
+          (+ start-y (* point (- end-y start-y))))
+        (js/setTimeout #(animate start-time end-time start-x start-y end-x end-y) 10)))))
+
+(defn scroll-to
+  ([x y] (.scrollTo js/window x y))
+  ([x y duration]
```
dmohs: I suggest:
```Clojure
(assert (<= duration 400) "Duration too long (> 400ms)")
```
to ensure that the caller cannot accidentally create a bad experience for the user.

dmohs:
:+1: This is, of course, fully awesome. Merge whenever.

PR #211, opened 2015-10-01
eliminate methodRepoConfig

tedsharpe:
I see that there's another PR open to address this issue.  This is what I did.  Maybe we can compare and contrast and take one or the other?


eddiebroad:
I note that lines 120 in endpoints.cljs are different in the two PRs, so would that suggest that at least one (or both) of the PRs need to be rebased, once the rebase is done we can compare/contrast?

Line 120 PR 210
image

Line 120 PR 211
image

tedsharpe:
Rebased.

PR #212, opened 2015-10-01
UI work for workspace clone

MattPutnam:
Depends on corresponding Orchestration PR.
I don't like the bevy of buttons we're creating on this page.  Ideas?

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
@@ -280,11 +280,13 @@
                          (style/create-form-label "Namespace:")
                          (style/create-text-field {:ref "namespace"
                                                    :style {:width "100%"}
+                                                   :defaultValue (get-in props [:workspace-id :namespace])
                                                    :placeholder "Required"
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary_tab.cljs
```diff
@@ -271,6 +321,10 @@
      [AclEditor {:workspace-id (:workspace-id props)
                  :dismiss-self #(swap! state dissoc :editing-acl?)
                  :update-owners #(swap! state update-in [:server-response :workspace] assoc "owners" %)}])
+   (when (:cloning? @state)
+     (common/scroll-to-top 100)
```
dmohs: I don't think we'd want to scroll on every re-render of this component (even though it probably works right now). Suggest moving this to WorkspaceCloner:component-did-mount.

dmohs:
:+1: Good to merge with one suggestion.

MattPutnam:
Will do the scroll thing.  What do we think about pre-populating the fields with the values of the current workspace?  I feel like it would be common to either create a (renamed) copy in the current namespace, or copy to a new namespace but keep the old name.  But I don't really know.

PR #214, opened 2015-10-05
fixed bug when saving attribute with space in the name or value

MatthewBemis:
Saving an attribute with a space in it would throw a server exception from rawls and the UI would be stuck behind an "Updating..." blocker. Now that doesn't happen.


MattPutnam:
This `:saving?` flag is not being handled well.  But I'm going to clean up this file and break out the various editor components into separate files anyway, and I think that will get fixed as part of that, so this is fine for now.

PR #216, opened 2015-10-05
Parse GCS URIs and render them if they pass

MattPutnam:
I left the links unstyled, since they're actually external links this time.


eddiebroad:
:+1:

PR #218, opened 2015-10-06
Separated components into their own files

MattPutnam:
This is mostly just moving code, with some minor cleanup in the attribute viewer/editor


dmohs:
:+1: :+1: :+1: :+1: :+1:

PR #222, opened 2015-10-08
Lock/unlock workspaces

MattPutnam:
The button and service calls for doing this are pretty easy, what was a bit uglier was getting the UI disabled everywhere that was needed.

This feels particularly rough.  Please check the functionality thoroughly when reviewing.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -23,23 +23,27 @@
      {:color (:button-blue style/colors)})
    :render
    (fn [{:keys [props]}]
-     [:a {:title (:title-text props)
-          :style {:display "inline-block"
-                  :backgroundColor (:color props)
-                  :color "white" :fontWeight 500
-                  :borderRadius 2 :padding (if (:icon props) "0.7em" "0.7em 1em")
-                  :fontFamily (when (:icon props) "fontIcons")
-                  :fontSize (when (:icon props) "80%")
-                  :textDecoration "none"}
-          :href "javascript:;"
-          :onClick (fn [e] ((:onClick props) e))
-          :onKeyDown (common/create-key-handler [:space :enter] (:onClick props))}
-      (or (:text props) (icons/icon-text (:icon props)))
-      (when (= (:style props) :add)
-        [:span {:style {:display "inline-block" :height "1em" :width "1em" :marginLeft "1em"
-                        :position "relative"}}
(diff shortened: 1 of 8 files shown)


eddiebroad:
I tried to test this with a workspace on the DEV RAWLS server, but got the error "Method not allowed"?
Is that expected?
image

Did you have to do anything Matt to allow the method?  Has RAWLS implemented this yet is it known?
I observed no functionality with fake data turned "on"

MattPutnam:
You'll need the orchestration branch too.  Mock data doesn't do anything, we don't have a good way to capture state like this.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/summary_tab.cljs
```diff
@@ -107,6 +115,18 @@
...
+     (swap! state assoc :locking? locked-now?)
+     (endpoints/call-ajax-orch
+       {:endpoint (endpoints/lock-or-unlock-workspace (:workspace-id props) locked-now?)
+        :on-done (fn [{:keys [success? status-text status-code]}]
+                   (if success?
+                     (swap! state dissoc :server-response)
+                     (if (and (= status-code 409) (not locked-now?))
+                       (js/alert "Could not lock workspace, one or more analyses are currently running")
+                       (js/alert (str "Error: " status-text))))
+                   (swap! state dissoc :locking?))}))
```
dmohs: Each one of these swaps causes a component-did-update, which reloads the workspace. Try:
```Clojure
(swap! state #(if success? (dissoc % :locking? :server-response) (dissoc % :locking?)))
```

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -23,23 +23,26 @@
...
+                    :backgroundColor (:color props)
+                    :WebkitFilter (when disabled? "grayscale()")
+                    :cursor (when disabled? "default")
+                    :color "white" :fontWeight 500
+                    :borderRadius 2 :padding (if (:icon props) "0.7em" "0.7em 1em")
+                    :fontFamily (when (:icon props) "fontIcons")
+                    :fontSize (when (:icon props) "80%")
+                    :textDecoration "none"}
+            :href "javascript:;"
+            :onClick (when-not disabled? (fn [e] ((:onClick props) e)))
```
dmohs: Would prefer a `js/alert` here rather than modifying the title. If I'm clicking frantically on a button because I don't know why it's disabled, I may not hover long-enough to see the reason.

dmohs:
:+1: Functionality is great and the look is great. Two minor suggestions. I don't need to see this again—just merge when ready.

PR #224, opened 2015-10-09
View WDL and other method details when viewing a selected MC (from in a user workspace not method repo)

eddiebroad:
by having the WDL in view when in a selected MC:
1)  you don't have to go to the method repo to know the WDL
2)  the WDL, being known to the user, tells the user what the MC will do upon launch/execution

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_config_editor.cljs
@@ -8,11 +8,14 @@
     [org.broadinstitute.firecloud-ui.common.style :as style]
     [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
     [org.broadinstitute.firecloud-ui.page.workspace.launch-analysis :as launch]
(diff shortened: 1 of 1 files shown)


MattPutnam:
Instead of doing this, we should use the EntityDetails component.

eddiebroad:
@MattPutnam I had thought to use that in the gray area above (where there is the render-top-bar) but then the expandable/collapsable text pushes everything under it up and down so that's why I thought to put the viewer down below, so that it doesn't make stuff under it go up and down (cause there's nothing much under it!)

MattPutnam:
What I mean is, completely replace `render-top-bar` with the component.

eddiebroad:
@MattPutnam  Re : "What I mean is, completely replace render-top-bar with the component."
If the collapsable/exapandable WDL is there, then everything below the top-bar can go up/down depending on the WDL being collapsed/expanded.  Is that really desired?

MattPutnam:
I don't think it's a bad thing.  That's already how the method repo works.

eddiebroad:
I put the EntityDetail viewer at the top-bar now

davidangb:
Please recreate this PR against develop!

PR #225, opened 2015-10-09
Workflow details

MattPutnam:
Notes:
* The call ID currently shows as "Some(foo)" instead of "foo".  This is a bug in Cromwell, which has already been reported.
* Not showing fields "backend" and "backendStatus".  Initial thought is that the user won't care.
* Should GCS links be downloadable from here?


dmohs:
:+1: Love this. Looks great! Maybe squash your commits before merging, but merge when ready.

PR #226, opened 2015-10-09
Added :as-text property to table columns.

MattPutnam:
:as-text creates a text representation of the table data, and acts as a fallback to several other properties.  Started using it, and made a change to the Method Configurations tab, but didn't do a full overhaul.


dmohs:
Just one question about this pattern: Now that we know :as-text will not be used for creating TSV files (we have to do those in the server to support REST clients), does it still make sense? Is it still an improvement?

MattPutnam:
Even if we don't end up adding arbitrary table export, I think this will have value in cleaning up table column declarations.  Maybe I should go through and more thoroughly update them in this PR.

dmohs:
That might be helpful to verify this claim, but I trust your judgement either way. :+1: Good to merge whenever.

PR #228, opened 2015-10-13
Update endpoint to align with change in orchestration endpoint.

rushtong:
Aligns with https://github.com/broadinstitute/firecloud-orchestration/pull/109


dmohs:
:+1: I like this name.

PR #230, opened 2015-10-15
Show google bucket name on the workspace summary page

rushtong:


MattPutnam:
I think we can assume that the bucket will always be present, so we can get rid of the condition check.  Also, style/create-paragraph can take the child directly, no need to wrap in a div.

rushtong:
@MattPutnam - Feedback implemented - thank you!

PR #231, opened 2015-10-16
Redesigned filter buttons to use less space, and put them inline with the table toolbars

MattPutnam:
Note: this also addresses DSDEEPB-1344 and DSDEEPB-1503

Part of this involved creating a way to extend the toolbars on tables.  This was used on the method configs tab to inline the import button.

A TODO is to get the filter bars to center when there's space, but then be pushed over when there isn't.  David Mohs suggested looking at flex to do this.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -46,28 +46,45 @@
             "+"]])]))})
 
 
-(react/defc FilterButtons
-  (let [Button
-        (react/create-class
-          {:render
-           (fn [{:keys [props]}]
-             [:div {:style {:float "left" :textAlign "center"
-                            :backgroundColor (if (:active? props)
-                                               (:button-blue style/colors)
-                                               (:background-gray style/colors))
-                            :color (when (:active? props) "white")
-                            :marginLeft "1em" :padding "1ex" :width "16ex"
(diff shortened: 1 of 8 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -46,28 +46,45 @@
...
-     (fn [{:keys [props]}]
-       [:div {:style {:display "inline-block" :marginLeft "-1em"}}
-        (map (fn [button] [Button button])
-          (:buttons props))
-        (common/clear-both)])}))
+(react/defc FilterBar
+  {:get-initial-state
+   (fn [{:keys [props]}]
+     {:selected-index 0
+      :filtered-data (into {} (map-indexed
```
dmohs: This pattern causes pieces of the table rows to be stored in the state of a few different components. If I say, "can you make this component work more like Paginator," would you know what I mean?

MattPutnam: Maybe?  I don't feel too confident implementing it though

dmohs: Okay, can you make a tech debt bug for it and assign to me?

dmohs:
:+1: Looks good. Merge after creating follow-up bug about making FilterBar work more like Paginator.

PR #234, opened 2015-10-16
if the public user has the OWNER setting then change it to READER!

eddiebroad:
no action is taken if the change is to READER or NO ACCESS

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
@@ -24,7 +24,7 @@
 
 (react/defc AgoraPermsEditor
   {:render
-   (fn [{:keys [props state this]}]
+   (fn [{:keys [props  refs state this]}]
      [comps/Dialog
       {:width "75%"
        :blocking? true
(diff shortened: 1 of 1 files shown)


MattPutnam:
We should simply not have that option.  We certainly shouldn't be clobbering the user's choice.

eddiebroad:
Re : "we should simply...."  Can you explain more?  This is for if the user of the ACL is the public.  The main motivation is to prevent the situation where anyone who can log in is granted owner privelege

eddiebroad:
Re: "we should simply...." maybe these sort of things should be occurring in orchestration or in Agora?

eddiebroad:
Re: "we should simply..."  Is it thought that a checkbox for "Make Publicly Readable" would be better?

MattPutnam:
You're providing the user the option to select "OWNER", then yanking it out from under them if they choose it.  Just don't provide the option.

MattPutnam:
Yes, it seems like it's a binary option, so the checkbox sounds like a better idea.

eddiebroad:
Re: "Yes, it seems...." okay this idea I had thought of maybe might be better for consistency with other users.  I will cancel this PR

PR #236, opened 2015-10-19
Add some required props that were missing from launch analysis.

rushtong:
Also, the returned submission object is no longer an array, but a single object.


davidangb:
works for me locally!

PR #237, opened 2015-10-19
Multi-copy of entities between workspaces

MattPutnam:
Not that much different visually, now clicking entities toggles their selection, and a copy button appears for nonempty selections.  Error reporting has been lost, but there's another story for previewing conflicts that would overwrite any work I would do on that, so I'm leaving it as-is for now.


rushtong:
:+1: I can't begin to understand what's happening in `table.cljs`, but I like the functionality. Works for me locally and doesn't seem to introduce any regressions.

PR #239, opened 2015-10-20
workspace namespace -> Google project

MattPutnam:
Also some conversion from ':' to '/' and including the Google project in the banner across the top.
It's really hard to search for these, but I think I got them all. Please review thoroughly though!


rushtong:
:+1: I went through the app as much as I could and could not find any other incorrect "namespace" instances. Looks good!

PR #242, opened 2015-10-21
add the dev tag to  use that tagged image

eddiebroad:


dmohs:
:+1: Thanks for this.

PR #243, opened 2015-10-21
Monitor tab improvements

MattPutnam:
From the JIRA item:
* *Default sort by submission date descending* -- was already done
* *Show running vs. done status and error status in submissions list* -- This column already existed, so I added a red "!" icon in case it's the "done but with errors" status.  Other than that, I can't think of a good way to include more status/error info in the table without over-cluttering things.  That's what the detail page is for.
* *Details: status icon takes up too much space* -- I think it's fine.  It's the same component as the workspace summary tab.

Also, I changed the algorithm for determining the workspace's overall state.  Previously, it was reporting as "error" if any past submission had failed.  Now it checks the most recent success and failure times and only reports failure if the most recent submission failed.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -97,11 +97,11 @@
 
 
 (defn compute-status [workspace]
-  (let [lastSuccess (js/moment (get-in workspace ["workspaceSubmissionStats" "lastSuccessDate"]))
-        lastFailure (js/moment (get-in workspace ["workspaceSubmissionStats" "lastFailureDate"]))
-        count (get-in workspace ["workspaceSubmissionStats" "runningSubmissionsCount"])]
-    (cond (pos? count) "Running"
-          (.isAfter lastFailure lastSuccess) "Exception"
+  (let [last-success (js/moment (get-in workspace ["workspaceSubmissionStats" "lastSuccessDate"]))
+        last-failure (js/moment (get-in workspace ["workspaceSubmissionStats" "lastFailureDate"]))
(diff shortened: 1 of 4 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -97,10 +97,12 @@
 
 
 (defn compute-status [workspace]
-  (let [count (get-in workspace ["workspaceSubmissionStats" "runningSubmissionsCount"])]
-    (cond (not (nil? (get-in workspace ["workspaceSubmissionStats" "lastFailureDate"]))) "Exception"
-          (zero? count) "Complete"
-          :else "Running")))
+  (let [lastSuccess (js/moment (get-in workspace ["workspaceSubmissionStats" "lastSuccessDate"]))
```
dmohs: Clojure naming conventions here? Unless this makes more sense as JavaScript naming. Not sure if that was intentional.

src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -97,10 +97,12 @@
...
+        lastFailure (js/moment (get-in workspace ["workspaceSubmissionStats" "lastFailureDate"]))
+        count (get-in workspace ["workspaceSubmissionStats" "runningSubmissionsCount"])]
```
dmohs: I see you didn't choose this name, but let's not clobber the `count` function.

dmohs:
:+1: Code looks good with tiny tweaks. I shared my workspace with you because I don't see an error status that I expect. Let's look that one over together before merging.

PR #244, opened 2015-10-21
DSDEEPB-1505 and DSDEEPB-1531

MattPutnam:
* 1505: Disabled the expression text box when the selected entity is of the root entity type
* 1531: Use OKCancelForm for the launch analysis dialog


rushtong:
:+1: Looks good!

PR #245, opened 2015-10-22
Register page first pass

dmohs:


src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
```diff
@@ -514,3 +514,32 @@
...
       :crc32c "avsUEw=="
       :etag "CNjyodHyuscCEAI="})})
+
+
+(defn profile-get
+  ([on-done]
+   (utils/ajax-orch
+    "/profile"
+    {:on-done on-done
+     :mock-data {:userId "55"
```
MattPutnam: We need a `:canned-response` instead

dmohs: Thanks for that catch.

PR #246, opened 2015-10-22
Un-bold the username part of the displayed email address.

rushtong:
Tested locally.


MattPutnam:
Let's just get rid of this function completely, and in the summary tab we can remove the map that calls it.

rushtong:
I had considered that approach, but wasn't sure if we still wanted to keep it around for nostalgia's sake :-) Will remove and rebase.

rushtong:
Screwed up the merge ... closing this.

PR #247, opened 2015-10-22
DSDEEPB-489 and DSDEEPB-1618 for UI functionality and for a checkbox for public readability (depending on checked status) 

eddiebroad:
have UI report what is retireved from ORCH
set up for add new users to ACL

added helper utilities for retrieving public status

add random public user to endpoints

add filter utility and make the UI still represent values

put in a check to make sure user does not enter public for user

cleanup and prep for saving to back end

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
@@ -22,11 +22,11 @@
   ["READER" "OWNER" "NO ACCESS"])
 
 (def ^:private reader-level (nth access-levels 0))
-(def ^:private  owner-level (nth access-levels 1))
-(def ^:private  no-access-level (nth access-levels 2))
+(def ^:private owner-level (nth access-levels 1))
+(def ^:private no-access-level (nth access-levels 2))
 
 (defn- index-to-access-level [idx]
-  (str (nth access-levels idx)))
+  (nth access-levels idx))
 
 (defn- access-level-to-index [access-level]
   (case access-level
@@ -37,12 +37,9 @@
 (def ^:private column-width "calc(50% - 4px)")
 
 (defn- correspondsToReader [access-level]
-  (if
     (or
       (= access-level reader-level)
-      (= access-level owner-level))
-    true
-    false))
+      (= access-level owner-level)))
 
 (defn- filter-public [acl-vec]
   (let [hasNotPublicUser
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
@@ -21,12 +21,64 @@
 (def ^:private access-levels
   ["READER" "OWNER" "NO ACCESS"])
 
+(defn- index-to-access-level [idx]
+  (str (nth access-levels idx)))
+
+(defn- access-level-to-index [access-level]
+  (case access-level
+    "READER" 0
+    "OWNER" 1
+    "NO ACCESS" 2))
+
 (def ^:private column-width "calc(50% - 4px)")
+
+(defn- correspondsToReader [access-level]
+  (if
+    (or
+      (= access-level "READER")
+      (= access-level "OWNER"))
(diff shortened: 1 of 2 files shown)


rushtong:
Tested functionality and it seems to work as advertised.

rushtong:
There are a lot of cases where a single plain string is repeated, i.e. "READER", "OWNER", etc. Can these be refactored into a reusable object in some way? It would reduce copy/paste error and allow for easier updates when new or changed ACLs come along.

rushtong:
Done with review. Looks good as far as my clojure knowledge goes.

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -21,11 +21,66 @@
 (def ^:private access-levels
   ["READER" "OWNER" "NO ACCESS"])
 
+(def ^:private reader-level (nth access-levels 0))
+(def ^:private  owner-level (nth access-levels 1))
+(def ^:private  no-access-level (nth access-levels 2))
```
MattPutnam: Extra spaces

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -21,11 +21,66 @@
...
+
+(defn- index-to-access-level [idx]
+  (str (nth access-levels idx)))
```
MattPutnam: Unnecessary "str"

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -21,11 +21,66 @@
...
+
 (def ^:private column-width "calc(50% - 4px)")
 
+(defn- correspondsToReader [access-level]
+  (if
+    (or
+      (= access-level reader-level)
+      (= access-level owner-level))
+    true
+    false))
```
MattPutnam: `(if <cond> true false)` => `<cond>`

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -21,11 +21,66 @@
...
+
+(defn- filter-public [acl-vec]
+  (let [hasNotPublicUser
+        (fn [m]
+          (not
+            (and
+              (map? m)
+              (contains? m :user)
```
MattPutnam: Do we really need to be testing whether the item is a map and conforms to the expected schema?  Seems like upstream code should guarantee this.

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -21,11 +21,66 @@
...
+(defn- make-ui-vec [a-map]
+  {:user (get a-map "user")
+   :role (get a-map "role")})
+
+(defn- extract-last-public-access-level [acl-vec]
+  (let [hasPublicUser
+        (fn [m]
+          (and
+            (map? m)
+            (contains? m :user)
```
MattPutnam: Same comment as above.  Also, these functions are just inverses of each other and could be factored out.

src/cljs/org/broadinstitute/firecloud_ui/page/methods_configs_acl.cljs
```diff
@@ -60,19 +115,33 @@
...
                               :spellCheck false
-                              :defaultValue (acl-entry "user")})
+                              :defaultValue (:user acl-entry)
+                              :onChange (fn [e]
+                                          (let [new-val (-> e .-target .-value)
+                                                new-val-is-public (= "public" new-val)]
+                                            (when new-val-is-public
+                                              (do
+                                                (js/alert "Cannot set value to 'public'!  Use the check-box instead.")
+                                                (set! (-> e .-target .-value) "")))))})
```
MattPutnam: As with the previous PR, don't do this.  Validate the text fields when the OK button is pressed.  What if someone's ID starts with "public"?

PR #249, opened 2015-10-22
Workspace details additions

MattPutnam:
* Created by/date
* empty descrition placeholder
* pluralized workspace owner
* display count of running submissions when running, in the status box
* display count of total submissions and failures at bottom

Also re-re-fixed issues with workspace status calculation.  The last-success vs. last-failure check was failing if one/both of those fields were missing.  Maybe Rawls should be doing this?


rushtong:
:+1: Love the reformatted summary - looks good to me!

PR #250, opened 2015-10-22
DSDEEPB-1748 and DSDEEPB-1048 GUI - Redact items from method repository

eddiebroad:
got it working now need to tidy

got an ondelete working that reloads the entities upon successful deletion

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
@@ -469,8 +469,9 @@
                 (range (inc (rand-int 5))))})
 
 
-(defn delete-agora-entity [type ns n sid]
-  {:path (str "/" type "s/" ns "/" n "/" sid)
+(defn delete-agora-entity [config? ns n sid]
+  {:path (let [base (if config? "configurations" "methods")]
+           (str "/" base "/" ns "/" n "/" sid))
    :method :delete})
 
 
--- a/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
@@ -469,6 +469,11 @@
                 (range (inc (rand-int 5))))})
 
 
+(defn delete-agora-entity [type ns n sid]
+  {:path (str "/" type "s/" ns "/" n "/" sid)
+   :method :delete})
+
+
 (defn persist-agora-method-acl [ent]
   {:path (let [ent-type (ent "entityType")
                name (ent "name")
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
@@ -16,9 +16,8 @@
   (let [{:keys [workspace-id on-back]} props
         workspaces-list (:workspaces-list @state)]
     [:div {}
-     (when (:blocking? @state)
-       [comps/Blocker {:banner (if (:blocking-text @state)
-                                 (:blocking-text @state) "Please wait...")}])
+     (when (:blocking-text @state)
+       [comps/Blocker {:banner (:blocking-text @state)}])
      [:div {:style {:paddingBottom "0.5em"}}
       (style/create-link #(on-back)
         (icons/font-icon {:style {:fontSize "70%" :marginRight "0.5em"}} :angle-left)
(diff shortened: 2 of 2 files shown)


eddiebroad:
I used this command for I in `seq 1 30`; do  firecloud     -m push dummy.wdl  -t Workflow -y blah ; done ;
and the dummy wdl below to test this
wm8b1-75c:WDL esalinas$ cat dummy.wdl 
task myDummyTask  {
	command {
		cat /dev/null
	}

	meta {author : "Eddie Salinas"}

	runtime { docker: "eddiebroad/fastac" }
}

workflow myDummyWorkflow {
    call myDummyTask  
}

src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
```diff
@@ -469,6 +469,11 @@
                 (range (inc (rand-int 5))))})
 
 
+(defn delete-agora-entity [type ns n sid]
+  {:path (str "/" type "s/" ns "/" n "/" sid)
```
MattPutnam: Rather than automatically pluralizing the type like this, I'd prefer the correct path segment to be passed in.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -16,9 +17,9 @@
   (let [{:keys [workspace-id on-back]} props
         workspaces-list (:workspaces-list @state)]
     [:div {}
-     (when (:importing? @state)
-       [comps/Blocker {:banner "Importing..."}])
-
+     (when (:blocking? @state)
+       [comps/Blocker {:banner (if (:blocking-text @state)
+                                 (:blocking-text @state) "Please wait...")}])
```
MattPutnam: A pattern I wrote for this case (needing multiple blockers) a while back was to just key the blocker on the text itself.  I don't know if that code is still in the codebase, but it would look like this:

    (when (:blocking-text @state)
      [comps/Blocker {:banner (:blocking-text @state)}])

Then when you want to start blocking, you `assoc :blocking-text "desired message"`, and when you want to stop blocking, you `dissoc :blocking-text` (or `assoc` it to `nil`).

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -33,10 +34,31 @@
...
-                              :onClick #(swap! state assoc :show-perms-overlay? true)}]])
-
+                              :onClick #(swap! state assoc :show-perms-overlay? true)}]
+        [comps/SidebarButton
+         {:style :light :margin :top :color :exception-red
+          :text "Redact" :icon :trash-can
+          :onClick
+          (fn []
+            (let [conf (js/confirm "Are you sure ?")]
+              (when conf
```
MattPutnam: Convert

    (fn []
      (let [conf (js/confirm "Are you sure ?")]
        (when conf

to

    #(when (js/confirm "Are you sure?")

Also note the text.  Spaces don't go before punctuation.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -33,10 +34,31 @@
...
+                (let [type (if (contains? entity "method")
+                             "configuration" "method")
```
MattPutnam: First of all, there's already a `(let [config? (contains? entity "method")]` above.  Second, `create-import-form` is being called from two places that know if the entity is a method or config, so we should just pass that info to the function.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -33,10 +34,31 @@
...
+                      name (entity "name")
+                      namespace (entity "namespace")
+                      snapshotId (entity "snapshotId")]
+                  (do
+                    (swap! state assoc :blocking-text "Redacting..." :blocking? true)
+                    (endpoints/call-ajax-orch
+                      {:endpoint (endpoints/delete-agora-entity
+                                   type namespace name snapshotId)
+                       :on-done (fn [{:keys [success? status-text]}]
+                                  (do
```
MattPutnam: Don't need a `do` in functions.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -33,10 +32,25 @@
...
+                  snapshotId (entity "snapshotId")]
+              (swap! state assoc :blocking-text "Redacting...")
+              (endpoints/call-ajax-orch
+                {:endpoint (endpoints/delete-agora-entity
+                             config? namespace name snapshotId)
+                 :on-done (fn [{:keys [success? status-text]}]
+                            (swap! state dissoc :blocking-text)
+                            (if success?
+                              ((:on-delete props))
+                              (js/alert (str "Error ! Message : " status-text))))})))}]])
```
MattPutnam: I would prefer just "Error: "

MattPutnam:
I don't have a method that I can delete, but I'll trust that hitting the endpoint does what you'd expect.  :+1: aside from a cosmetic thing.

PR #255, opened 2015-10-26
Refactoring for DSDEEPB-1645

dmohs:
See individual commits.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -6,6 +6,7 @@
     [org.broadinstitute.firecloud-ui.common.icons :as icons]
     [org.broadinstitute.firecloud-ui.common.style :as style]
     [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
+    [org.broadinstitute.firecloud-ui.utils :as utils]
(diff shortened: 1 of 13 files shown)


MattPutnam:
Looks great overall.  My only issue is the placement of the filters in the data tab--I had intentionally put them in another row since having ~12 of them is apparently a real use case.

MattPutnam:
Also, it looks like the columns got messed up for data tables (including the data tab, the launch analysis dialog, and the copy dialog).  Each view is now showing the merging of all columns across all types, not just the columns for that type.

rushtong:
I think these kinds of large-scale code-moving refactors are near impossible to review. There's no clean way to compare them for small changes that may or may not have crept in from the copy/paste.

dmohs:
@MattPutnam Thank you. I didn't realize that. I'll move the filters back in the data tab and fix the other issue. @rushtong If that is true then I made a mistake. The code move commit should be mostly pure, with the changes to the code in subsequent commits.

dmohs:
@MattPutnam Thanks for catching the column bug. It's fixed. Made the code a bit more messy, but it shouldn't be tough to refine it in follow-up refactorings.

I left the filters in the data tab where they were. In my experiments they seemed to degrade gracefully. Happy to be proven wrong there.

PR #256, opened 2015-10-26
Update due to :select change

MattPutnam:
The change to the `:select` control makes it so that the combo box operates on its index value, rather than the display value.  In this case, it means a bit of a workaround where the editing is operating off index, but the endpoint uses string values.  Is this how we want to do this?


dmohs:
:+1:

PR #260, opened 2015-10-27
Remove Shibboleth Apache module

dmohs:


dmohs:
Your comments are all correct. This won't have any effect on devops' deployment. They are already setting the ORCH_URL_ROOT variable to https://firecloud-orchestration.dsde-dev.broadinstitute.org/.

eddiebroad:
:+1:

PR #261, opened 2015-10-28
switch to use docker compose and vault

dvoet:
docs https://github.com/broadinstitute/firecloud-environment


script/dev/start-server.sh
```diff
@@ -13,14 +13,6 @@ if [[ "$PWD" != "$HOME"* ]]; then
   exit 1;
 fi
 
-ORCH_URL_ROOT=${ORCH_URL_ROOT:-'https://firecloud-orchestration.dsde-dev.broadinstitute.org'}
+docker run --rm -it -v "$PWD":/working broadinstitute/dsde-toolbox render-templates.sh local $VAULT_TOKEN
```
dmohs: Add `src/docker` to the end of this so we can use paths that make more sense for this project.

src/main/config/ca-bundle.crt.ctmpl
```diff
@@ -0,0 +1 @@
+{{with $secret := vault "secret/dsde/local/local_broadinsititute_org" }}{{$secret.Data.ca_bundle_crt}}{{end}}
```
dmohs: Move this and others to `src/docker`.

src/main/config/docker-compose.yaml.ctmpl
```diff
@@ -0,0 +1,22 @@
...
+  image: {{.Data.ui_image}}
+  ports:
+    - "80:80"
+    - "443:443"
+    {{.Data.ui_additional_ports}}
+  {{.Data.ui_dns}}
+  {{.Data.ui_volumes}}
+  environment:
+    GOOGLE_CLIENT_ID: {{.Data.env_client_id}}
+    ORCH_URL_ROOT: {{.Data.env_orchestration_url}}
```
dmohs: As far as I can tell, this is the only harm done to our use cases. Now, developing against a local instance of orchestration is not as simple as setting this environment variable before calling `start-server.sh`. What is the path forward for this use case?

script/dev/start-server.sh
dvoet: I was hoping to convince you that it made more sense to make this project look like the others, push existing src sub dirs down into /src/main. There is nothing scala-esqe about it and it leaves room for other types of src like test.

plus I think the render-templates script you wrote assumes this structure so it makes it easy if all the projects are the same.

src/main/config/docker-compose.yaml.ctmpl
dvoet: I would say it is simpler now, either stand up the ui alone using this docker file (pointing to the dev instances) or stand up the entire env locally using the master docker compose file in firecloud-environment. It would also be easy to setup a docker compose file for just ui and orchestration.

src/main/config/ca-bundle.crt.ctmpl
dvoet: this has nothing to do with docker

script/dev/start-server.sh
dmohs: I do like cross-project consistency, but it's more important to me that the codebase is telling the story of exactly what it is doing today rather than making predictions about the future. I did not write render-templates (just dockerized what was already there). I don't want to add a `src/main` until we have `src/test`, and I don't want to predict we will have a `src/test` because 100% of the time I predict code will move in a certain direction it turns out to be wrong.

That said, I see the consistency argument, so I'm happy to leave this to your judgement.

src/main/config/docker-compose.yaml.ctmpl
dmohs: Okay. There's an example in the README for doing this, which will be wrong when this PR gets merged. Let's close the loop by updating those instructions as part of this PR.

rushtong:
:+1: Reviewed with @dvoet and @dmohs

PR #262, opened 2015-10-28
Update UI to call new delete endpoint

MattPutnam:
Depends on corresponding Orchestration PR (#139).


MattPutnam:
As mentioned in the updated subtask, the recursive option will be done in Rawls.  They've de-prioritized the issue so I'll make the last remaining UI work (add a checkbox and pass the option through) a separate story.

davidangb:
:+1: looks good to me. I'll leave it up to you about when to merge, since this depends on the orch PR. Merge when you're ready.

PR #263, opened 2015-10-28
Add field to view/edit root entity type

MattPutnam:
Also removed `:cursor "pointer"` from the validation check/x marks since it makes them look clickable

Prerequisites had already been removed from the UI


dmohs:
:+1:

PR #264, opened 2015-10-29
Fix infinite recursion by turning off advanced optimizations

dmohs:
I tried figuring out the problem, but couldn't diagnose it. The error was triggered from React, but I can't be certain that is where the problem originated. It's easily triggered by clicking the "Create New Workspace..." button after compiling with advanced optimizations. Turning them off fixes the problem, which is a bad UI bug, so it's worth doing and figuring out the underlying issue later.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -28,7 +28,10 @@
                                                             ":3449/figwheel-ws")}}}}}
              :minimized {:cljsbuild
                          {:builds {:client {:compiler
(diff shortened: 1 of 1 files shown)


MattPutnam:
Strange.  I wasn't seeing the issue with the "Create New Workspace..." button, but I was seeing it sporadically in tab loading.
At any rate, this seems to fix it.

PR #265, opened 2015-10-29
es_DSDEEPB-1500 for putting attributes in the summary area

eddiebroad:
more progress

some progress

got saving to work

some tidy

got delete working

fixed an unused item which caused a compilation issue

some tidy

arranged the attributes floating to the right

tidy

fixed an improper arity

set focus to last of fields

remove TODO message

fix bug

got focus working

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
@@ -56,98 +56,66 @@
                     (on-success)))})))
 
 
-(react/defc AttributeViewer
-  {:get-initial-state
-   (fn [{:keys [props]}]
-     {:editing? false :attrs-list (:attrs-list props)})
-   :render
-   (fn [{:keys [this props state]}]
-     [:div {:style {:margin "45px 25px"}}
-      (react/call :render-sidebar this)
-      [:div {:style {:display "inline-block"}}
-       [:div {:style {:display "inline-block"}}
-        (style/create-section-header "Workspace Attributes")
-        (create-section
-          (when (or (:saving? @state) (:deleting? @state))
-            [comps/Blocker {:banner "Updating..."}])
-          [:div {}
-           (map-indexed
-             (fn [i a]
-               [:div {}
-                [:div {:style {:float "left" :marginRight "0.5em"}}
-                 (style/create-text-field
-                   {:value (first a)
-                    :onChange #(swap! state update-in [:attrs-list i]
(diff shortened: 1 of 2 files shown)


MattPutnam:
A few issues:
* This is not filtering out the "description" property (if present)
* The layout is poor.  The location of the attributes jumps when you go to edit them.  My suggestion would be to move the "Created By" and "Submissions" sections to the right column (and move that column over further), and put the attributes below the bucket.  Maybe the bucket could be on the right too.
* There is no way to cancel editing, aside from bailing on the page.
* The Clone, Lock, and Delete buttons should be hidden when editing.

eddiebroad:
Re : "This is not filtering out the "description" property (if present)"  What is meant by this? Is it mean a workspace description?  Can clarification be provided?

MattPutnam:
We hijack a "description" attribute to use for the workspace description, since there isn't a real field for it.  So we need to filter out that attribute.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -114,13 +144,8 @@
                (let [owner? (= "OWNER" (workspace "accessLevel"))
                      writer? (or (= "WRITER" (workspace "accessLevel")) owner?)
                      status (common/compute-status workspace)]
-                 (if (:viewing-attributes? @state)
-                   [AttributeViewer {:ws workspace :writer? writer? :on-done #(swap! state dissoc :viewing-attributes?)
-                                     :attrs-list (mapv (fn [[k v]] [k v])
-                                                   (dissoc (get-in workspace ["workspace" "attributes"]) "description"))
```
MattPutnam: Here's where the old version was filtering out "description"

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
```diff
@@ -56,98 +56,66 @@
...
+          (:attrs-list @state))
+        [:div {:style {:display (when-not (:editing? @state) "none")}}
+         [comps/Button {:style :add :text "Add new"
+                        :onClick
+                        #(do
+                          (swap! state update-in [:attrs-list] conj ["" ""])
+                          (js/setTimeout (fn [] (common/focus-and-select
+                                           (-> (@refs (str "field" (- (count (:attrs-list @state)) 1))) .getDOMNode))) 250 )
+                          )
+                        }]]])
```
MattPutnam: Fix the formatting around here, and I think the timeout delay can be quite a bit shorter.

PR #266, opened 2015-10-29
Implemented error viewer, used for the file import

MattPutnam:
Also used this viewer in the publish and delete dialogs in the method config editor, in order to test more functionality.  The layouts may not be perfect for those components, but since this story is only concerned with the file import, I'm leaving future improvement for another story.

Also moved a bunch of stuff around.  Sorry about that.


eddiebroad:
I experience the nice new error viewer!  Error reporting is important!  Looks good!
image

eddiebroad:
I think this is ready to merge!

PR #267, opened 2015-10-31
eddie work on DSDEEPB-1490

eddiebroad:
add sorting on the status and name  columns

clean up and use capitalization func

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
@@ -119,7 +119,7 @@
          [{:sort-by (fn [m] (:status m))
            :header [:div {:style {:marginLeft -6}} "Status"] :starting-width 60
            :content-renderer (fn [data] [StatusCell {:data data}]) :filter-by :none}
-          {:sort-by (fn [m] (:name m))
+          {:sort-by :text  :as-text :name
            :header "Workspace" :starting-width 450
            :content-renderer (fn [data] [WorkspaceCell {:data data}]) :filter-by :name}
           {:header "Description" :starting-width 400
@@ -131,8 +131,7 @@
            :sort-initial :asc
            :content-renderer
            (fn [accessLevel]
-             [:div {:style {:padding "1.1em 0 0 14px"
-                            :fontStyle (when-not accessLevel "oblique")}}
+             [:div {:style {:padding "1.1em 0 0 14px"}}
               (clojure.string/capitalize accessLevel)])}]
          :data (:workspaces props)
          :->row (fn [ws]
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
@@ -127,8 +127,8 @@
                                [:div {:style {:padding "1.1em 0 0 14px"
                                               :fontStyle (when-not description "oblique")}}
                                 (or description "No description provided")])}
-          {:header "Access Level" :starting-width 400 :sort-by WSAccessLevelKeyFunc
-           :sort-initial :asc
+          {:header "Access Level"  
+           :sort-by WSAccessLevelKeyFunc :sort-initial :asc
            :content-renderer
            (fn [accessLevel]
              [:div {:style {:padding "1.1em 0 0 14px"}}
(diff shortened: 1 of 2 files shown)


MattPutnam:
I'm getting this:

screen shot 2015-11-02 at 1 48 21 pm

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -108,17 +116,24 @@
...
                                [:div {:style {:padding "1.1em 0 0 14px"
                                               :fontStyle (when-not description "oblique")}}
-                                (or description "No description provided")])}]
+                                (or description "No description provided")])}
+          {:header "Access Level" :starting-width 400 :sort-by WSAccessLevelKeyFunc
+           :sort-initial :asc
+           :content-renderer
+           (fn [accessLevel]
+             [:div {:style {:padding "1.1em 0 0 14px"
+                            :fontStyle (when-not accessLevel "oblique")}}
```
MattPutnam: Why is this font style here?  If there's no `accessLevel` there won't be any text to see.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -108,17 +116,24 @@
                    {:text "Running" :pred #(= "Running" (:status %))}
                    {:text "Exception" :pred #(= "Exception" (:status %))}]
          :columns
-         [{:header [:div {:style {:marginLeft -6}} "Status"] :starting-width 60
-           :content-renderer (fn [data] [StatusCell {:data data}])
-           :filter-by :none}
-          {:header "Workspace" :starting-width 450
-           :content-renderer (fn [data] [WorkspaceCell {:data data}])
-           :filter-by :name}
+         [{:sort-by (fn [m] (:status m))
```
MattPutnam: You can just do `:sort-by :status`.  But, I don't think we should have sorting on this column anyway.

eddiebroad:
Re: "I'm getting this:" wow!  when I loaded it I didn't get that at all.  Does it arrange better if you resize your browser?  I didn't experiment with that

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -108,17 +116,24 @@
...
+           :header [:div {:style {:marginLeft -6}} "Status"] :starting-width 60
+           :content-renderer (fn [data] [StatusCell {:data data}]) :filter-by :none}
+          {:sort-by (fn [m] (:name m))
```
MattPutnam: `:sort-by :name`.

You can also add `:as-text :name`, then use `:sort-by :text` and get rid of the `:filter-by`.

MattPutnam:
Looks like the browser window needs to be at least 1423 pixels wide in order to render properly, which I only have on my secondary monitor.

MattPutnam:
If you change line 98 of workspaces_list.cljs to:

    :header-style {:padding "0.5em 0 0.5em 14px" :overflow nil}

That will fix the issue with the alignment of the sort arrows.

eddiebroad:
referring to the issue of which you posted a screenshot, some tweaks from commits and PR suggestions helped some.....but on my screen if I resize the window small enough, then the table rows again show up like what you show in the screenshot.  What do you think?

MattPutnam:
Yeah, there's something wrong with the div width calculations that's causing the table rows to wrap.  That's beyond the scope of this issue, so as long as it still looks okay at a standard resolution let's leave it like that for now.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -87,8 +95,8 @@
          :cell-padding-left nil
          :header-row-style {:fontWeight nil :fontSize "90%"
                             :color (:text-light style/colors) :backgroundColor nil}
-         :header-style {:padding "0 0 1em 14px" :overflow nil}
-         :resizable-columns? false :reorderable-columns? false :sortable-columns? false
+         :header-style {:padding "0.5em 0 0.5em 14px" :overflow nil}
+         :resizable-columns? false :reorderable-columns? false :sortable-columns? true
```
MattPutnam: sortable columns is the default so you can actually take that property off

eddiebroad:
Re: "Yeah, there's something wrong...." okay.

eddiebroad:
Re: "sortable columns is the default.....".  I will remove the key/val pair

PR #268, opened 2015-11-02
DSDEEPB-1528 Method repo: default order by namespace then name

eddiebroad:

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
@@ -231,7 +231,8 @@
      [table/Table
       {:columns [{:header "Type" :starting-width 100}
                  {:header "Namespace" :starting-width 150}
-                 {:header "Name" :starting-width 200 :as-text #(% "name") :sort-by :text
+                 {:header "Name" :starting-width 200 :as-text #(% "name")
+                  :sort-by (fn [m] (str (m "namespace") (m "name"))) :sort-initial :asc
                   :content-renderer
                   (fn [item]
                     (style/create-link
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
@@ -230,7 +230,7 @@
    (fn [{:keys [props state]}]
      [table/Table
       {:columns [{:header "Type" :starting-width 100}
-                 {:header "Name" :starting-width 450 :as-text #(% "name")
+                 {:header "Item" :starting-width 450 :as-text #(% "name")
                   :sort-by (fn [m]
                              [(m "namespace") (m "name") (int (m "snapshotId"))])
                   :sort-initial :asc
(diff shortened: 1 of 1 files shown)


MattPutnam:
It's misleading to display one thing but then sort on another thing.  I think the right solution is to merge the namespace, name, and snapshot ID columns into one like in the Method Configurations tab.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -230,15 +230,18 @@
...
                   :content-renderer
                   (fn [item]
                     (style/create-link
                       (let [func (if (= :method (:type item)) :on-method-selected :on-config-selected)]
                         #((func props) (dissoc item :type)))
-                      (item "name")))}
-                 {:header "Snapshot ID" :starting-width 100}
+                      (str  (item "namespace") "/"
+                        (item "name")
+                        " Snapshot ID:" (item "snapshotId"))))}
```
MattPutnam: Use style/render-entity

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -230,15 +230,18 @@
    (fn [{:keys [props state]}]
      [table/Table
       {:columns [{:header "Type" :starting-width 100}
-                 {:header "Namespace" :starting-width 150}
-                 {:header "Name" :starting-width 200 :as-text #(% "name") :sort-by :text
+                 {:header "Name" :starting-width 450 :as-text #(% "name")
+                  :sort-by (fn [m]
+                             [(str (m "namespace") (m "name")) (int (m "snapshotId"))])
```
MattPutnam: Don't concatenate the namespace and name, leave it a vector of 3 items.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -230,15 +230,18 @@
    (fn [{:keys [props state]}]
      [table/Table
       {:columns [{:header "Type" :starting-width 100}
-                 {:header "Namespace" :starting-width 150}
-                 {:header "Name" :starting-width 200 :as-text #(% "name") :sort-by :text
+                 {:header "Name" :starting-width 450 :as-text #(% "name")
```
MattPutnam: We need to come up with a better header than "Name".  Not sure what it is though.  "Entity" is too technical.  "Item" maybe?

eddiebroad: Re: "We need to come up with a better header than "Name"....."  Either entity or item I think is okay, but what about "Method or Configuration" because those describe exactly the item.  Or what about "Method or Configuration Name"?

MattPutnam: I don't like "Method or Configuration".  It just feels weird.  I'd say go with either "Item" or "ID".

MattPutnam:
:+1:

PR #272, opened 2015-11-03
DSDEEPB-1500 for putting attributes in the summary area

eddiebroad:


MattPutnam:
* If I delete an attribute, but then cancel, I don't get the deleted attribute back.
* If I create an attribute, but then cancel, the attribute is still visible until I refresh the page.
* Bad merge--the code for navigating to a cloned workspace got blasted

PR #273, opened 2015-11-04
Refresh current tab by clicking it

MattPutnam:
Used :component-will-receive-props to accomplish this, however some tabs were also triggering reloads on :component-did-update, which was causing multiple loads.  So I added a simple locking mechanism to each tab to prevent simultaneous data load calls.  This is a bit of a hack; a future improvement would be to modify the TabBar component to trigger a specific reload rather than relying on React lifecycle methods.

Also fixed an issue with the tab bar's margins, and standardized some padding and the centering of the "loading..." spinner across all tabs.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
@@ -112,9 +112,12 @@
 
 
 (react/defc WorkspaceData
-  {:render
+  {:get-initial-state
+   (fn []
+     {:load-counter 0})
+   :render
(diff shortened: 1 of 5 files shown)


eddiebroad:
Does the :load-count implement  the "simple locking mechanism to each tab to prevent simultaneous data load calls."?

eddiebroad:
I think many people will appreciate the ability to click and cause a refresh

dmohs:
:+1: Two things. (1) Are you confident this slightly hacky pattern is applied only where necessary and (2) have you/will you create a follow-on story for the improvement?

MattPutnam:
(1) I think so.
(2) Done.

PR #275, opened 2015-11-04
DSDEEPB-1911: ensure we are working with a number when calling toFixed

davidangb:
when we first retrieve metadata from GCS, Google's json lists file sizes as strings:
```
{
  ...
 "size": "102"
  ...
}
```

If the size was large enough, the loop executed at least once, and we divided the string by 1000, which allowed JS to magically turn the string into a number.

But, if the size was small enough that we attempted to call toFixed() in the first iteration of the loop, we threw an error - String.toFixed doesn't exist. 

Fix: always ensure it's a float. In subsequent iterations of the loop, this is overkill, but it feels safe.


dmohs:
:+1:

PR #276, opened 2015-11-06
Use new /login endpoint for sign-in

dmohs:


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -110,15 +110,7 @@
...
        [:div {}
-        [:div {:style {:padding "1em" :borderBottom (str "1px solid " (:line-gray style/colors))}}
-         [:div {:style {:float "right" :fontSize "70%"}}
-          [:span {:style {:marginRight "1ex" :color (:link-blue style/colors)}}
-           (-> (utils/get-current-user)
-               (utils/call-external-object-method :getBasicProfile)
-               (utils/call-external-object-method :getName))]
-          ;; Sadly, reloading the page seems to be the only reliable way to log out.
-          [:a {:href "javascript:;" :onClick (fn [e] (-> js/window .-location (.reload)))}
-           "Log-Out"]]
```
davidangb: Please add a story to follow up on this. From memory, I think a logout button is required for fisma, so we'll need to have that (which possibly requires orch-side changes too). Maybe a separate story for the name display; that's nice but not strictly necessary.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -169,71 +161,36 @@
...
-   "https://www.googleapis.com/auth/devstorage.full_control"
-   "https://www.googleapis.com/auth/compute"])
+     [:div {:style {:padding "50px 25px"}}
+      [:div {:style {:marginBottom "2em"}} (text-logo)]
+      [:div {}
+       [comps/Button {:text "Sign In"
+                      :href (str "/service/login?path="
+                                 (js/encodeURIComponent (let [hash (nav/get-hash-value)]
+                                                          (if (clojure.string/blank? hash)
+                                                            "workspaces"
```
davidangb: if workspaces is ever not the default, we'll have to change this in two places. No action necessary for this PR; just talking out loud.

dmohs: This is why the path parameter should actually be optional (or at least blank is okay).

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Done.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -243,15 +200,22 @@
...
       (footer)])
    :component-did-mount
    (fn [{:keys [this state]}]
-     (.addEventListener js/window "hashchange" (partial react/call :handle-hash-change this)))})
+     (.addEventListener js/window "hashchange" (partial react/call :handle-hash-change this))
+     (let [hash (-> js/window .-location .-hash)
+           at-index (utils/str-index-of hash "?access_token=")
+           [_ token] (clojure.string/split (subs hash at-index) #"=")]
+       (when-not (neg? at-index)
+         (swap! state assoc :access-token token)
```
davidangb: So, this changes end-user behavior. Now, since we're only storing the access token in state, the user must re-login on every full browser reload - such as loading from a bookmark or opening a new tab.

I don't want to hold up this PR for this, but I do think we need to go back to the old behavior - maybe store the token in a cookie.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: I tried to use an empty value but I got an error.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -35,14 +35,7 @@
 
 
 (react/defc Page
-  {:get-initial-state
-   (fn [{:keys [props]}]
-     (when (:new-registration? props)
-       (let [google-profile (-> (utils/get-current-user)
-                                (utils/call-external-object-method :getBasicProfile))]
-         {:initial-values {:name (utils/call-external-object-method google-profile :getName)
```
davidangb: did you experiment at all with using the google api, and manually init-ing it with the access token? (I don't know if this is possible, or if maybe you need the id token too, or what). Again, nothing to change for this PR, but I think that is an interesting line of research, to see if we can get the user's name etc. back.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
davidangb: we can/should tackle this later - but I like having the path param be required and adding the default from the UI - that way, orchestration doesn't know anything about the page (err, fragment) layout of the UI. But honestly, either way.

davidangb:
:+1: works for me locally. I have comments, but I think all of my comments should be handled as follow-on stories. Feel free to merge when ready - if you want to clean any of this up before merging, go for it, otherwise click the big green button!

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: This is exactly how the UI works. If a blank path were allowed, and came back in the redirect, the UI would do the right thing.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
dmohs: Added a story for this.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Added story.

PR #277, opened 2015-11-06
Cleaned up use of data tables throughout UI

MattPutnam:
* Created table/EntityTable, used in the data tab and launch analysis dialog
* Redesigned copy data from workspace to use EntitySelector instead
* Redesigned launch analysis dialog to not use radio button column
* Data tables now also strip out the entity name from references to other entities

Other notes:
* Some rework to the table regarding default properties: using the `(or (:propname props) default)` pattern instead of `:get-default-props` makes it a little cleaner to override them.  With `:get-default-props` passing a `nil` value counts as overriding the default.
* EntitySelector was just too different to effectively merge.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -1,5 +1,6 @@
 (ns org.broadinstitute.firecloud-ui.common.table
   (:require
+   [clojure.set :refer [union]]
    [dmohs.react :as react]
    [org.broadinstitute.firecloud-ui.common :as common]
    [org.broadinstitute.firecloud-ui.common.components :as comps]
@@ -106,9 +107,7 @@
       :resizable-columns? true
       :reorderable-columns? true
       :sortable-columns? true
-      :filterable? true
-      :empty-message "There are no rows to display."
-      :toolbar identity})
+      :filterable? true})
    :get-initial-state
    (fn [{:keys [this props]}]
      (set! (.-filtered-data this) (if-let [filters (:filters props)]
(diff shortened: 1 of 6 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -240,3 +239,48 @@
...
+                       {:header "Entity Name" :starting-width 100
+                        :content-renderer (or (:entity-name-renderer props)
+                                            (fn [entity] (entity "name")))}]
+                      (map (fn [k] {:header k :starting-width 100
+                                    :content-renderer
+                                    (fn [attr-value]
+                                      (if (and (map? attr-value)
+                                            (= (set (keys attr-value)) #{"entityType" "entityName"}))
+                                        (attr-value "entityName")
+                                        ((:attribute-renderer props) attr-value)))})
```
dmohs: This is a null call on launch analysis for me.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
```diff
@@ -63,45 +63,21 @@
...
-                      (m "entityType")
-                      (m "name")]
-                     (map (fn [k] (get-in m ["attributes" k])) attribute-keys)))}])]
+       [:div {:style {:marginBottom "1em" :fontSize "140%"}}
+        (str "Selected: "
+          (if-let [e (:selected-entity @state)]
+            (str (e "name") " (" (e "entityType") ")")
+            "None"))]
+       [table/EntityTable
+        {:entities (:entities props)
```
dmohs: Needs an attribute renderer? And maybe an assertion in the EntityTable when this is accidentally left out?

dmohs:
:+1: Really nice cleanup. Looks like one missing piece. Merge when ready or I can retest.

dmohs:
:+1: Works great. Probably want to squash then merge.

PR #278, opened 2015-11-06
Cleanup to the method config editor

MattPutnam:
Moved method details to bottom, removed old view, added loading indicator


dmohs:
:+1: Nice.

PR #279, opened 2015-11-09
Remove oauth requirements from the ui proxy

rushtong:
See https://broadinstitute.atlassian.net/browse/DSDEEPB-1832

With oauth in front of orchestration, it is not needed in front of UI.

### DIFF
--- a/src/docker/run-apache.sh
+++ b/src/docker/run-apache.sh
@@ -49,13 +49,6 @@ LOCATION_DIRECTIVES=$(cat << EOF
     ProxyPass ${ORCH_URL_ROOT}
     ProxyPassReverse ${ORCH_URL_ROOT}
   </Location>
-
-  <Location "/service/api">
-    AuthType oauth20
-    Require valid-user
-    ProxyPass ${ORCH_URL_ROOT}/api
-    ProxyPassReverse ${ORCH_URL_ROOT}/api
-  </Location>
 EOF
 )
 


dmohs:
I think part of this cleanup should include the Dockerfile. It should no longer inherit from `broadinstitute/openidc-baseimage`. Instead, it should inherit from `httpd:2.4` or `nginx:1.9` (my preference as I prefer the performance characteristics of nginx over apache).

rushtong:
I would prefer to pose that question to devops - the current base-image is their call, as far as I understand things.

dmohs:
I promise they are on board, but feel free to verify that claim.

rushtong:
Yup - they have no opinion on the subject. I'll look into this and see if the current scripts work with nginx.

src/docker/run-apache.sh
```diff
@@ -51,8 +51,6 @@ LOCATION_DIRECTIVES=$(cat << EOF
   </Location>
 
   <Location "/service/api">
```
davidangb: I think you could remove this entire Location block - the one above it is a superset.

rushtong: Yes, I agree with that.

dmohs:
:+1: This is good to go with the follow-on story to update the Dockerfile.

davidangb:
Follow-on story is DSDEEPB-1963.

PR #281, opened 2015-11-10
Fixed error messages and cancel bug in workspace clone dialog

MattPutnam:
DSDEEPB-1973 was a simple oversight.

For DSDEEPB-1918, the error viewer was extended to take a map of "expected" error codes with associated messages, which is used for the 409 (conflict).


rushtong:
:+1: Bugs are indeed fixed.

PR #282, opened 2015-11-12
Add User Profile link and nav context

rushtong:
**DO NOT MERGE**: I would like to squash this when the time comes.

This PR:
* Common method for parsing the user profile info from orchestration.
* Provide a top-level nav structure for routes and route handlers
* Populate the user profile page with pre-existing information
* Show the user's name in the top right corner of the page for easy navigation.

Lots of kudos to @MattPutnam for his clojure map-deconvolution expertise.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -1,6 +1,7 @@
 (ns org.broadinstitute.firecloud-ui.main
   (:require
    [dmohs.react :as react]
+   [org.broadinstitute.firecloud-ui.common :as common]
    [org.broadinstitute.firecloud-ui.common.components :as comps]
    [org.broadinstitute.firecloud-ui.common.style :as style]
    [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
@@ -69,14 +70,18 @@
       [Link {:href "mailto:firecloud-help@broadinstitute.org?Subject=FireCloud%20Help" :text "Support" :target "_blank"}]]]))
 
 
-(def top-nav-bar-items
-  [{:key :workspaces
+(def routes
+  [{:key :profile
+    :render #(react/create-element profile-page/Page %)}
+   {:key :workspaces
     :name "Workspaces"
     :render #(react/create-element workspaces/Page %)}
    {:key :methods
     :name "Method Repository"
     :render #(react/create-element method-repo/Page %)}])
 
+(def top-nav-bar-items
+  (filter (fn [r] (contains? #{:workspaces :methods} (:key r))) routes))
 
 (react/defc TopNavBarLink
   {:render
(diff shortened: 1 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -126,24 +135,24 @@
...
-          (do
-            (set! (.-errorMessage this) status-text)
-            (swap! state assoc :registration-status :error))))))})
+       (fn [{:keys [success? status-text get-parsed-response]}]
+         (let [parsed-values (common/parse-profile (get-parsed-response))]
+           (cond
+             (and success? (= (:isRegistrationComplete parsed-values) "true"))
+             (swap! state assoc :registration-status :registered :name (:name parsed-values))
+             (not success?)
+             (swap! state assoc :registration-status :not-registered)
```
dmohs: This would be good time to fix another bug: 404 means the key isn't set, but anything else is a server error. The `:else` condition here is never actually hit, so a server error sends the user to the registration page (which isn't right).

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -4,6 +4,7 @@
    [dmohs.react :as react]
    [org.broadinstitute.firecloud-ui.common.components :as components]
    [org.broadinstitute.firecloud-ui.common.style :as style]
+   [org.broadinstitute.firecloud-ui.common :as common]
```
dmohs: nit: alphabetize.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -106,11 +111,15 @@
...
      (let [nav-context (nav/parse-segment (:nav-context props))
            page (keyword (:segment nav-context))]
-       (when-not (or (contains? (set (map :key top-nav-bar-items)) page)
+       (when-not (or (contains? (set (map :key routes)) page)
                      (= page :status))
          (nav/navigate (:nav-context props) "workspaces"))
        [:div {}
         [:div {:style {:padding "1em" :borderBottom (str "1px solid " (:line-gray style/colors))}}
+         [:div {:style {:float "right" :fontSize "70%"}}
+          [:a {:href "javascript:;"
```
dmohs: This should be `:href "#profile"` rather than having an onClick handler. That way, the user can open in a new tab or whatever they want. This was my bad, but let's grab the opportunity to correct it.

dmohs:
:+1: This is great as-is, but I'd love you to fix a couple of other bugs while you're touching this code.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: @dmohs - I'm not following this suggestion. The way I read the code, this is what happens:

* registered key exists -> assoc registered and name values - move user along to requested page (lines 132-140)
* registered key does not exist -> assoc not-registered and let the component render the registration page (lines 128, 129)
* Anything else (i.e. a server error when making the get-profile call) puts the error into :registration-status, which is then handled in lines 126 and 127.

As I see it, a server error should trigger line 127.

dmohs:
This is really great. Tested everything and works perfectly as far as I can tell. Merge when ready.

PR #283, opened 2015-11-13
Fix the encoding issue by decoding first, then encoding.

rushtong:
This approach preserves the original format of the user's original path and has the least impact on other navigation code.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -166,10 +166,10 @@
       [:div {}
        [comps/Button {:text "Sign In"
                       :href (str "/service/login?path="
-                                 (js/encodeURIComponent (let [hash (nav/get-hash-value)]
+                                 (js/encodeURIComponent (js/decodeURIComponent (let [hash (nav/get-hash-value)]
                                                           (if (clojure.string/blank? hash)
                                                             "workspaces"
-                                                            hash))))}]]
+                                                            hash)))))}]]
       [:div {:style {:marginTop "1em"}} [RegisterLink]]
       [:div {:style {:maxWidth 600 :paddingTop "2em" :fontSize "small"}}
        [:p {:style {:fontWeight "bold"}} "WARNING NOTICE"]


davidangb:
This works for the use case I tried, but it's not exactly the same. 

* navigate to #workspaces/broad-dsde-dev%3Adavidan_test1
* reload the browser, triggering a login
* notice that after login, I land at #workspaces/broad-dsde-dev:davidan_test1

I wonder if there is any case where this will break. I don't know of any right now.

rushtong:
@davidangb I saw the same issue - but that is from much older code in the navigation parser. I looked at trying to fix that too, but it wasn't *really* broken so I left it alone.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -166,10 +166,10 @@
       [:div {}
        [comps/Button {:text "Sign In"
                       :href (str "/service/login?path="
-                                 (js/encodeURIComponent (let [hash (nav/get-hash-value)]
+                                 (js/encodeURIComponent (js/decodeURIComponent (let [hash (nav/get-hash-value)]
```
dmohs: I think this value is coming back from orchestration double-encoded. If that is true, this is masking over the problem, which I'd really like us to not do. I'm working on verifying this theory.

rushtong: That could be true - I didn't check that theory either. What I do know is that when you navigate to a workspace normally, the ":" gets encoded somewhere in the nav-context code (I couldn't figure out how to undo that encoding). So, whenever it is passed to something else, it is *already* encoded before traveling to any other service.

rushtong: @dmohs - I'm pretty sure that going through the google sign in process, the original callback url that is part of the overall request gets encoded. By the time that url fragment gets to the orchestration layer, it has been encoded.

davidangb: After poking at this for a bit - yes, the issue appears to be in orchestration, specifically lines 188-195 of OAuthService.scala. But it's happening because of built-in rules of spray's Uri, and it's surprisingly hard to override. Meanwhile, I don't think this current PR is actually going to cause any problems - since this is all in the fragment, any character is valid.

rushtong: I will pull this back PR then - no need to add code if it doesn't really do anything.

PR #284, opened 2015-11-13
Move profile endpoints under /service/register

rushtong:
This PR goes hand in hand with an Orchestration component: https://github.com/broadinstitute/firecloud-orchestration/pull/153


davidangb:
:+1: looks good works good!

PR #287, opened 2015-11-13
Some minor touchup to the data tables

MattPutnam:
* Make the name column wider by default
* Allow attribute columns to be wider, up to a point, if there's room
* Render lists-of-references-to-data better


dmohs:
:+1: Looks good. Holding off merge until after standup discussion today.

PR #288, opened 2015-11-13
userId -> email

dvoet:
I am not entirely certain this works. rawls changed the update acl enpoint userId is not email.


eddiebroad:
if there's no member called "email", rawls returns a message like this 
image

PR #289, opened 2015-11-13
Fix broken Thurloe status link

rushtong:


src/cljs/org/broadinstitute/firecloud_ui/page/status.cljs
```diff
@@ -34,8 +34,12 @@
            [:div {} "Response text:" [:br] (-> (:xhr response) .-responseText)]]))])
    :component-did-mount
    (fn [{:keys [props state]}]
-     (utils/ajax-orch (:path (:request props))
```
dmohs: Can you try this:
```clojure
(utils/ajax-orch (:path (:request props))
                       {:on-done #(swap! state assoc :response %)}
                       :service-prefix (when (= (:path (:request props)) "/profile") "/service/register"))
```

rushtong: I was trying something similar that wasn't working ... will give that version a shot

rushtong: That ends up putting nil into `:service-prefix` so that all of the other api calls no longer have `"/service/api"`

dmohs:
:+1: And thanks for trying out my suggestion.

PR #291, opened 2015-11-16
Don't clobber configs on clean build

dmohs:


davidangb:
I personally like killing the configs on clean, but I suspect I'm in the minority. Code works and does what you intend; I'm okay with it.

Remove the set -x though? Did you want that there explicitly, or is that an artifact from debugging? (I don't like it; it sends scary output when calling the clean script)

dmohs:
I feel like we should be able to tweak the docker compose file in target/config. If it gets blasted on every build, this isn't possible. Does that sound reasonable to you?

I prefer the -x in all scripts so they tell us what they're doing. If that's just my own personal preference, I can take it out.

PR #292, opened 2015-11-16
Make the table display a little nicer.

rushtong:
Use the content of the workspace name and description to set up the starting width of the columns instead of using a large default value. 10 x the number of characters is a good starting estimate. I can get the window down below 750 px before it starts wrapping again.

This is a bit of a hacky approach. I would prefer using a fully responsive table but that looks like a significant change to the current table widget which is not in the cards at present.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
@@ -86,11 +86,26 @@
       [:div {:style {:padding "1em 0 0 1em" :fontWeight 600}}
        (get-in props [:data :name])]])})
 
+(defn- get-workspace-name-string [ws]
+  (str (get-in ws ["workspace" "namespace"]) "/" (get-in ws ["workspace" "name"])))
+
+(defn- get-max-workspace-name-length [workspaces]
+  (apply max (map count (map #(get-workspace-name-string %) workspaces))))
+
+(defn- get-workspace-description [ws]
+  (get-in ws ["workspace" "attributes" "description"]))
+
+(defn- get-max-workspace-description-length [workspaces]
+  (apply max (map count (cons (map #(get-workspace-description %) workspaces) ["No description provided"]))))
 
 (react/defc WorkspaceTable
   {:render
    (fn [{:keys [state props]}]
-     (let [border-style (str "1px solid " (:line-gray style/colors))]
+     (let [border-style (str "1px solid " (:line-gray style/colors))
+           max-workspace-name-length (get-max-workspace-name-length (:workspaces props))
+           max-description-length (get-max-workspace-description-length (:workspaces props))]
+       (utils/cljslog (str "max-workspace-name-length: " max-workspace-name-length))
+       (utils/cljslog (str "max-description-length: " max-description-length))
        [table/Table
         {:empty-message "No workspaces to display."
          :cell-padding-left nil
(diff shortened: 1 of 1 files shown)


eddiebroad:
dynamic is good.  I'm not sure if @MattPutnam  wanted to take a look?

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -86,11 +86,26 @@
...
        (get-in props [:data :name])]])})
 
+(defn- get-workspace-name-string [ws]
+  (str (get-in ws ["workspace" "namespace"]) "/" (get-in ws ["workspace" "name"])))
+
+(defn- get-max-workspace-name-length [workspaces]
+  (apply max (map count (map #(get-workspace-name-string %) workspaces))))
+
+(defn- get-workspace-description [ws]
+  (get-in ws ["workspace" "attributes" "description"]))
```
davidangb: (not a full code review, but ...) is there any way to integrate get-workspace-description and get-workspace-name-string with the renders in WorkspaceTable, so we're not duplicating logic? (i.e. "if no description entered, use 'No description provided'")

rushtong: I've been thinking about that exact thing too - I think I have a way to do that.

MattPutnam:
Hmm, I like the idea, but
* I wish there were a better way to calculate string width.  Tried a few approaches and couldn't get anything to work.  I tested with names like "iiii..." and "WWWW...." and both did sort of weird things.
* While this makes things better, it still doesn't fix the underlying problem

rushtong:
I completely agree, hence my comment that this is a hacky solution. The
real solution is to make the whole table completely responsive but that
seems like a major refactoring. This is really just an attempt to make it
less bad.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -121,9 +136,9 @@
...
            :content-renderer (fn [data] [StatusCell {:data data}])}
           {:as-text :name :sort-by :text
-           :header "Workspace" :starting-width 450
+           :header "Workspace" :starting-width  (* max-workspace-name-length 10)
            :content-renderer (fn [data] [WorkspaceCell {:data data}])}
-          {:header "Description" :starting-width 400
+          {:header "Description" :starting-width (* max-description-length 10)
            :content-renderer (fn [description]
                                [:div {:style {:padding "1.1em 0 0 14px"
                                               :fontStyle (when-not description "oblique")}}
```
davidangb: is there any way to integrate the "No description required" here with the one in get-max-workspace-description-length? Ideally, we wouldn't have that hardcoded string in both places.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -86,11 +86,26 @@
...
+  (apply max (map count (cons (map #(get-workspace-description %) workspaces) ["No description provided"]))))
 
 (react/defc WorkspaceTable
   {:render
    (fn [{:keys [state props]}]
-     (let [border-style (str "1px solid " (:line-gray style/colors))]
+     (let [border-style (str "1px solid " (:line-gray style/colors))
+           max-workspace-name-length (get-max-workspace-name-length (:workspaces props))
+           max-description-length (get-max-workspace-description-length (:workspaces props))]
+       (utils/cljslog (str "max-workspace-name-length: " max-workspace-name-length))
```
davidangb: remove log statements

davidangb:
I agree with both Matt and Greg - it would be nice to have a cleaner way to do this. Either it's a complete hack, doing something like offscreen-rendering a span or canvas containing the text and asking JS to calculate its width, then removing it; or a larger refactoring of making the table responsive. I say we keep going with this PR as-is (minor code comments inline), create a followup story for full responsiveness.

davidangb:
Just pulled and did some functional testing. One more request: at line 117, where the Table defines its row-style, add:
```
:overflow "hidden"
```
to prevent wrapping, even at very small widths. you'll still get wrapping in the filter bar, but that's out of scope.

And I hate to say it, but just adding overflow: hidden to the develop branch is probably good enough ... but I do like this PR better, as it is slightly more accurate for column widths.

rushtong:
Agreed, overflow: hidden is a lot nicer.

davidangb:
:+1:

PR #293, opened 2015-11-16
DSDEEPB-1714: download proxy

davidangb:
one-line fix to point the UI at the right endpoint for downloading TSVs.

new Location directive, containing a rewrite rule, for the proxy itself. This directive looks for a cookie named "FCtoken", and copies the value of that cookie into an Authorization: Bearer header.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
@@ -2,7 +2,7 @@
   (:require
     [dmohs.react :as react]
     [clojure.set :refer [union]]
-    [goog.net.cookies :as cks]
+    goog.net.cookies
     [org.broadinstitute.firecloud-ui.common :as common]
     [org.broadinstitute.firecloud-ui.common.components :as comps]
     [org.broadinstitute.firecloud-ui.common.dialog :as dialog]
@@ -11,10 +11,10 @@
     [org.broadinstitute.firecloud-ui.common.table-utils :as table-utils]
     [org.broadinstitute.firecloud-ui.common.style :as style]
     [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
-    [org.broadinstitute.firecloud-ui.utils :as utils :refer [access-token]]
     [org.broadinstitute.firecloud-ui.page.workspace.data.copy-data-workspaces :as copy-data-workspaces]
     [org.broadinstitute.firecloud-ui.page.workspace.data.entity-selector :refer [EntitySelector]]
     [org.broadinstitute.firecloud-ui.page.workspace.data.import-data :as import-data]
+    [org.broadinstitute.firecloud-ui.utils :as utils :refer [access-token]]
     ))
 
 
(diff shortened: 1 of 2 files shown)


dmohs:
What do you think about putting the Apache config into orchestration instead? If we do that, we leave the option open for deploying the UI code into Google's CDN, which means we don't have to worry about load balancing, etc.

davidangb:
tbh, I haven't tried this in orchestration, but I'm pretty sure it won't work. The rewrite needs to happen BEFORE the OAuth proxy, which is why I (in discussion with Andrew and Lukas) thought it best to put in the UI layer. Additionally, this keeps the orchestration proxy stock, like rawls/agora.

If we want to go the CDN route, we could spin up a new, single-purpose apache at something like download.firecloud.org just to handle this.

davidangb:
Updated to set a cookie (with 5-minute expiration) just prior to downloading TSV, based on feedback from https://github.com/broadinstitute/firecloud-orchestration/pull/155. If this ui PR is good, the orchestration PR can be abandoned.

rushtong:
I can't get this to work with the just the UI changes (ui pointing to remote orchestration). Exploring the UI+Orch combination now.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
```diff
@@ -2,6 +2,7 @@
   (:require
     [dmohs.react :as react]
     [clojure.set :refer [union]]
+    [goog.net.cookies :as cks]
```
dmohs: It doesn't look like you're using the short name, so could could just replace:
```clojure
[goog.net.cookies :as cks]
```
with
```clojure
goog.net.cookies
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
```diff
@@ -10,6 +11,7 @@
     [org.broadinstitute.firecloud-ui.common.table-utils :as table-utils]
     [org.broadinstitute.firecloud-ui.common.style :as style]
     [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
+    [org.broadinstitute.firecloud-ui.utils :as utils :refer [access-token]]
```
dmohs: nit: alphabetize

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
```diff
@@ -61,7 +63,8 @@
                 (when-let [selected-entity-type (or (:selected-entity-type @state) (:initial-entity-type @state) (first (:entity-types @state)))]
                   [:a {:style {:textDecoration "none" :float "left" :margin "7px 0 0 1em"}
                        :href (str "/service/api/workspaces/" (:namespace (:workspace-id props)) "/"
-                               (:name (:workspace-id props)) "/" selected-entity-type "/tsv")
+                               (:name (:workspace-id props)) "/entities/" selected-entity-type "/tsv")
+                       :onClick (fn [] (.set goog.net.cookies "FCtoken" @access-token 300))
```
dmohs: It is so awesome that you did this instead of my custom cookies stuff. I always forget that this entire library is available to us.

dmohs:
@rushtong Because of the modification to run-apache.sh, the cookie won't be processed unless you rebuild the container.

This worked great for me. I even got an unauthorized after waiting for a while to reload the download page after opening the link in a separate tab, so I can confirm that works too. I agree with the logic about keeping the stock oauth proxy on orch, so I'm 100% on board with this solution. Merge when ready.

rushtong:
I did have this working once when I was running the tsv orchestration branch. I'm still trying to get this to work with just the UI code in place.

rushtong:
I'll hand over the review to @dmohs - I can't get this to work locally. If it works for him, I'm happy.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/data/tab.cljs
davidangb: ggrrkgkkl!!!rl!!%@#@%S! I *meant* to alphabetize. "utils" is before "workspace". in my sleepiness I forgot about the extra "page" package. humph.

PR #294, opened 2015-11-17
DSDEEPB-1898: remove checkmarks in workspace method config viewer

davidangb:
jira ticket was a bit sparse; I'm hoping this is what was asked for.

before:
before

after:
after


MattPutnam:
Yeah, that's what I'm assuming the JIRA item means.

Super nitpicky, but we're preferring `when` instead of `if` when there's only a positive case.  Also, merge the closing parens onto the end of the previous line.

PR #299, opened 2015-11-18
DSDEEPB-1919: rename Submissions to Analysis Submissions

davidangb:
smallest PR ever.


eddiebroad:
it qualifies as itsy-bitsy-teeny-tiny-weenie!

eddiebroad:
From the DSDEEPB comment :This is such a simple task, I ranked it to the top so we could just get it out of the way.

PR #300, opened 2015-11-18
DSDEEPB-2042: download-tsv link available on initial tab load

davidangb:
look for the selected entity type, then an explicit value for the initial entity type, then use the first entity type.


rushtong:
:+1:

PR #301, opened 2015-11-18
DSDEEPB-2054: quote boolean value in docker compose for 1.9 compatibility

davidangb:
so small.


rushtong:
I see how this should work, but it doesn't seem to work in practice. Running
```
docker run --rm -it -v "$PWD":/working -v $HOME:/root broadinstitute/dsde-toolbox render-templates.sh local
```
still has the effect of leaving the bareword unquoted.

PR #304, opened 2015-11-18
DSDEEPB-2040: Highlight selected entity in launch analysis

MattPutnam:
Also converted entities to entity ids in this form


dmohs:
If I check out this branch and go to launch an analysis, the entities list is empty.

dmohs:
:+1: Very nice!

PR #305, opened 2015-11-19
fix build error

MattPutnam:


rushtong:
I did take a look at this for https://broadinstitute.atlassian.net/browse/DSDEEPB-2074 - the code looks fine to me. I think it should get at least one more set of eyes on it before calling that story complete.

PR #306, opened 2015-11-19
DSDEEPB-1843: Edit workspace description

MattPutnam:
Left as separate commits for ease of reading.  This involved some rearranging of things, both in terms of UI layout and code layout.


dmohs:
:+1: This works great. The only thing I noticed is that it degrades somewhat poorly as the window gets narrower. I added a commit to address this. If you're okay with it, feel free to merge.

PR #307, opened 2015-11-19
DSDEEPB-2053: Profile registration now makes single call to orchestration

davidangb:
See me on how to test this, as it involves manually deleting your existing registration keys, and that info is not something I want to put into a public repository.

The change here is to make a single call to orchestration's profile service, instead of one call per profile key.

I'll clean up the orchestration side of things in a follow-on story, when I tackle DSDEEPB-2046.


rushtong:
:+1: Perfect!

PR #308, opened 2015-11-23
DSDEEPB-2095: key summary to cause refresh

MattPutnam:


davidangb:
Functionally, to an end user, what does this change in the UI?

MattPutnam:
It causes the summary view to do a full refresh post-clone.

davidangb:
super simple code change, works great for me functionally. :+1:

PR #309, opened 2015-11-23
Remove unnecessary encoding

rushtong:
The encoded protocol + location breaks local https development. Without that, both http and https sign-ins work as expected.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -175,8 +175,6 @@
        [comps/Button
         {:text "Sign In"
          :href (str "/service/login?callback="
-                    (js/encodeURIComponent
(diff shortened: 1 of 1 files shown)


davidangb:
But doesn't this break being able to run a local UI, talking to a dev instance of orchestration?

rushtong:
I didn't try that combination ... doing that now.

rushtong:
With this change in place, I can still run a local UI against a remote dev. The one thing that is unusual is that if you start out with http://firecloud.dsde-dev.broadinstitute.org, after signing in, you will be at https://firecloud.dsde-dev.broadinstitute.org. When running both UI+Orch locally, you will stay on the same protocol after sign-in.

rushtong:
No, this does not work in that case. It instead redirects you to the dev instance. Closing and will look at another way around this problem.

PR #312, opened 2015-11-24
DSDEEPB-1494: Select Google project from list of available

MattPutnam:
Also includes a possibly better pattern for loading multiple things at once.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
@@ -35,7 +35,8 @@
               (when (:creating-wf @state)
                 [comps/Blocker {:banner "Creating Workspace..."}])
               (style/create-form-label "Google Project")
-              (style/create-select {:onChange #(swap! state assoc :selected-project (-> % .-target .-value))}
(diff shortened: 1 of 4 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -16,47 +16,60 @@
...
+       (react/create-element
+         [dialog/OKCancelForm
+          {:header "Create New Workspace"
+           :content
+           (react/create-element
+             [:div {:style {:marginBottom -20}}
+              (when (:creating-wf @state)
+                [comps/Blocker {:banner "Creating Workspace..."}])
+              (style/create-form-label "Google Project")
+              (style/create-select {:onChange #(swap! state assoc :selected-project (-> % .-target .-value))}
```
dmohs: I think you should also include `:value (:selected-project @state)` here. Otherwise, if we ever change `:get-initial-state` to be something besides the first one, this will break.

dmohs:
:+1: Looks good. Please merge after minor fix.

dmohs:
Also, I do like your new loading pattern.

PR #318, opened 2015-12-01
DSDEEPB-2101: break out ErrorViewer, CauseViewer, StackTraceViewer into separate components

davidangb:
to avoid the JS error mentioned in the bug.

Note: there are still other problems with the ErrorViewer:
1) The "Show Stack Trace" and "Hide Cause" links run up against each other visually, in an ugly way
2) There is a similar but not identical JS problem I found during copying data which prevents the modal from being closed.

I will file separate tickets for both of these - I'd rather get this refactoring in "clean" without those changes.

update: DSDEEPB-2187 tracks item 2 above. I decided to just fix item 1 as part of this PR because it's so small; I've done that as a separate commit to keep the code changes cleaner.


MattPutnam:
:+1:

PR #319, opened 2015-12-02
DSDEEPB-1534: Initial work on workspace navigation

MattPutnam:
This PR doesn't complete the story, so don't move the JIRA item through the board.  I'd like to get this in as a major checkpoint, and to let people start looking at it.

-----

Made a breadcrumb chain along the top, which threads through to all views except submission details.  This involves using the browser hash instead of state variables to control the view.

Known shortcomings:
* ~~Second link (pointing to workspace namespace:name) is weird--clicking on it will direct the browser to the summary page, but without "/Summary" in the URL or breadcrumbs.  Doesn't break anything but is sort of wonky.~~ **Edit:** second commit fixes this.
* Should also be implemented for submission details
* style/create-link is left in limbo.  The finishing PR will clean this up.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -126,3 +126,8 @@
 (defn parse-profile [unparsed-profile]
   (let [unparsed-values (get unparsed-profile "keyValuePairs")]
     (into {} (map (fn [m] [(keyword (m "key")) (m "value")]) unparsed-values))))
+
+(defn get-id-from-nav-segment [segment]
+  (when-not (clojure.string/blank? segment)
+    (let [[ns n] (clojure.string/split segment #":")]
+      {:namespace ns :name n})))
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -268,14 +268,12 @@
 
 
 (defn- validate-crumb [crumb]
-  (and (contains? crumb :text)
-       (or (contains? crumb :on-click)
-           (contains? crumb :href))))
+  (contains? crumb :text))
 
 (react/defc Breadcrumbs
   {:push
    (fn [{:keys [state]} new-crumb]
-     (assert (validate-crumb new-crumb) "Crumb must have :text and :on-click or :href")
+     (assert (validate-crumb new-crumb) "Crumb must have :text")
      (swap! state update-in [:crumbs] conj new-crumb))
    :set-crumbs
    (fn [{:keys [state]} crumbs]
@@ -284,7 +282,7 @@
    :get-initial-state
    (fn [{:keys [props]}]
      (assert (every? validate-crumb (:initial-crumbs props))
-       "Each initial crumb must have :text and :on-click or :href")
+       "Each initial crumb must have :text")
      {:crumbs (vec (:initial-crumbs props))})
    :render
    (fn [{:keys [state]}]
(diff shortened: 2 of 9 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -126,3 +126,8 @@
 (defn parse-profile [unparsed-profile]
   (let [unparsed-values (get unparsed-profile "keyValuePairs")]
     (into {} (map (fn [m] [(keyword (m "key")) (m "value")]) unparsed-values))))
+
+(defn get-id-from-nav-segment [segment]
+  (when-not (clojure.string/blank? segment)
+    (let [[ns n] (clojure.string/split segment #":")]
```
davidangb: many things break if I create a workspace with a colon in its name ... but this is another one of them. (probably best to prevent those characters in workspace names instead of fixing here)

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -259,3 +265,41 @@
...
+           crumbs (:crumbs @state)]
+       (case (count crumbs)
+         0 [:div {}]
+         1 [:div {} (:text (first crumbs))]
+         [:div {}
+          (interpose sep
+            (map-indexed
+              (fn [index {:keys [text on-click href]}]
+                (if (or on-click href)
+                  (style/create-link2 {:href href :text text
```
davidangb: validate-crumb checks that breadcrumbs must contain :text, but it looks like you're also counting on it to contain :href.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -208,30 +208,41 @@
...
-  (when-not (clojure.string/blank? segment)
-    (let [[ns n] (clojure.string/split segment #":")]
-      {:namespace ns :name n})))
+(defn- create-breadcrumbs-from-hash [hash]
+  (let [segments (split hash #"/")]
+    (map-indexed
+      (fn [index segment]
+        (case index
+          0 {:text "Workspaces" :href "#workspaces"}
+          1 {:text (replace (js/decodeURIComponent segment) ":" "/")}
```
davidangb: Without an :href here, there's no way to use the breadcrumbs to get back to the workspace summary? Is that the wonky second link issue you mentioned?

MattPutnam: Before I special-cased out that segment and got rid of the link, you could click on that, and it would nav you to workspaces/namespace:name, without the "/Summary" on the end.  It would still work--you could click on tabs and the problem would correct itself, and if you bookmarked that URL and hit it it would fix on load (redirecting to /Summary).

You can always click on the tab, so I thought having that crumb navigable was just causing more problems than it was worth.

src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
```diff
@@ -86,6 +86,13 @@
...
    children])
 
+(defn create-link2 [{:keys [href style onClick text]}]
+  [:a (merge
+        {:href (or href "javascript:;")
+         :style (merge {:textDecoration "none" :color (:button-blue colors)} style)}
+        (when onClick {:onClick onClick}))
+   text])
+
 (defn create-link [onClick & children]
```
davidangb: yup, need to figure out how to consolidate create-link and create-link2, as you said you'll do in a follow-on PR

MattPutnam: I'm actually just going to replace all usages of the old one with the new one (and then rename).  But that would touch almost every UI page file.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
davidangb: Can this be solved somehow in the workspace summary page, by somehow defaulting to the Summary tab? Fine to do this in a follow-on PR, but I don't like having the hole in the breadcrumbs - it even looks inconsistent visually.

davidangb:
:+1: lots of stuff noted for follow-on PRs, but otherwise working well and code looks good. I'm sure we may find functional corner cases that need to be addressed later on, but this is a good interim PR.

PR #320, opened 2015-12-03
Add gs-dowload dialog to the task input/output display.

rushtong:
To test, you will need a successfully run submission. I can share my OncotatorTestClone workspace with anyone that wants to see it.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/dialog.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/dialog.cljs
@@ -91,7 +91,7 @@
    (fn [{:keys [props state this]}]
      (assert (:bucket-name props) "No bucket name provided")
      (assert (:object props) "No GCS object provided")
-     [:div {}
+     [:div (if-let [style-props (:style-props props)] style-props {})
       [:a {:href "javascript:;"
            :onClick #(react/call :show-dialog this)}
        (:gcs-uri props)]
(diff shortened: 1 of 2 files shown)


davidangb:
This is mostly working for me. However, it changes the UI layout of the input/output display. Previously, the keys and values were on the same line; now the value (i.e. the link) is on a new line, and I think it looks worse.

This is happening because the dialog/GCSFilePreviewLink automatically wraps the link in a div. At minimum, I suggest modifying (defc GCSFilePreviewLink ) to use a span at line 94 instead of a div, to prevent wrapping. The only other place GCSFilePreviewLink is used, I think, is in the entity table, and a span should not cause a regression. A more ideal solution would be to allow callers of GCSFilePreviewLink to pass in CSS styling, and have the input/output display pass in {:display "inline"}.

Finally, how about adding the GCSFilePreviewLink to the stderr/stdout links in CallDetail?

rushtong:
All good points - will revise.

rushtong:
Moving stdout and stderr to full links makes the screen look a little less consistent. I think those two links should be moved to the Output section so they display similarly to the others.

Screenshot:
screen shot 2015-12-03 at 2 09 44 pm

davidangb:
Looks good/works good. Only tweak I see - and I should have called this out earlier - is that by losing the download="stderr" and download="stdout" attributes on the stdout/stderr links, the names of the downloaded files are now different (and much longer). Is this something we care about? I like the long, explicit names - harder to confuse between jobs - but they might be too long.

PR #321, opened 2015-12-03
Redirect user to home page 

rushtong:
If an unauthorized message comes back from any ajax call. This is a temporary solution and needs further work to make use of the ability to fully logout a user. Also, this does not address the problem of authenticated user with disabled LDAP settings getting an unauthenticated error.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
@@ -88,12 +88,15 @@
           call-on-done (fn []
                          (let [status-code (.-status xhr)
                                get-parsed-response #(parse-json-string (.-responseText xhr))]
-                           (on-done {:xhr xhr
-                                     :status-code status-code
-                                     :success? (and (>= status-code 200)
-                                                    (< status-code 300))
-                                     :status-text (.-statusText xhr)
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -88,12 +88,15 @@
...
                                get-parsed-response #(parse-json-string (.-responseText xhr))]
-                           (on-done {:xhr xhr
-                                     :status-code status-code
-                                     :success? (and (>= status-code 200)
-                                                    (< status-code 300))
-                                     :status-text (.-statusText xhr)
-                                     :get-parsed-response get-parsed-response})))]
+                           ; TODO: Fix this with a real log-out once the login bug is fixed and logout is implemented.
+                           (if (= status-code 401)
+                             (set! (-> js/window .-location) (str js/window.location.protocol "//" js/window.location.hostname))
```
davidangb: could this be simplified to:

(set! (-> js/window .-location) "/")

otherwise, looks good!

rushtong: I will give that a shot. The hard part is just testing the timeout :-)

PR #323, opened 2015-12-03
DSDEEPB-1523 Part 2/3 (and DSDEEPB-1923, DSDEEPB-2000): finish nav work in workspace tabs

MattPutnam:
STILL not 100% done, so again don't move the JIRA story.  This at least mostly finishes the work to be done with the workspace tab view navigation.  It also fixes DSDEEPB-1923, which was causing problems with the new rendering scheme.  Also, DSDEEPB-2000 is obviated by this.

Two issues outstanding:
* Data is being loaded twice when each tab is selected.  Probably has something to do with the way the tab bar is getting rendered.  Taking the `:key` off of the tab bar fixes this, but then causes a ton of other problems.
* The second breadcrumb (showing the workspace project/name) is deliberately unclickable.  It should ideally be a link to /workspaces/[workspace-id]/Summary, but this clashes with the way I've designed the breadcrumb component:  When a link is clicked it automatically pops off the remaining segments.  So if you give it the proper href (by adding `:href (str "#workspaces/" segment "/Summary")` to the map on workspaces_list.cljs:217), when you click it, the "Summary" segment pops off.

Part 3 will add this breadcrumb component to the workspace data import dialog, and possibly to the method config importer, as well as hopefully fixing the two issues above.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -96,11 +96,8 @@
                                 (when f (f e))))}])
            (:items props))
          (common/clear-both)]
-        (let [active-item (nth (:items props) (:active-tab-index @state))
-              render (:render active-item)]
-          (if (fn? render)
-            [:div {} (apply render (.-renderArgs this))]
-            [:div {} render]))])
+        (let [active-item (nth (:items props) (:active-tab-index @state))]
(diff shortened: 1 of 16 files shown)


dmohs:
Code looks good but I'm scared to merge this with the double-load problem. Maybe fix first?

MattPutnam:
Well... the fix for that is buried in my commit history for Part 3/3.

PR #324, opened 2015-12-07
Add logic for handling incorrect 401s from Rawls proxy

rushtong:
In the case that Rawls sends back a 401 unauthorized message, that could be incorrect if it is a logged in user whose administrative ldap activation has not taken place yet. This branch tries to capture that case and provide an "Access Disabled" message with a 403 header. 

There is a corresponding Orchestration PR (https://github.com/broadinstitute/firecloud-orchestration/pull/172) that is required for testing this feature. To fully test, log in with a new user that has not been ldap-enabled. 

Would love to see a better organization/approach for the nested ajax call.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
@@ -40,6 +40,11 @@
   (js->clj (js/JSON.parse x)))
 
 
+(defn keywordize-keys [m]
+      (into {} (map (fn [[k v]]
+                        [(keyword k) (if (map? v) (keywordize-keys v) v)])
+                    m)))
+
 (defn local-storage-write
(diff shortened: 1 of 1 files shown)


MattPutnam:
:+1:

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -40,6 +40,11 @@
   (js->clj (js/JSON.parse x)))
 
 
+(defn keywordize-keys [m]
+      (into {} (map (fn [[k v]]
+                        [(keyword k) (if (map? v) (keywordize-keys v) v)])
+                    m)))
+
```
rushtong: I should have added ... this code snippet was shamelessly stolen from @MattPutnam

PR #328, opened 2015-12-08
Make empty description placeholder oblique

MattPutnam:
This is some styling that got lost by a table column revamp.


rushtong:
I remember writing the code that removed this - glad you're around to fix my mistakes!

rushtong:
I also remember that there didn't seem to be any effect of the oblique font. The oblique descriptions look exactly the same as the non-oblique, despite the fact that it is clearly there in the generated html.
screen shot 2015-12-08 at 1 33 28 pm

MattPutnam:
Interesting, it definitely works in my browser.  Mine is a completely vanilla install--have you done any mucking with settings?

rushtong:
Just updated my previous comment ... I only see it in Firefox.

rushtong:
:+1:

PR #332, opened 2015-12-08
DSDEEPB-2247: show ErrorViewer when profile fails to save

davidangb:
This is in service of DSDEEPB-2247, though it doesn't solve that issue.

Previously, if we encountered an error:
* saving the user profile to Thurloe
* checking for a pre-existing user registration in rawls
* registering the user in rawls
we would display a simple "Profile failed to save" error to the end user.

Now, we add the ErrorViewer to show the error details. This (hopefully) will help end users in debugging any problems they have with registration.


davidangb:
Updated this PR. With orchestration's https://github.com/broadinstitute/firecloud-orchestration/pull/173, we can now rely on server-side errors entirely, so I've pulled out the hardcoded error message in the UI.

MattPutnam:
:+1:

PR #334, opened 2015-12-09
DSDEEPB-2277: open some files in browser instead of downloading

davidangb:
I considered changing the "Open" link to use style/create-link, but decided not to - I like that the link stands out a bit compared to the "More Info" link.


MattPutnam:
:+1: Nice, I like this better.

PR #335, opened 2015-12-10
Create redact dialog with error message for methods/configs in method repo

MattPutnam:
A few things came up in this:

* The code that @rushtong added for checking 401s seemed to have a bug.  I think it's working right now, but whoever reviews this should take an especially close look at the changes to utils.
* Rawls/Agora return a nice JSON structure for error messages, with "source", "statusCode", "causes", "stackTrace" and "message" fields.  This is what gets passed to the comps/ErrorViewer.  However, the "unauthorized" message is coming back with just "code" and "message".  So an extra field check was added to comps/ErrorViewer, but I suppose it would be better to make the JSON responses consistent.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/redact.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/redact.cljs
@@ -1,0 +1,44 @@
+(ns org.broadinstitute.firecloud-ui.page.method-repo.redact
+  (:require
+    [dmohs.react :as react]
+    [org.broadinstitute.firecloud-ui.common :as common]
+    [org.broadinstitute.firecloud-ui.common.components :as comps]
+    [org.broadinstitute.firecloud-ui.common.dialog :as dialog]
+    [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
+    ))
+
+
+(react/defc Redacter
+  {:render
+   (fn [{:keys [props state this]}]
+     [dialog/Dialog
+      {:width 550
+       :dismiss-self (:dismiss-self props)
+       :content
+       (react/create-element
+         [dialog/OKCancelForm
+          {:header "Confirm redact"
+           :dismiss-self (:dismiss-self props)
+           :content
+           [:div {}
+            (when (:redacting? @state)
+              [comps/Blocker {:banner "Redacting..."}])
+            [:p {} (str "Are you sure you want to redact this method"
+                     (when (:is-config? props) " configuration") "?")]
+            [comps/ErrorViewer {:error (:error @state)
+                                :expect {401 "Unauthorized"}}]]
+           :ok-button [comps/Button {:text "Redact" :onClick #(react/call :redact this)}]}])}])
+   :component-did-mount
+   (fn []
+     (common/scroll-to-top 100))
+   :redact
+   (fn [{:keys [props state]}]
+     (swap! state assoc :redacting? true :error nil)
+     (let [[namespace name snapshotId] (map #(get-in props [:entity %]) ["namespace" "name" "snapshotId"])]
+       (endpoints/call-ajax-orch
+         {:endpoint (endpoints/delete-agora-entity (:is-config? props) namespace name snapshotId)
+          :on-done (fn [{:keys [success? get-parsed-response]}]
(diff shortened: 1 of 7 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -94,34 +99,34 @@
...
+                                   (let [us-status (.-status us-xhr)
+                                         parsed-us-response (parse-json-string (.-responseText us-xhr))
+                                         user-info (keywordize-keys parsed-us-response)]
+                                     (cond
+                                       (and (= us-status 200) (not (:ldap (:enabled (:userInfo user-info)))))
+                                       (on-done {:xhr (adjust-xhr xhr)
+                                                 :status-code 403
+                                                 :success? false
+                                                 :status-text "Access Disabled"
+                                                 :get-parsed-response get-parsed-response})
```
rushtong: I'm not following this. In this case, `get-parsed-response` is always the generic 401 Unauthorized message. I think it makes more sense to return the response from the secondary user-status request because that has real actionable information. The 401 response can sometimes come from the proxy and is pretty much useless.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -94,34 +99,34 @@
...
+                                       ;TODO: Fix this with a real log-out once the login bug is fixed and logout is implemented.
+                                       (= us-status 401) (set! (-> js/window .-location) "/")
+                                       :else (on-done {:xhr xhr
+                                                       :status-code status-code
+                                                       :success? false
+                                                       :status-text (.-statusText xhr)
+                                                       :get-parsed-response get-parsed-response})))))
```
rushtong: In this case, `status-code` is always 401 and `get-parsed-response` is the default unauthorized text which is misleading. I think it makes more sense to return the info from the secondary user-status request.

src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/redact.cljs
```diff
@@ -0,0 +1,44 @@
...
+          {:header "Confirm redact"
+           :dismiss-self (:dismiss-self props)
+           :content
+           [:div {}
+            (when (:redacting? @state)
+              [comps/Blocker {:banner "Redacting..."}])
+            [:p {} (str "Are you sure you want to redact this method"
+                     (when (:is-config? props) " configuration") "?")]
+            [comps/ErrorViewer {:error (:error @state)
+                                :expect {401 "Unauthorized"}}]]
```
rushtong: I think this is not the right logic. If rawls/agora are returning a 401, they should instead be returning a 403 forbidden. 401 means the user's authentication is not valid.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
rushtong: To clarify, all 401 responses *should* come from the proxy and generally not from rawls.

rushtong:
:+1: in general - code works for me. I have some comments that I think need to be addressed, but might need further discussion with the team and/or rawls/agora teams. Back to you @MattPutnam for any re-work you'd like to do.

MattPutnam:
Well... I'm trying to display an error message, but I'm getting a 200 back.  So I really don't think we want to be returning the user status response in any case.

MattPutnam:
I've been mucking around with this and everything feels awful.  If Rawls/Agora are returning 401 when they should be returning 403, then let's get that fixed instead.

rushtong:
I agree, this is painful. There are three cases:

* Agora returns a 401. Brad is aware of that problem and has a story for changing them to 403's.
* Rawls proxy returns a 401 when the authenticated user is not in LDAP. That is the primary case this code trues to address. 
* Rawls returns a 401 instead of a 403. I'm not sure where this happens, but we do need to ping the Blue team on this. Alex is aware, but I don't know of a story on the subject.

Lastly, David An and I were brainstorming and we thought that we might be better served by an orchestration service that can be queried for the right status. That way, we could limit the amount of logic required here.

MattPutnam:
I'm gonna close this then.  I don't like how this is working right now.

PR #336, opened 2015-12-10
DSDEEPB-2037: Only show the GCS object in the table

MattPutnam:
Also fixed a bug regarding sorting a data table column that contains a map reference to another entity


rushtong:
:+1: Nice!

PR #337, opened 2015-12-11
When a user is not found in Rawls

rushtong:
We redirect them to the profile page so that registration can be completed.


dmohs:
We should re-think this a bit. With this code in place, a user might be navigating around a workspace and suddenly be placed on their profile page. This would be jarring. If you want to punt, please create a story to address this before merging.

rushtong:
Matt Putnam is handling this case with an updated 401 check

PR #338, opened 2015-12-16
DSDEEPB-2298: on 401 ajax responses, don't try to parse as json always

davidangb:
this is an ugly temporary fix; I think this needs a more elegant refactoring, but I want to land something quick for DSDEEPB-2298, then refactor later.

If the user's access token had timed out - which requires waiting a full hour without using the system - the proxy will respond with a 401 (yes, I always saw a 401), but the body of the response is an HTML page. When our ajax wrapper detected a 401, it then fired off a request to /service/register, attempted to parse the response as json, and inspect the keys of the response. However, /service/register would *also* respond with a 401 html page, and our json parsing failed, leading to JS errors in console and a hung UI.

With my changes, we don't attempt to parse the json until we know that /service/register/ returned a 200. This leads to ugly nested conditionals; this is where I want to refactor later.

I also updated the handling for when we see a 401 from /service/register - I now delete our local cookie and reload the page instead of using location="/".

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
@@ -75,6 +75,15 @@
 
 (def access-token (atom nil))
 
+(defn set-access-token-cookie [token]
+      (.set goog.net.cookies "FCtoken" token 300))
+
+(defn get-access-token-cookie []
+      (.get goog.net.cookies "FCtoken"))
+
+(defn delete-access-token-cookie []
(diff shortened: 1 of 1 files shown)


dmohs:
Change looks fine to me. Feel free to merge, unless you want to address the technical debt: `ajax`, a generic function, now makes up to two ajax calls, one of which has a circular dependency on endpoints.cljs which has been masked by copying the code and URL ("/service/register") directly into the function.

davidangb:
I'm gonna merge and create a new story for refactoring - thanks!

PR #340, opened 2015-12-21
Link NIH account (DSDEEPB-1925)

dmohs:
See individual commits. This will require a change to the Jenkins builds:
```bash
cp src/main/config/dsde-$ENV/config.json target/config.json
```

### DIFF
--- a/script/dev/create-config.sh
+++ b/script/dev/create-config.sh
@@ -11,3 +11,5 @@ fi
 
 docker run --rm -it -v "$PWD":/working broadinstitute/dsde-toolbox \
   render-templates.sh local $VAULT_TOKEN
+
+cp src/main/config/dsde-local/config.json target
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -1,4 +1,7 @@
-(ns org.broadinstitute.firecloud-ui.common)
+(ns org.broadinstitute.firecloud-ui.common
+  (:require
+   [org.broadinstitute.firecloud-ui.utils :as utils]
+   ))
 
 
 (def keymap
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
@@ -5,6 +5,7 @@
    [org.broadinstitute.firecloud-ui.common :as common]
    [org.broadinstitute.firecloud-ui.common.components :as components]
    [org.broadinstitute.firecloud-ui.common.style :as style]
+   [org.broadinstitute.firecloud-ui.config :as config]
    [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
    [org.broadinstitute.firecloud-ui.nav :as nav]
    [org.broadinstitute.firecloud-ui.utils :as utils]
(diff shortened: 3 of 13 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -1,4 +1,7 @@
-(ns org.broadinstitute.firecloud-ui.common)
+(ns org.broadinstitute.firecloud-ui.common
+  (:require
+   [org.broadinstitute.firecloud-ui.utils :as utils]
```
davidangb: is this used anywhere?

dmohs: No. It's just annoying to add this line when you want to use the debugging tools (which is exactly what happened and why I left it in).

src/main/config/dsde-alpha/config.json
```diff
@@ -0,0 +1 @@
+{"shibbolethUrlRoot": "https://shibboleth.dsde-aplha.broadinstitute.org"}
```
davidangb: typo - "alpha" vs "aplha"

script/dev/create-config.sh
```diff
@@ -11,3 +11,5 @@ fi
 
 docker run --rm -it -v "$PWD":/working broadinstitute/dsde-toolbox \
   render-templates.sh local $VAULT_TOKEN
+
+cp src/main/config/dsde-dev/config.json target
```
davidangb: I don't feel super strong about this, but I'd prefer to add a src/main/config/dsde-local directory, so that this script only concerns itself with the "local" environment. I realize the config in dsde-local would point at https://shibboleth.dsde-dev.broadinstitute.org.

src/cljs/org/broadinstitute/firecloud_ui/common.cljs
davidangb: ah, okay :)

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -32,13 +36,58 @@
...
+                                  "/nih-username-token={token}")))}
+             "Log-In to NIH to link your account"]
+            :else
+            linkedNihUsername)]]
+        [:div {:style {:display "flex" :marginTop "1em"}}
+         [:div {:style {:flex "0 0 20ex"}} "dbGaP Authorization:"]
+         [:div {:style {:flex "0 0 auto"}}
+          (if (nil? linkedNihUsername)
+            [:i {} "N/A"]
+            (if isDbgapAuthorized
```
davidangb: when I set my profile in Thurloe to isDbgapAuthorized=false, this still renders "Authorized" in the UI.

Possibly this is a problem somewhere in js->cljs truthy/falsey? Remember that Thurloe stores and retrieves a string value, so the profile is going to return as:
{
    "key": "isDbgapAuthorized",
    "value": "false"
  }

davidangb:
A few minor comments inline. Looks good other than that.

This PR stops short of sending the JWT to orchestration in order to create the link, correct? It only links to NIH to start the process and displays information about an existing link, but it won't make a new one?

src/main/config/dsde-alpha/config.json
dmohs: Thank you.

script/dev/create-config.sh
dmohs: I completely agree. Done.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
dmohs: Thanks for catching that. Fixed.

dmohs:
Correct. Sending the link will be a follow-up PR.

PR #342, opened 2016-01-04
fix regression with react upgrade

davidangb:


dmohs:
:+1:

https://github.com/facebook/react/blob/3b96650e39ddda5ba49245713ef16dbc52d25e9e/src/renderers/dom/cl...

and

```javascript
> document.createElement('input').nodeType === 1
true
```

PR #343, opened 2016-01-04
Call orchestration endpoint to link NIH account

dmohs:


davidangb:
:+1: merging. We probably need to add a feature to allow a user to explicitly *unlink* their NIH account - there is no way to do so currently without manually poking Thurloe's data.

PR #344, opened 2016-01-05
DSDEEPB-2284: create/clong workspace billing project dropdown

davidangb:
We need to send the value, not the index, to rawls when creating/cloning a project.


rushtong:
:+1: Merging.

PR #347, opened 2016-01-07
Improve 401 check to use new /me endpoint

MattPutnam:


davidangb:
do you have repro steps to show before/after behavior?

rushtong:
Some comments wrt testing:
* Logging in with a new user with no rawls registration should redirect one to the profile page so they can register. This was also working before.
* Logging in with a registered/not activated user should show an error message saying that on the workspace page. This was showing an OK message - I think because the rawls response had changed.
* Getting a 401 from rawls when trying to edit method permissions (when a user doesn't have access) shows an Unauthorized message. This was failing and getting the wrong response back from the 401 handling ajax code.

davidangb:
Love it, looks good and works good.

Only question I have is that I see you are ignoring 500 responses from /me. I went back and forth on this - if you got a response from the original request, but /me is throwing an error, what should we do? You're passing the original request back, which is probably fine - but was this an intentional choice?

If that's what you intended, feel free to merge yourself, or respond here and I'll merge.

rushtong:
I don't think we considered the /me 500 case. The intent was to try and minimize what we intercepted and only handle the real cases we knew we could redirect. 401, 403, and 404 are the cases where we have a known action for. Anything else should (in theory) be returned as the original request. I still think that's valid - if we do get a 500 from /me, the original response seems to me to be more useful than the 500 response.

davidangb:
ok, works for me - I'll call that a :+1: .

PR #348, opened 2016-01-07
DSDEEPB-1929: use dropdown for root entity type

davidangb:
Manually tested:
* editing a config that pre-existed in your workspace
* importing a method from agora
* importing a config from agora

(note that in the last use case, you do not have the option of changing the root entity type; I tested this to ensure there was no regression)

At the code level, I don't like having both style/create-select and style/create-identity-select; these could be refactored back into a single function, but doing so opens a large-ish risk of regression.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -131,3 +131,6 @@
   (when-not (clojure.string/blank? segment)
     (let [[ns n] (clojure.string/split segment #":")]
       {:namespace ns :name n})))
+
+(def root-entity-types
+  '("participant","sample","pair","participant_set","sample_set","pair_set"))
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -133,4 +133,4 @@
       {:namespace ns :name n})))
 
 (def root-entity-types
-  '("participant","sample","pair","participant_set","sample_set","pair_set"))
+  ["participant" "sample" "pair" "participant_set" "sample_set" "pair_set"])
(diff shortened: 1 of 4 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -131,3 +131,6 @@
   (when-not (clojure.string/blank? segment)
     (let [[ns n] (clojure.string/split segment #":")]
       {:namespace ns :name n})))
+
+(def root-entity-types
+  '("participant","sample","pair","participant_set","sample_set","pair_set"))
```
MattPutnam: Can this just be a vector (not a list)?  Also, drop the commas.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
```diff
@@ -51,10 +51,15 @@
...
          [:div {:style {:float "left" :marginRight "0.5em"}}
-          (style/create-form-label (:label field))
-          [input/TextField {:defaultValue (entity (:key field))
-                            :ref (:key field) :placeholder "Required"
-                            :predicates [(input/nonempty "Fields")]}]])
+            (style/create-form-label (:label field))
+              (cond (= (:type field) "identity-select")
+                (style/create-identity-select {:ref (:key field)
+                  :value (entity (:key field))}
+                  root-entity-types)
```
MattPutnam: Should this be `(:options field)`?

MattPutnam:
Check indentation throughout.  Otherwise, just two little things.

src/cljs/org/broadinstitute/firecloud_ui/page/method_config_importer.cljs
davidangb: yes. yes it should.

davidangb:
back to you, Matt, and thanks!

MattPutnam:
:+1:

PR #349, opened 2016-01-07
DSDEEPB-2047: Implement full dialog for method/config redact, with error viewer for 401s

MattPutnam:
Also some minor code cleanup and UI styling


davidangb:
:+1: working for me!

PR #350, opened 2016-01-08
killed box-shadow

helgridly:
Removed box-shadow from table paginator because it looks weird. It's also not used anywhere else except text boxes, and only very subtly -- so I removed it there too.

This code is released under the terms of the [Unlicense](http://unlicense.org/) (i.e. all rights waived, all uses permitted, software released as-is, no holding me liable).


MattPutnam:
This came from the original PixelMedia designs.  I'm okay with removing it, I think it looks fine either way.

PR #351, opened 2016-01-08
Made last breadcrumb element sane size

helgridly:
It's never styled EXTRA BIG on normal websites.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -272,5 +272,5 @@
                    text)])
               (butlast crumbs)))
           sep
-          [:span {:style {:verticalAlign "middle"}}
+          [:span {:style {:fontSize "60%" :verticalAlign "middle"}}
            (:text (last crumbs))]])))})


davidangb:
sizing was intentional and reviewed by PO. Is this in response to a specific request?

helgridly:
No. I was offended by the styling and made a drive-by PR.

MattPutnam:
The history of this:

We used to just display "Workspaces" or "Workspaces > [project]/[name]" in that big font.  Then we wanted further breadcrumbs through the tabs (and corresponding browser hashing), which I added, but keeping the whole thing at that big font made the breadcrumbs too long, even at full screen.  But reducing the font size looked weird, especially when just at the "Workspaces" level.  This was my way to reconcile those two things.

I also object to "other websites don't do this" as a sole reason to change things.

Finally, if we really want to change this, the proper way would be to get rid of the existing `:fontSize "60%" on the all-but-last crumbs, and reduce the size where the component is used.

helgridly:
> I also object to "other websites don't do this" as a sole reason to change things.

Well, I'm going for "other websites don't do this" and "it looks weird and ugly", so that's two reasons ;) In seriousness though: principle of least surprise. Users have expectations set from prior experiences and I'd argue that being consistent with those expectations should be the default, all other things being equal.

That said, I'm not going to fight about this. Feel free to close the PR without merging if you don't want it.

PR #352, opened 2016-01-08
NIH account link timeout warning

dmohs:
Also some minor UI fixes

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -8,6 +8,7 @@
    [org.broadinstitute.firecloud-ui.config :as config]
    [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
    [org.broadinstitute.firecloud-ui.nav :as nav]
+   [org.broadinstitute.firecloud-ui.nih-link-warning :as nih-link-warning]
    [org.broadinstitute.firecloud-ui.page.method-repo :as method-repo]
    [org.broadinstitute.firecloud-ui.page.profile :as profile-page]
    [org.broadinstitute.firecloud-ui.page.status :as status-page]
@@ -105,7 +106,8 @@
       (map (fn [item] [TopNavBarLink {:name (item :name)
                                       :selected (= (:selected-item props) (item :key))
                                       :onClick (fn [e] ((:on-nav props) (item :key)))}])
-           top-nav-bar-items)])})
+           top-nav-bar-items)
+      [nih-link-warning/NihLinkWarning]])})
 
 
 (react/defc LoggedIn
--- a/src/cljs/org/broadinstitute/firecloud_ui/nih_link_warning.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/nih_link_warning.cljs
@@ -16,22 +16,21 @@
                                     :responseText (.-responseText xhr)})
                     (on-done m)))]
     (utils/ajax-orch
-     "/profile"
-     (merge ajax-args {:on-done on-done})
-     :service-prefix "/service/register")))
+     "/nih/status"
+     (merge ajax-args {:on-done on-done}))))
 
 
 (react/defc NihLinkWarning
   {:render
    (fn [{:keys [props state]}]
-     (when-let [profile (:profile @state)]
-       (let [expire-time (-> (:linkExpireTime profile) js/parseInt (* 1000) (js/moment.))
+     (when-let [status (:status @state)]
+       (let [expire-time (-> (get status "linkExpireTime") (* 1000) (js/moment.))
              _24-hours-from-now (.add (js/moment.) 24 "hours")]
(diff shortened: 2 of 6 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
```diff
@@ -1,5 +1,6 @@
 (ns org.broadinstitute.firecloud-ui.page.workspace.summary.attribute-editor
   (:require
+    clojure.set
```
davidangb: is this necessary?

src/cljs/org/broadinstitute/firecloud_ui/nih_link_warning.cljs
```diff
@@ -0,0 +1,45 @@
...
+(defn- get-profile [ajax-args on-done]
+  (let [on-done (if (or (nil? (:canned-response ajax-args)) (:use-canned-response? ajax-args))
+                  on-done
+                  (fn [{:keys [xhr] :as m}]
+                    (utils/cljslog {:status (.-status xhr)
+                                    :statusText (.-statusText xhr)
+                                    :responseText (.-responseText xhr)})
+                    (on-done m)))]
+    (utils/ajax-orch
+     "/profile"
```
davidangb: ah. I was hoping you would use the /nih/status endpoint instead of /profile, so we can centralize the logic for timing into a single place - otherwise, we have that logic duplicated (and therefore possibly out of sync) across UI and orch.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -105,7 +106,8 @@
       (map (fn [item] [TopNavBarLink {:name (item :name)
                                       :selected (= (:selected-item props) (item :key))
                                       :onClick (fn [e] ((:on-nav props) (item :key)))}])
-           top-nav-bar-items)])})
+           top-nav-bar-items)
+      [nih-link-warning/NihLinkWarning]])})
```
davidangb: Personally, I liked it better at the top of the window! I'm fine with either.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -12,6 +12,13 @@
    ))
 
 
+(defn get-nih-link-href []
+  (str (get @config/config "shibbolethUrlRoot")
+       "/link-nih-account?redirect-url="
+       (js/encodeURIComponent
+        (str (-> js/window .-location .-href) "/nih-username-token={token}"))))
+
+
```
davidangb: Something isn't quite right here ... I re-linked my account and ended up at:
firecloud___broad_institute

davidangb:
Much of this is in good shape, but I have two blocking issues:
1) See the inline comment - after re-linking, I am not brought to the right place
2) I hope you will use a different endpoint for checking link status. Let's talk in person (hipchat, whatever) if you have concerns about that.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
dmohs: Yeah it's used in the file. I wish compilation was more strict about dependencies.

src/cljs/org/broadinstitute/firecloud_ui/nih_link_warning.cljs
dmohs: Done. Note that I would not tend to use something like `descriptionUntilExpires` in UI code because it makes it unnecessarily difficult to localize down the road. It's the same reason I prefer error codes to error messages returned from API endpoints.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Looks-wise, I preferred it at the top. Because it is such an important message, I agree with Hussein that it could be easy to miss there.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
dmohs: Thanks for checking that. Fixed.

davidangb:
Looks like there is still a functional issue here.
1) Set your profile keys such that the warning is displayed.
2) Click the link in the warning to re-link your profile, and authenticate with NIH.
3) Get redirected back to the UI. At this point, notice that the warning banner still appears, showing your previous expire time. But if you query the DB directly, you'll see that the key has been updated.
4) Reload the page and notice that the banner disappears.

I haven't tried to debug this, so I don't know why this is happening ...

davidangb:
Other than the functional issue I just mentioned, code looks good. We might want to trap for and display a more stringent warning if the user's link has already expired, but that can be done in followup.

davidangb:
:+1:

PR #354, opened 2016-01-08
DSDEEPB-2048: Overhaul method/config ACL editor

MattPutnam:
* Use validatable input fields, and error displaying utilities
* Put UI in the OKCancelForm
* Significant cleanup throughout


MattPutnam:
In other news, marking a user as "NO ACCESS" doesn't remove them from the list, like it does for workspaces.  Seems to be an existing issue.

davidangb:
The no-access issue is DSDEEPB-1970.

dmohs:
:+1:

PR #355, opened 2016-01-11
DSDEEPB-2248: use version number for Thurloe profile instead of true/false

davidangb:
now, when we check the user's profile in Thurloe, we read a version number instead of a true/false indicator. This allows us to change the profile fields over time.

we rely on cljs to return 0 when casting "true" to int.


rushtong:
:+1: Looks good, merging.

PR #357, opened 2016-01-11
DSDEEPB-2266: Use /inputsOutputs endpoint and fix method config edit

MattPutnam:
* Using proper endpoint to get all inputs and outputs, not just ones that have values
* Don't save empty strings, instead remove the entry from the payload
* Display input/output type and if it's optional
* General cleanup of the editor code


dmohs:
:+1: Works great. Edited some methods with and without optional inputs.

PR #361, opened 2016-01-12
Fix Workspace ACL editing

dmohs:


src/cljs/org/broadinstitute/firecloud_ui/common/input.cljs
```diff
@@ -24,8 +24,11 @@
                                :style (merge (or (:style props) {})
                                         (when (:invalid @state)
                                           {:borderColor (:exception-red style/colors)}))
-                               :onChange #(swap! state dissoc :invalid)
+                               :onChange #(do (swap! state dissoc :invalid)
+                                              (when-let [x (:onChange props)]
+                                                (x (.. % -target -value))))
                                :defaultValue (:defaultValue props)
+                               :value (:value props)
```
davidangb: is there a point in setting both value and defaultValue?

src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
```diff
@@ -63,7 +63,9 @@
 
 (defn create-select [props options]
   [:select (deep-merge {:style select-style} props)
-   (map-indexed (fn [i opt] [:option {:value i} opt]) options)])
```
davidangb: I know this is my doing in the first place, but now this really needs to be combined with create-identity-select

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/acl_editor.cljs
```diff
@@ -52,22 +36,25 @@
...
+              :onChange #(swap! state assoc-in [:acl-vec i :email] %)}]
             (style/create-select
               {:ref (str "acl-value" i)
                :style {:float "right" :width column-width :height 33}
-               :defaultValue (:accessLevel acl-entry)}
-              access-levels)
+               :value (:accessLevel acl-entry)
+               :onChange #(swap! state assoc-in [:acl-vec i :accessLevel]
+                                 (.. % -target -value))}
+              (reduce (fn [r x] (assoc r x x)) {} access-levels))
```
davidangb: ah - this could just use create-identity-select instead of transforming access-levels into a x->x map. I'm fine with either approach - just an example of how create-select and create-identity-select should be combined.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/acl_editor.cljs
```diff
@@ -7,29 +7,13 @@
...
-(def ^:private READER 2)
-(def ^:private NO_ACCESS 3)
-(defn- access-level-to-index [string]
-  (case string
-    "OWNER" OWNER
-    "WRITER" WRITER
-    "READER" READER
-    "NO ACCESS" NO_ACCESS))
-(defn- index-to-access-level [index]
-  (nth access-levels index))
```
davidangb: nice cleaning this up!

davidangb:
See comments inline, especially about create-select vs. create-identity-select.

Additionally - and this appears to be a regression - the email address field no longer trims values. Previously, a value of "dmohs.dev@gmail.com  " (note the two trailing spaces) would work. With this PR, we send the trailing spaces through to rawls, resulting in an error: rawls can't find that email address.

other than that, looks good, and glad to see this cleaned up!

src/cljs/org/broadinstitute/firecloud_ui/common/input.cljs
```diff
@@ -24,8 +24,11 @@
                                :style (merge (or (:style props) {})
                                         (when (:invalid @state)
                                           {:borderColor (:exception-red style/colors)}))
-                               :onChange #(swap! state dissoc :invalid)
+                               :onChange #(do (swap! state dissoc :invalid)
+                                              (when-let [x (:onChange props)]
+                                                (x (.. % -target -value))))
```
rushtong: May I ask for a walkthrough on each element of this line? I understand that x is the onChange handler, but that's it.

src/cljs/org/broadinstitute/firecloud_ui/common/input.cljs
dmohs: It wouldn't make sense to set both, but it's okay here to just pass them along and let the underlying control handle that case.

src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
dmohs: I should have just used identity select, which is what I did.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/acl_editor.cljs
dmohs: Maybe it's fine to have both. In any case, my usage was clearly the identity one, so I just fixed it.

dmohs:
David, thanks for the trim catch. Fixed.

src/cljs/org/broadinstitute/firecloud_ui/common/input.cljs
dmohs: When an onChange handler has been defined in props (that is, the :onChange value in that map is not nil), call it with the value in the input field. onChange is passed the event, so (.. % -target -value) would be the same as JavaScript function(e) { return e.target.value; }

I maybe should have just passed the event along though to keep things consistent with normal input elements. That might be better.

davidangb:
In the future, when updating a PR, could you make changes in a separate commit? It makes it much, much easier to see what has changed. We can squash as necessary after the PR is finalized.

dmohs:
Will do.

PR #364, opened 2016-01-12
DSDEEPB-2381: fix status icons

MattPutnam:
* Use "executionStatus" (cromwell status) instead of "backendStatus" (JES status) for call-level detail
* Separate workflow vs. call statuses, centralize lists of values, and warn if match not found


davidangb:
:+1:

PR #367, opened 2016-01-13
Change the display of submission status counts

rushtong:
I changed the approach to show all of the submission statuses with their associated count values. This should account for any kind of status that might come down the pike and doesn't depend on a pre-determined list.


MattPutnam:
This is accumulating all of the workflow statuses within the submissions.  So if you had 2 submissions with 2 workflows each we'd say "2 submissions:" but then show a total of 4 items in the list.  Changing it to accumulate submission statuses makes this consistent, but is uninteresting (both succeeded and failed just report "Done").

@broadinstitute/silver what do we want to do here?

rushtong:
I see that detail now. I can re-look at this and try to narrow down the statuses up to the submission level, but that means there could be the case where in a submission, one workflow succeeds and one fails - that seems like a failure case in any event.

PR #369, opened 2016-01-13
UI work for adding new fields to the profile page

rushtong:
This requires orchestration work in this PR: https://github.com/broadinstitute/firecloud-orchestration/pull/190

I am open to feedback on the form design. It is currently very full. I'd like to spend more time redesigning it, but I wanted to get something out for review even if it can use some improvement.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -150,7 +150,12 @@
           (let [parsed-values (when success? (common/parse-profile (get-parsed-response)))]
             (cond
               (and success? (>= (int (:isRegistrationComplete parsed-values)) 1))
-              (swap! state assoc :registration-status :registered :name (str (:firstName parsed-values) (:lastName parsed-values)))
+              (swap! state assoc :registration-status :registered
+                :name (let [name (str (:firstName parsed-values) " " (:lastName parsed-values))]
+                        (if (-> name clojure.string/trim empty?)
+                          ; this is here primarily for old users without :firstName/:lastName fields
+                          "Profile"
+                          name)))
               success? ; partial profile case
               (swap! state assoc :registration-status :not-registered)
               :else
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -149,7 +149,7 @@
         (fn [{:keys [success? status-text get-parsed-response]}]
           (let [parsed-values (when success? (common/parse-profile (get-parsed-response)))]
             (cond
-              (and success? (>= (int (:isRegistrationComplete parsed-values)) 1))
+              (and success? (>= (int (:isRegistrationComplete parsed-values)) 2))
               (swap! state assoc :registration-status :registered
                 :name (let [name (str (:firstName parsed-values) " " (:lastName parsed-values))]
                         (if (-> name clojure.string/trim empty?)
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -150,7 +150,7 @@
           (let [parsed-values (when success? (common/parse-profile (get-parsed-response)))]
             (cond
               (and success? (>= (int (:isRegistrationComplete parsed-values)) 1))
-              (swap! state assoc :registration-status :registered :name (:name parsed-values))
+              (swap! state assoc :registration-status :registered :name (str (:firstName parsed-values) (:lastName parsed-values)))
               success? ; partial profile case
               (swap! state assoc :registration-status :not-registered)
               :else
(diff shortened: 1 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -29,15 +29,49 @@
...
+        (if (= (get-in @state [:values key]) value)
+          [:input {:type "radio"
+                   :value value
+                   :checked true
+                   :name key
+                   :onChange #(swap! state assoc-in [:values key] value)} value]
+          [:input {:type "radio"
+                   :value value
+                   :name key
+                   :onChange #(swap! state assoc-in [:values key] value)} value])])
```
rushtong: I would like to see a cleaner way of doing this - feedback appreciated.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
```diff
@@ -29,15 +29,49 @@
...
+            [:div {:style {:fontSize "88%"}} "Program Location:"
+              [:div {}
+                (react/call :render-nested-field this :programLocationCity "*City:")
+                (react/call :render-nested-field this :programLocationState "*State/Province:")
+                (react/call :render-nested-field this :programLocationCountry "*Country:")]]
+                (react/call :render-field this :pi "*Principal Investigator/Program Lead:")
+            [:div {:style {:marginBottom "1em" :fontSize "88%"}} "*NonProfit Status:"
+              (react/call :render-radio-field this :nonProfitStatus "Profit")
+              (react/call :render-radio-field this :nonProfitStatus "Non-Profit")]
+            (react/call :render-field this :billingAccountName "Google Billing Account Name")
```
rushtong: I know we don't like camel-casing, but if I used dashes, the json comes through un-camel-cased and broken. Would like a approach to bridge that gap.

rushtong:
There's one more big issue here - we need some error handling. That wasn't there before, and I didn't add it in this PR.

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
MattPutnam: [:input {:type "radio :value value :name key
             :checked (= (get-in @state [:values key]) value)
             :onChange #(swap! state assoc-in [:values key] value)} value]

src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
MattPutnam: Yeah, that's fine.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -150,7 +150,7 @@
           (let [parsed-values (when success? (common/parse-profile (get-parsed-response)))]
             (cond
               (and success? (>= (int (:isRegistrationComplete parsed-values)) 1))
-              (swap! state assoc :registration-status :registered :name (:name parsed-values))
+              (swap! state assoc :registration-status :registered :name (str (:firstName parsed-values) (:lastName parsed-values)))
```
MattPutnam: Space between them?

MattPutnam:
Small problem: When loading this branch, I initially couldn't see the profile link--because my profile didn't have a `:firstName` or `:lastName`, so the text for the link was empty.  Should we:

* Detect if both of those fields are missing, and see if the old value is there?
* Detect if both fields are missing, and display a placeholder instead?
* Not worry about it, since it only affects users from before the release?

davidangb:
Regarding the last comment, I think we should detect if both :firstName and :lastName are missing and display a placeholder instead.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -150,7 +150,12 @@
           (let [parsed-values (when success? (common/parse-profile (get-parsed-response)))]
             (cond
               (and success? (>= (int (:isRegistrationComplete parsed-values)) 1))
-              (swap! state assoc :registration-status :registered :name (:name parsed-values))
+              (swap! state assoc :registration-status :registered
+                :name (let [name (str (:firstName parsed-values) " " (:lastName parsed-values))]
+                        (if (-> name clojure.string/trim empty?)
+                          ; this is here primarily for old users without :firstName/:lastName fields
+                          "Profile"
+                          name)))
```
rushtong: Excellent - I was going to do something like this too.

rushtong:
@davidangb - @MattPutnam beat you to it - he applied that change already.

rushtong:
@MattPutnam - Can you look at line 167?
```(assoc new-state :server-error (get-parsed-response))```
This *should* be showing an error when one of the required fields is empty, but I can't see why it isn't happening.

davidangb:
This may be a larger change than we want to take on - but should this use the TextField component instead of style/create-text-field? TextField would give you validation controls.

rushtong:
I was not aware of TextField that we were migrating - I will look into it and see what I can sort out.

rushtong:
I'm closing this PR for now - there's still more work to do with error handling.

PR #370, opened 2016-01-13
Security fixes per DSDEEPB-2534

dmohs:


davidangb:
should this add the secure-cookie rewrite? It would never be exercised, but it might help us check off boxes during a security audit.

PR #371, opened 2016-01-13
DSDEEPB-2462: better error message for inactive users

davidangb:
Please see comments on the associated Jira ticket.


MattPutnam:
When I registered a new user, it still hung on the workspaces page, and only showed the error after refreshing the page.

davidangb:
When I register a new user, I do get the error on the workspaces page ... BUT the failed requests to rawls took 60-90 seconds. So I had to wait what felt like an eternity before the error message showed up.

MattPutnam:
Ah, I wasn't that patient.

:+1:

PR #374, opened 2016-01-14
Update NIH link expiration text

rushtong:


davidangb:
two changes:
1) Now that the warning text is significantly longer, if you narrow your browser window you'll see it wraps right-aligned. I think this looks funny - suggest adding either :text-align "center" (my preference) or :text-align "left" to the wrapping div.
2) The (.calendar expire-time) function adds the words "today" or "tomorrow", as appropriate, so we should not hardcode "today" into the message ... especially because our window of "within 24 hours" could include tomorrow. See screenshot. I tested this by manually updating my timestamp in Thurloe to see the effect on the banner.

davidangb:
See #375

davidangb:
:+1:!

PR #378, opened 2016-01-14
DSDEEPB-2579: further updates to NIH link warning banner verbiage

davidangb:
more verbiage updates, by request.

Note that I did not use the *exact* language from the Jira ticket - I changed "TCGA controlled workspaces" to "TCGA Controlled Access workspaces" to be consistent with the previous use of that phrase.


rushtong:
:+1: I haven't been able to finagle my settings to make it display, but the text looks good. Merging.

PR #381, opened 2016-01-21
use supported url for GCS downloads

davidangb:
a la https://cloud.google.com/storage/docs/authentication#cookieauth

This does not attempt to address DSDEEPB-2380. It's simply about putting us on a supported url format instead of an unsupported reverse-engineered url format.


dmohs:
:+1:

PR #388, opened 2016-01-29
Configuration updates

dmohs:
- simplifies docker image by using a standard image for compilation
- upgrades clojurescript, figwheel, and lein-cljsbuild dependencies
- fixes console warning from React
- simplifies project structure to use static index.html page


Dockerfile
```diff
@@ -46,7 +14,7 @@ EXPOSE 443
 ENV HTTPD_PORT=80 SSL_HTTPD_PORT=443
 ENV SERVER_ADMIN=devops@broadinstitute.org
 ENV LOG_LEVEL=warn
-ENV ORCH_URL_ROOT=http://orch:8080
+ENV ORCH_URL_ROOT=https://firecloud-orchestration.dsde-dev.broadinstitute.org
```
dmohs: Set to 'FIXME'

script/build-once.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -euox pipefail
+IFS=$'\n\t'
+
+./script/clean-build-dir.sh
+exec docker run --rm -it -w /w -v "$PWD":/w -v "$HOME"/.m2:/root/.m2 \
```
dmohs: /w -> /work

davidangb:
I was able to build the docker image, compile the js, and see the site run ... yay! But I can't get hot reloading working. Here's what I know ...
- I am running my docker machine on a non-default IP, due to conflicts with Broad's Cisco VPN routing. I suspect other developers will be doing the same.

```
wm45c-7da:~/work/src/firecloud-environment>docker-machine env firecloud
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://10.255.0.100:2376"
export DOCKER_CERT_PATH="/Users/davidan/.docker/machine/machines/firecloud"
export DOCKER_MACHINE_NAME="firecloud"
# Run this command to configure your shell: 
# eval "$(docker-machine env firecloud)"

wm45c-7da:~/work/src/firecloud-environment>echo $DOCKER_HOST
tcp://10.255.0.100:2376
```
- The hot reloader script does not pick up my DOCKER_HOST env var:

```
wm45c-7da:~/work/src/firecloud-ui>./script/start-hot-reloader.sh 
+ IFS=''
+ ./script/clean-build-dir.sh
+ mkdir -p target
+ find target -mindepth 1 -maxdepth 1 -and -not -name 'config*' -exec rm -rf '{}' ';'
+ exec docker run --rm -it -w /w -v /Users/davidan/work/src/firecloud-ui:/w -v /Users/davidan/.m2:/root/.m2 -p 3449:3449 dmohs/clojure rlfe lein do resource, figwheel
***
*** You did not specify a DOCKER_HOST environment variable.
*** Using the default: 192.168.99.100
***
Copy src/static/.DS_Store to /w/target/.DS_Store
Copy src/static/assets/broad_logo.png to /w/target/assets/broad_logo.png
```
- Even if the script did pick up my env var, I'm not convinced it would work correctly. My env var has a value of "tcp://10.255.0.100:2376" but project.clj seems to expect an IP address (or hostname) without a protocol or port.

dmohs:
@davidangb I'm so sorry. There were several problems with that and I clearly hadn't tested it. Try pulling my latest changes and doing this instead:

```
FIGWHEEL_HOST=10.255.0.100 ./script/start-hot-reloader.sh
```

davidangb:
:+1: looks good, works good. Back to you for two things:
1. please update the README to explain the usage of FIGWHEEL_HOST (personally, I'll end up putting this in my .bash_profile
2. figure out how to coordinate deployment of this PR with the required Jenkins changes

dmohs:
Is the warning not sufficient? I'd prefer that over a README change which may get overlooked as things evolve.

PR #389, opened 2016-02-02
DSDEEPB-2238: preserve whitespace in description paragraph

MattPutnam:


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -153,7 +153,7 @@
                                                               :defaultValue description
                                                               :style {:width 400}
                                                               :rows 5}))
-                description description
+                description [:div {:style {:whiteSpace "pre"}} description]
```
dmohs: `pre` means no line wraps. I think you want `pre-wrap`.

dmohs:
One change. Otherwise looks good. Merge when ready.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
rushtong: Back in the day, when I learned html in the early 90's, pre meant pre-formatted and preserved both white spaces and line wraps. I think that's still true today: http://www.w3schools.com/tags/tag_pre.asp

Edit: Looks like it's the css pre, which of course, has different semantics. Sorry.

dmohs: `:whiteSpace "pre"` has the same semantics as the `<pre>` tag. The problem with it is that it will also preserve the lack of line wraps, so if the user types an entire paragraph into the box, the whole thing will be forced into a single line.

`pre-wrap` preserves white space and line wraps while also adding soft wraps when the content reaches the edge of the container.

rushtong: Yes, thanks - for the briefest of moments last night, I thought my ancient knowledge was useful :-)

PR #391, opened 2016-02-02
Fix missing :sort-initial functionality

MattPutnam:
This got lost during some column state cleanup.  This fix is sort of ugly in that respect, @dmohs please see if this can be done better.


dmohs:
@MattPutnam I think this achieves the same result without adding `:sort-initial` and `:sort-by` to the column's state:
https://github.com/broadinstitute/firecloud-ui/pull/394

PR #394, opened 2016-02-05
Fix initial sorting in table component

dmohs:


MattPutnam:
:+1: Yep, this works.  Merging this one and deleting the other.

PR #395, opened 2016-02-05
Send requests straight through to orch instead of proxying them

dmohs:
Depends on https://github.com/broadinstitute/firecloud-orchestration/pull/197 and config changes in https://github.com/broadinstitute/firecloud-develop/pull/1.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -183,7 +183,7 @@
       [:div {}
        [comps/Button
         {:text "Sign In"
-         :href (str (get @config/config "apiUrlRoot") "/login?callback="
+         :href (str (config/api-url-root) "/login?callback="
                     (js/encodeURIComponent
                      (str js/window.location.protocol "//" js/window.location.hostname))
                     (let [hash (nav/get-hash-value)]
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -183,7 +183,7 @@
       [:div {}
        [comps/Button
         {:text "Sign In"
-         :href (str "/service/login?callback="
+         :href (str (get @config/config "apiUrlRoot") "/login?callback="
                     (js/encodeURIComponent
                      (str js/window.location.protocol "//" js/window.location.hostname))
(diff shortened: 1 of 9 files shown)


davidangb:
This mostly worked for me, good stuff.
- I manually updated my config.json to include "apiUrlRoot": "https://local.broadinstitute.org:20443" and I was able to talk to my local orchestration. Need to inform the dev team how to do this.
- When talking to https://firecloud-orchestration.dsde-dev.broadinstitute.org, the default, my ajax requests all failed on the CORS OPTIONS requests. Is this expected to work right now?

I did not do full-scale testing to ensure that all instances of /service/endpoint were replaced with /endpoint.

dmohs:
@davidangb I refactored this a bit and re-added the local option in the start-server script. When the new orch is deployed, the ajax requests will work.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -111,7 +112,6 @@
                                               :get-parsed-response #(parse-json-string (.-responseText xhr))}]
                            (if (= status-code 401)
                              (let [us-xhr (js/XMLHttpRequest.)]
-                               (set! (.-withCredentials us-xhr) true)
```
davidangb, line 114: isn't this necessary to send the Authorization header?

ninja edit: I guess not. I can see the Authorization header being sent.

src/docker/run-apache.sh
```diff
@@ -2,9 +2,18 @@
 set -euo pipefail
 IFS=$'\n\t'
 
-cp /config/config.json /app/target/config.json
 
-SITE_CONF=$(cat << EOF
+CONFIG_JSON=$(cat <<EOF
```
davidangb: why do it this way? This seems to go against the model we're trying to set, where config files are stored in firecloud-develop. (or firecloud)

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -182,7 +182,7 @@
       [:div {}
        [comps/Button
         {:text "Sign In"
-         :href (str "/service/login?callback="
+         :href (str (get @config/config "apiUrlRoot") "/login?callback="
```
davidangb: I think this fixes DSDEEPB-2007, too.

PR #396, opened 2016-02-08
Replace WDL display textarea with CodeMirror component

MattPutnam:
The WDL parsing isn't perfect--it does some weird things on embedded commands--but it's at least as good as what we had before.


davidangb:
This looks so good. I love it.

(I had weird problems in my dev env getting the changes to index.html to take effect. Hopefully this works for others.)

:+1: merging.

PR #398, opened 2016-02-10
fixed layout of paginator

MattPutnam:
was screwy and then broke when I changed the combo box height


dmohs:
:+1: Looks good. Cleaner.

PR #399, opened 2016-02-11
GAWB-3: Just display the stupid message.

MattPutnam:
This may cause cosmetic problems if the status text of other errors is missing or insufficient, but we are considering that to be a separate issue.


helgridly:
I am Obama, and I approve this PR. Hit it.

PR #400, opened 2016-02-16
Move firecloud-ui to new base image

coreone:
- Rebuild container using apache-proxy image instead of openidc-proxy image
- Include stackdriver status module
- Include all FISMA security features


coreone:
@dmohs FYI: apache-proxy:latest is now up to date, so this can be merged whenever you wish.

dmohs:
:+1: Tested.

PR #401, opened 2016-02-16
Two quick fixes for stupid bugs I introduced

MattPutnam:
- Now properly capping initial workspace description column width in main workspace list
- Undid change to text input styling that broke the text areas for workspace description


dmohs:
:+1:

PR #410, opened 2016-03-01
Untangle login flow

dmohs:
Fixes GAWB-330. More straightforward handling of authentication flow. If you want to test this out without checking out code, add this to your /etc/hosts:

```
104.197.34.9 local.broadinstitute.org
```

and use this account:

```
vault read secret/dsde/dev/common/firecloud-test-user
```

Once the user is registered, you can enable them like so:

```
curl -i -d '' -H 'Authorization: Bearer '"$DEV_ACCOUNT_AUTH_TOKEN" 'https://local.broadinstitute.org:20443/api/user/112805780302623794408/enable'
```

If you want to start over (as if the account was never registered), let me know.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
@@ -104,37 +104,13 @@
                     (aset xhr (name k) v))
                   xhr))
           call-on-done (fn []
-                         (let [status-code (.-status xhr)
-                               orig-response {:xhr xhr
-                                              :status-code status-code
-                                              :success? (< 199 status-code 300)
-                                              :status-text (.-statusText xhr)
-                                              :get-parsed-response #(parse-json-string (.-responseText xhr))}]
-                           (if (= status-code 401)
-                             (let [us-xhr (js/XMLHttpRequest.)]
-                               (.addEventListener us-xhr "loadend"
-                                 (fn []
-                                   (let [us-status (.-status us-xhr)
-                                         parsed-us-response (parse-json-string (.-responseText us-xhr))
-                                         keywordized-us-response (keywordize-keys parsed-us-response)]
-                                     (cond
-                                       ;TODO: Fix this with a real log-out once the login bug is fixed and logout is implemented.
-                                       (= us-status 401) (do (delete-access-token-cookie)
-                                                           (.reload (.-location js/window)))
-                                       (= us-status 403)
-                                       (on-done {:xhr us-xhr
-                                                 :status-code us-status
-                                                 :success? false
-                                                 :status-text "You are successfully registered, but your account is currently inactive. You will be contacted via email when your account is activated."
-                                                 :get-parsed-response parsed-us-response})
-                                       (= us-status 404) (.replace (.-location js/window) "#profile")
-                                       :else (on-done orig-response)))))
-                               (.open us-xhr "GET" (str (config/api-url-root) "/me"))
-                               (.setRequestHeader us-xhr "Authorization" (str "Bearer " @access-token))
-                               (.setRequestHeader us-xhr "Content-Type" "application/json")
-                               (.setRequestHeader us-xhr "Accept" "application/json")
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -104,37 +104,13 @@
...
-                           (if (= status-code 401)
-                             (let [us-xhr (js/XMLHttpRequest.)]
-                               (.addEventListener us-xhr "loadend"
-                                 (fn []
-                                   (let [us-status (.-status us-xhr)
-                                         parsed-us-response (parse-json-string (.-responseText us-xhr))
-                                         keywordized-us-response (keywordize-keys parsed-us-response)]
-                                     (cond
-                                       ;TODO: Fix this with a real log-out once the login bug is fixed and logout is implemented.
-                                       (= us-status 401) (do (delete-access-token-cookie)
```
davidangb, line 122: does this deletion mean we realistically never call delete-access-token-cookie any more? After this PR, it looks like the only place we call it any more is in the else clause of set-access-token-cookie, which is an edge case.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -156,12 +132,19 @@
...
-         :headers (merge {"Authorization" (str "Bearer " @access-token)}
-                         (:headers arg-map)))))
+  (let [on-done (:on-done arg-map)]
+    (ajax (assoc
+           arg-map :url (str (config/api-url-root) service-prefix path)
+           :headers (merge {"Authorization" (str "Bearer " @access-token)}
+                           (:headers arg-map))
+           :on-done (fn [{:keys [status-code] :as m}]
+                      ;; Handle auth token expiration
+                      (if (= status-code 401)
```
davidangb, line 145: this moves the check for 401s from ajax to ajax-orch. There are still two places that call ajax directly (for /config.json and /me), and they look safe. But maybe add a comment to the defn ajax to ensure nobody else unintentionally calls it directly?

davidangb:
Two minor notes inline.

I like this a lot better - it's much cleaner. It's working for me in all the cases I tested, though I confess I did not test every single user path.

There is a minor behavior change in a corner case. Consider a user who logs in successfully and starts browsing the site. During the user session, a FireCloud admin deactivates the user (removes the user from LDAP). Previously, the user would immediately get a human-readable message upon the next ajax request: "You are successfully registered, but your account is currently inactive. You will be contacted via email when your account is activated." As of this PR, we would simply display the HTTP/proxy default Forbidden message. Personally, I think this is fine, and it should be rare, but I want to make sure PO signs off on it.

Otherwise, :+1:

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
dmohs, line 122: I like the idea of keeping things clean. Added a delete on 401.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
dmohs, line 145: I don't like commenting on when a function should be used (I can't know all potential use cases), so I just updated the two non-orch call sites so if a developer is looking at any ajax call in our code, they either see `ajax-orch` or `ajax` with an explanation of why it is an exceptional case.

dmohs:
If they get a 401, the page is reloaded, which will walk them through authentication from the beginning. So, if they are deactivated, the page will reload and they will see the "awaiting activation" message.

davidangb:
Looks fine to me, go ahead and rebase/merge.

We disagree philosophically regarding comments on a function - I think it's perfectly fine to add comments that say "this function does X; if you want Y behavior please use this other function". But I sure ain't gonna argue this point and hold up the PR!

:+1:

PR #426, opened 2016-03-28
Display submission statuses directly

jmthibault79:
GAWB-251
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
@@ -173,23 +173,13 @@
       (style/create-section-header "Analysis Submissions")
       (style/create-paragraph
         (let [count-all (count submissions)
-              success-count (->> submissions (filter all-success?) count)
-              running-count (->> submissions (filter any-running?) count)
-              fail-count (->> submissions (filter any-failed?) count)
-              unknown-count (- count-all success-count running-count fail-count)]
+              grouped-by-status (group-by #(% "status") submissions)]
              [:div {}
               (str count-all " Submission" (when-not (= 1 count-all) "s"))
-              (when (or (pos? fail-count) (pos? running-count) (pos? success-count) (pos? unknown-count))
+              (when (pos? count-all)
                     [:ul {:style {:marginTop "0"}}
-                     (when (pos? success-count)
-                           [:li {} (str success-count " succeeded")])
-                     (when (pos? fail-count)
-                           [:li {} (str fail-count " failed")])
-                     (when (pos? running-count)
-                           [:li {} (str running-count " running")])
-                     (when (pos? unknown-count)
-                           [:li {} (str unknown-count " unknown")])])]))]]))
-
+                     (for [[status subs] grouped-by-status]
+                        [:li {} (str (count subs) " " status)])])]))]]))
 
 (react/defc Summary
   {:refresh


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -173,23 +173,13 @@
...
-                           [:li {} (str success-count " succeeded")])
-                     (when (pos? fail-count)
-                           [:li {} (str fail-count " failed")])
-                     (when (pos? running-count)
-                           [:li {} (str running-count " running")])
-                     (when (pos? unknown-count)
-                           [:li {} (str unknown-count " unknown")])])]))]]))
-
+                    (for [[status subs] grouped-by-status]
+                        [:li {} (str (count subs) " " status)])])]))]]))
```
jmthibault79, line 182: This does not include the ! icon when it's Done with errors.  Do we want that level of detail here?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -173,23 +173,13 @@
...
         (let [count-all (count submissions)
-              success-count (->> submissions (filter all-success?) count)
-              running-count (->> submissions (filter any-running?) count)
-              fail-count (->> submissions (filter any-failed?) count)
-              unknown-count (- count-all success-count running-count fail-count)]
+              grouped-by-status (group-by #(% "status") submissions)]
              [:div {}
               (str count-all " Submission" (when-not (= 1 count-all) "s"))
-              (when (or (pos? fail-count) (pos? running-count) (pos? success-count) (pos? unknown-count))
+              (when (pos? count-all)
```
jmthibault79, line 179: This logic is equivalent, unless I missed something

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -173,23 +173,13 @@
       (style/create-section-header "Analysis Submissions")
       (style/create-paragraph
         (let [count-all (count submissions)
-              success-count (->> submissions (filter all-success?) count)
```
jmthibault79, line 176: I was unable to find the original motivation for this logic.  I saw @rushtong and @MattPutnam in the change history here, so I would appreciate any recollection you have of this.

jmthibault79:
FYI @davidangb @dmohs

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
rushtong, line 176: I think the idea here is to find the submissions that have all of their workflows in a successful state - according to a definition of successful (see https://github.com/broadinstitute/firecloud-ui/blob/develop/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/monitor/common.cljs#L24). Since submission and workflow states can be different, we needed a way to bucket the submission based on _all_ of the workflows inside it.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -173,23 +173,13 @@
...
-              running-count (->> submissions (filter any-running?) count)
-              fail-count (->> submissions (filter any-failed?) count)
-              unknown-count (- count-all success-count running-count fail-count)]
+              grouped-by-status (group-by #(% "status") submissions)]
```
jmthibault79, line 176: I may be using `#(% "status")` incorrectly because I don't really understand the syntax, but it seems to work as I intended.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
jmthibault79, line 176: OK.  That is useful information, and can be seen in the Workflows section of the Monitor tab.  But based on this story and a conversation with David Siedzik I don't think it belongs _here_.  Here we want to match the Submissions section of the Monitor tab.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
MattPutnam, line 176: This is correct.  The function, fully written out, would be:

```
(fn [submission] (get submission "status"))
```

Data structures can act as their own getter function, so the `get` is unnecessary:

```
(fn [submission] (submission "status"))
```

Then we use the function macro since it's a function of one variable that's only referenced once:

```
#(% "status")
```

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
MattPutnam, line 182: Sounds like that's been confusing and there have been requests to get rid of that icon anyway.  Let's leave this as-is.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
MattPutnam, line 179: Seems like it, especially how you've redone things.

MattPutnam:
:+1: Looks great.  One minor thing is indentation--it seems that you and Greg have it set up to work differently than me or dmohs.  There are already mixed styles so I'm not going to be a stickler, but kick line 181 over one space so the <ul>'s children are indented.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
jmthibault79, line 176: Cool.  Thanks for the explanation!

davidangb:
The only functional change I see here is that the UI now reports the exact status string as contained in the submission (by design), instead of abstracting that string. For instance, the previous code could easily have displayed "success, yay!" instead of what's in the submission. I'm fine with this, until/if PO decides we need friendlier submission statuses.

PR #444, opened 2016-05-24
Add user's position in workflow queue to status

jmthibault79:
GAWB-715

Depends on https://github.com/broadinstitute/rawls/pull/436
- [x] **Submitter**: Rebase to develop. DO NOT SQUASH
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Make sure documentation for code is complete
- [x] **Submitter**: Review code comments; remove done TODOs, create stories for remaining TODOs
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Add description or comments on the PR explaining the hows/whys (if not obvious)
- [x] **Submitter**: Tell tech lead that the PR exists if s/he wants to look at it
- [x] **Submitter**: Anoint a lead reviewer (LR). **Assign PR to LR**
- [x] **LR**: Initial review by LR and others.
- [x] Comment / review / update cycle:
  - Rest of team may comments on PR at will
  - **LR assigns to submitter** for feedback fixes
  - Submitter updates documentation as needed
  - Submitter rebases to develop again if necessary
  - Submitter makes further commits. DO NOT SQUASH. **Reassign to LR** for further feedback
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
- [x] **Assign to submitter** to finalize
- [x] **Submitter**: Squash commits, rebase if necessary
- [x] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Merge to develop 
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: Check configuration files in Jenkins in case they need changes
- [x] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't 
  working for ANY reason!
- [x] **Submitter**: Verify swagger UI on dev server still works after deployment
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email
- [x] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

PR #440, opened 2016-05-17
Display queue submission status in the Launch Analysis pane

jmthibault79:
GAWB-662

Thanks to @MattPutnam for horsefishing!
- [x] **Submitter**: Rebase to develop. DO NOT SQUASH
- [x] **Submitter**: Make sure Swagger is updated if API changes
- [x] **Submitter**: Make sure documentation for code is complete
- [x] **Submitter**: Review code comments; remove done TODOs, create stories for remaining TODOs
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Add description or comments on the PR explaining the hows/whys (if not obvious)
- [x] **Submitter**: Tell tech lead that the PR exists if s/he wants to look at it
- [x] **Submitter**: Anoint a lead reviewer (LR). **Assign PR to LR**
- [x] **LR**: Initial review by LR and others.
- [x] Comment / review / update cycle:
  - Rest of team may comments on PR at will
  - **LR assigns to submitter** for feedback fixes
  - Submitter updates documentation as needed
  - Submitter rebases to develop again if necessary
  - Submitter makes further commits. DO NOT SQUASH. **Reassign to LR** for further feedback
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
- [x] **Assign to submitter** to finalize
- [x] **Submitter**: Squash commits, rebase if necessary
- [x] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Merge to develop 
- [x] **Submitter**: Delete branch after merge
- [x] **Submitter**: Check configuration files in Jenkins in case they need changes
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't 
  working for ANY reason!
- [x] **Submitter**: Verify swagger UI on dev server still works after deployment
- [x] **Submitter**: Inform other teams of any API changes via hipchat and/or email
- [ ] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

PR #445, opened 2016-05-24
Add a global status line for submissions in queue and active.

rushtong:
See https://broadinstitute.atlassian.net/browse/GAWB-666

There are companion PRs for the local/dev configs here (and have been merged):
https://github.com/broadinstitute/firecloud-develop/pull/78
https://github.com/broadinstitute/firecloud-develop/pull/79
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -141,6 +141,34 @@
         [nih-link-warning/NihLinkWarning])])})
 
 
+(react/defc GlobalSubmissionStatus
+  {:render
+   (fn [{:keys [state]}]
+     (let [{:keys [status-error status-counts]} @state
+           {:keys [queued active queue-position]} status-counts]
+       [:div {} (str "Workflows: "
+                     (cond status-error status-error
+                           status-counts (str queued " Queued; " active " Active; " queue-position " ahead of yours")
+                           :else "loading..."))]))
+   :component-did-mount
+   (fn [{:keys [this locals]}]
+     ;; Call once for initial load
+     (react/call :load-data this)
+     ;; Add a long-polling call for continuous updates
+     (swap! locals assoc :interval-id (js/setInterval #(react/call :load-data this) (config/submission-status-refresh))))
+   :component-will-unmount
+   (fn [{:keys [locals]}]
+     (js/clearInterval (:interval-id @locals)))
+   :load-data
+   (fn [{:keys [state]}]
+     (endpoints/call-ajax-orch
+       {:endpoint (endpoints/submissions-queue-status)
+        :on-done (fn [{:keys [success? status-text get-parsed-response]}]
+                   (if success?
+                     (swap! state assoc :status-error nil :status-counts (common/queue-status-counts (get-parsed-response)))
+                     (swap! state assoc :status-error status-text :status-counts nil)))}))})
+
+
 (react/defc LoggedIn
   {:render
    (fn [{:keys [this props state]}]
@@ -150,9 +178,11 @@
                      (= page :status))
          (nav/navigate (:nav-context props) "workspaces"))
        [:div {}
-        [:div {:style {:float "right" :fontSize "70%"}}
-         [:a {:style {:marginRight "1ex" :color (:link-blue style/colors)}
-              :href "#profile" } (:name @state)]]
+        [:div {:style {:float "right" :fontSize "70%" :textAlign "right" :marginRight "1ex"}}
+         [:a {:style {:color (:link-blue style/colors)}
+              :href "#profile" }
+          (:name @state)]
+         [GlobalSubmissionStatus]]
         (text-logo)
         (case (:registration-status @state)
           nil [:div {:style {:margin "2em 0" :textAlign "center"}}
@@ -208,7 +238,7 @@
 
 (react/defc LoggedOut
   {:render
-   (fn [{:keys [props]}]
+   (fn [{:keys []}]
      [:div {}
       [:div {:style {:marginBottom "2em"}} (text-logo)]
       [:div {}
@@ -278,7 +308,7 @@
           [LoggedOut])]]
       (footer)])
    :component-did-mount
-   (fn [{:keys [this state locals]}]
+   (fn [{:keys [this]}]
      (react/call :load-config this))
    :component-will-unmount
    (fn [{:keys [locals]}]
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -129,6 +129,13 @@
     (let [[ns n] (clojure.string/split segment #":")]
       {:namespace ns :name n})))
 
+;; GAWB-666 Globally show Queued and Cromwell (active) counts
+(defn queue-status-counts [{:strs [workflowCountsByStatus estimatedQueueTimeMS workflowsBeforeNextUserWorkflow]}]
+  {:queue-time (or estimatedQueueTimeMS 0)
+   :queue-position (or workflowsBeforeNextUserWorkflow 0)
+   :queued (apply + (map workflowCountsByStatus ["Queued" "Launching"]))
+   :active (apply + (map workflowCountsByStatus ["Submitted" "Running" "Aborting"]))})
+
 (def root-entity-types
   ["participant" "sample" "pair" "participant_set" "sample_set" "pair_set"])
 
--- a/src/cljs/org/broadinstitute/firecloud_ui/config.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/config.cljs
@@ -8,3 +8,4 @@
 (defn debug? [] (get @config "isDebug"))
 (defn tcga-namespace [] (get @config "tcgaNamespace"))
 (defn workflow-count-warning-threshold [] (get @config "workflowCountWarningThreshold" 100))
+(defn submission-status-refresh [] (get @config "submissionStatusRefresh" 60000)) ;; milliseconds
--- a/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
@@ -639,7 +639,12 @@
    :method :get
    :mock-data {"estimatedQueueTimeMS" (rand-int 10000000)
                "workflowCountsByStatus"
-               {"Launching" (rand-int 10000)
-                "Queued" (rand-int 10000)
-                "Some Other Status" (rand-int 5000)
-                "Another Generic Status" (rand-int 5000)}}})
+               {"Queued" (rand-int 10000)
+                "Launching" (rand-int 10000)
+                "Submitted" (rand-int 10000)
+                "Running" (rand-int 1000)
+                "Aborting" (rand-int 1000)
+                "Failed" (rand-int 1000)
+                "Succeeded" (rand-int 1000)
+                "Aborted" (rand-int 1000)
+                "Unknown" (rand-int 1000)}}})
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
@@ -18,34 +18,24 @@
 (defn- entity->id [entity]
   {:type (entity "entityType") :name (entity "name")})
 
-;; GAWB-662 / GAWB-715
-;; show Queued first, ignore Launching, and collapse all remaining states into Active
-(defn queue-status-remap [queue-status]
-  (let [status-count-map (queue-status "workflowCountsByStatus")]
-    (assoc queue-status "workflowCountsByStatus"
-      [["Queued" (get status-count-map "Queued" 0)]
-       ["Active" (apply + (vals (dissoc status-count-map "Queued" "Launching")))]])))
 
-(defn queue-status-table-row [[status count]]
+(defn- row [label content]
   [:div {}
-   [:div {:style {:width 100 :float "right"}} status]
-   [:div {:style {:marginRight 10 :float "right"}} count]
-   (common/clear-both)])
+   [:div {:style {:display "inline-block" :width 200 :textAlign "right" :marginRight "1ex"}} label]
+   [:div {:style {:display "inline-block" :width 240}} content]])
 
 (defn queue-status-table [state]
-  (let [error-msg (:submission-queue-error-message @state)
-        queue-status (:submission-queue-status @state)
-        {:strs [estimatedQueueTimeMS workflowsBeforeNextUserWorkflow workflowCountsByStatus]} queue-status]
+  (let [{:keys [queue-status queue-error]} @state
+        {:keys [queue-time queue-position queued active]} queue-status]
     [:div {:style {:marginBottom "1em" :float "right"}}
      (cond
-       error-msg (style/create-server-error-message error-msg)
+       queue-error (style/create-server-error-message queue-error)
        (not queue-status) [comps/Spinner {:text "Loading submission queue status..."}]
        :else
        [:div {}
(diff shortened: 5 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,39 @@
       (when (:show-nih-link-warning? props)
         [nih-link-warning/NihLinkWarning])])})
 
+;; GAWB-666 Globally show Queued and Cromwell (active) counts
+(defn queue-status-counts [workflow-counts]
+  (let [count-map (get workflow-counts "workflowCountsByStatus" (empty {}))]
+    {:queued (+ (get count-map "Queued" 0) (get count-map "Launching" 0))
+     :active (+ (get count-map "Submitted" 0) (get count-map "Running" 0) (get count-map "Aborting" 0))
+     :terminal (+ (get count-map "Failed" 0) (get count-map "Succeeded" 0) (get count-map "Aborted" 0))}))
```
rushtong: Terminal isn't currently in use in the UI, but is a category of statuses defined in Rawls. See https://github.com/broadinstitute/rawls/blob/e6be332405bfa36c35f935d7ea613e46a11feca8/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala#L294

jmthibault79:
Can we reuse the layout from #440 ?  I'm updating that code with #444 so I would prefer to do it in only one place.

rushtong:
To be honest, I'm not a fan of trying to squeeze different displays into the same layout just because we're using similar data. The two areas have different intentions (from what I understand from the tickets) and merit their own view. I am not opposed to refactoring the top nav to be more accommodating to a larger display, but I feel that is beyond the scope of this work and should be addressed as a separate story with input from Alex, et al.

One thing I think we could do is to add different render handlers to the react component so that both are in the same location. Depending on how the component is called, it can choose which layout to use.

jmthibault79:
My interpretation of this ticket was to make a copy of the other layout and display it elsewhere.  Sounds like that is incorrect - so do what makes sense in context!

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,40 @@
       (when (:show-nih-link-warning? props)
         [nih-link-warning/NihLinkWarning])])})
 
+;; GAWB-666 Globally show Queued and Cromwell (active) counts
+(defn queue-status-counts [workflow-counts]
+  (let [count-map (get workflow-counts "workflowCountsByStatus" (empty {}))
```
MattPutnam: `(empty {})` -> `{}`

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,40 @@
...
+        queue-time (get workflow-counts "estimatedQueueTimeMS" 0)]
+    {:queue-time queue-time
+     :queued (+ (get count-map "Queued" 0) (get count-map "Launching" 0))
```
MattPutnam: Since you're adding, you can get rid of the default and just use `(count-map "Queued")` etc.  If the value doesn't exist, it defaults to `nil`, which will coerce to 0 on adding.  You could then reduce to:

```
(apply + (map count-map ["Queued" "Launching"]))
```

rushtong: Sweet - that saves a lot of boilerplate.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,40 @@
...
+   (fn [{:keys [locals]}]
+     (js/clearInterval (:interval-id @locals)))
+   :load-data
+   (fn [{:keys [state]}]
+     (endpoints/call-ajax-orch
+       {:endpoint (endpoints/submissions-queue-status)
+        :on-done (fn [{:keys [success? status-text get-parsed-response]}]
+          (if success?
+            (swap! state assoc :status-counts (queue-status-counts (get-parsed-response)))
+            ()))}))})
```
MattPutnam: Do we want to display an error on failure?

rushtong: I struggled with that decision. There isn't really space for a full error, and I'm not sure what to display. Suggestions welcome!

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,40 @@
...
+     :active (+ (get count-map "Submitted" 0) (get count-map "Running" 0) (get count-map "Aborting" 0))
+     :terminal (+ (get count-map "Failed" 0) (get count-map "Succeeded" 0) (get count-map "Aborted" 0))}))
+
+(react/defc GlobalSubmissionStatus
+  {:render
+   (fn [{:keys [state]}]
+     (when-let [status-counts (:status-counts @state)]
+       [:span {:style {:marginRight "1ex"}}
+         [:br {}],
+         "Submissions: ", (:queued status-counts), " Queued; ", (:active status-counts), " Active"]))
```
MattPutnam: This seems like strangely structured HTML.  I always get suspicious of <br> tags.  Make this span a div, and make the <a> (for the profile) not float.  To get the right justification, put `:textAlign "right"` on the div containing the <a> and [GlobalSubmissionStatus].  You can also move the `:marginRight "1ex"` to that div and get rid of it from the two children.

Also the commas are unnecessary.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
MattPutnam: Maybe just

Submissions: Error--see console

rushtong: 👍

davidangb: what about the error case? If the queueStatus endpoint fails, this keeps displaying the last-known counts. Maybe it should display "Submissions: ?" or "Submissions: 504 Gateway Timeout" or the whole thing should disappear, or ... something?

rushtong: See Matt's feedback - will implement something along those lines.

davidangb: hah, sorry about the cross-commenting! I really don't like pointing people at the console, though.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
davidangb: agreed - was gonna comment about the same HTML, saw I cross-commented about the error case, came here to look for this ... +1 this comment :)

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: I think I like `Submissions: 504 Gateway Timeout` best, or some variant of that. It's just very un-informative and not terribly user-friendly.

davidangb:
Looks good to me w/ minor comments inline about HTML structure and ajax error cases.

Here's a thought: if this is going to ping the FireCloud back end (orch/rawls) every minute, it's possible that this endpoint will be the first to notice that the user's auth token has expired. I can't think right now what to DO with this information, but let's remember it.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: Yeah, I'm looking at re-formatting it.

jmthibault79:
BTW #444 has been merged

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,42 @@
...
+     :terminal (apply + (map count-map ["Failed" "Succeeded" "Aborted"]))}))
+
+(react/defc GlobalSubmissionStatus
+  {:render
+   (fn [{:keys [state]}]
+     (let [status-error (:status-error @state)
+           status-counts (:status-counts @state)]
+       (cond
+         status-error [:span {} [:br {}] "Submissions: " status-error]
+         status-counts [:span {} [:br {}] "Submissions: " (:queued status-counts) " Queued; " (:active status-counts) " Active"]
```
rushtong: @davidangb, @MattPutnam - So, without the `br` tag, the status text and the user name butt up against each other in an ugly way, which is why I had that in there in the first place. This version is slightly cleaner, html-wise.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,42 @@
...
+     :active (apply + (map count-map ["Submitted" "Running" "Aborting"]))
+     :terminal (apply + (map count-map ["Failed" "Succeeded" "Aborted"]))}))
+
+(react/defc GlobalSubmissionStatus
+  {:render
+   (fn [{:keys [state]}]
+     (let [status-error (:status-error @state)
+           status-counts (:status-counts @state)]
+       (cond
+         status-error [:span {} [:br {}] "Submissions: " status-error]
```
rushtong: You can test the error case by changing the queue endpoint url to something non-existent and see that show up in the Submissions line.

rushtong:
@davidangb 

> Here's a thought: if this is going to ping the FireCloud back end (orch/rawls) every minute, it's possible that this endpoint will be the first to notice that the user's auth token has expired. I can't think right now what to DO with this information, but let's remember it.

This is a great point. I seem to remember a story about expiration of auth tokens and having a user-friendly way to re-login, but I can't find it in the backlog at the moment.

jmthibault79:
The token expiration / login story is https://broadinstitute.atlassian.net/browse/GAWB-463

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,42 @@
...
+   :component-will-unmount
+   (fn [{:keys [locals]}]
+     (js/clearInterval (:interval-id @locals)))
+   :load-data
+   (fn [{:keys [state]}]
+     (endpoints/call-ajax-orch
+       {:endpoint (endpoints/submissions-queue-status)
+        :on-done (fn [{:keys [success? status-text get-parsed-response]}]
+                   (if success?
+                     (swap! state assoc :status-counts (queue-status-counts (get-parsed-response)))
```
davidangb: in the success case, I think you also need to clear the previous error from the state. Here's my use case:
1) Load a UI page, get successful queue status information, notice that it reloads every 60 secs, happy
2) Suddenly rawls has a transient error, and one of the reloads gets an exception. We display the exception.
3) As soon as rawls returns, we STILL display the exception - because the exception is still in the state.

I tested this by running a local stack, loading a UI page, and manually changing the rawls queueStatus endpoint behind the scenes to throw an error, then re-edited it to be successful.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,42 @@
       (when (:show-nih-link-warning? props)
         [nih-link-warning/NihLinkWarning])])})
 
+;; GAWB-666 Globally show Queued and Cromwell (active) counts
+(defn queue-status-counts [workflow-counts]
+  (let [count-map (get workflow-counts "workflowCountsByStatus" {})
+        queue-time (get workflow-counts "estimatedQueueTimeMS" 0)]
+    {:queue-time queue-time
+     :queued (apply + (map count-map ["Queued" "Launching"]))
```
jmthibault79: These don't match the counts from `launch_analysis.cljs`, and there's a discrepancy from what Rawls returns as well.

The Rawls endpoint returns only its `queuedStatuses` and `runningStatuses` so `:terminal` here will always be zero.  This is misleading.

`launch_analysis.cljs` drops the `Launching` status instead of adding it to `Queued`.  I think choosing either way is acceptable, but we should be consistent by reusing `queue-status-remap` or similar.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: Will implement that piece - great catch.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: The counts I'm showing here are the same as those defined in Rawls (see https://github.com/broadinstitute/rawls/blob/e6be332405bfa36c35f935d7ea613e46a11feca8/src/main/scala/org/broadinstitute/dsde/rawls/model/ExecutionModel.scala#L294) I'm taking the Rawls version as the correct counts, but that may be incorrect. I ended up not using `queue-status-remap` because it seems to be doing things differently than what's shown in `ExecutionModel`

rushtong: `:terminal` is not actually in use - I have it here purely for completeness.

jmthibault79: What I mean is that Rawls will never populate `workflowCountsByStatus` with the terminal statuses - so if we do end up using `:terminal` someday, it will be incorrect.  So I would prefer not to have it exist at all.

I agree with what you're saying about consistency - but since the other UI story intentionally differs from Rawls' internal DB states, we can't be consistent here with both.  My opinion is that we should be consistent with the other UI.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
MattPutnam: Looks like now you can change this span into a div and get rid of the br.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -140,6 +140,42 @@
...
+     :active (apply + (map count-map ["Submitted" "Running" "Aborting"]))
+     :terminal (apply + (map count-map ["Failed" "Succeeded" "Aborted"]))}))
+
+(react/defc GlobalSubmissionStatus
+  {:render
+   (fn [{:keys [state]}]
+     (let [status-error (:status-error @state)
+           status-counts (:status-counts @state)]
```
MattPutnam: Minor thing: this let could now be:

```
(let [{:keys [status-error status-counts]} @state]
  ... )
```

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong: I originally had this as a separate div, but that generates a full paragraph break which is much more than what I want and interferes with the methods and workspace tabs.

davidangb:
btw, here's what happens with the 60-second refresh and token expiration ... this ajax call acts as an auto-logout. If you have a page open, you're either working or just staring at the page ... this ajax call will keep happening. As soon as it gets a 401, it will redirect to the login screen. I wonder if we want this.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
davidangb: After looking at what's rendered in Chrome's DOM inspector, I'm fine with keeping this as a span; it's not a big deal.

However, I suggest a minor change from 

```
[:span {} [:br {}] "Submissions: " (:queued status-counts) " Queued; " (:active status-counts) " Active"]
```

to

```
[:span {} [:br {}] (str "Submissions: " (:queued status-counts) " Queued; " (:active status-counts) " Active")]
```

to keep the DOM as light as possible. The former produces:

```
<span data-reactid=".2.0.0.0.0.1.1">Submissions: </span>
<span data-reactid=".2.0.0.0.0.1.2">0</span>
<span data-reactid=".2.0.0.0.0.1.3"> Queued; </span>
<span data-reactid=".2.0.0.0.0.1.4">2</span>
<span data-reactid=".2.0.0.0.0.1.5"> Active</span>
```

while the latter produces:

```
<span data-reactid=".1.0.0.0.0.1.1">Submissions: 0 Queued; 2 Active</span>
```

davidangb:
👍 squash and go!

PR #452, opened 2016-06-08
GAWB-727: Complete Server-Side Pagination in the UI

MattPutnam:
This has a bunch of stuff in it.
1. SSP has a slightly different row sorting model than we had been using in the UI.  SSP requires a sort column, CSP does not.  So an option was added to the table to force it to always be sorting.
2. The set type's count column did not play nice with SSP, since it's derived client side.  To fix this, Rawls made list types sort by length, and the count column was removed.  The corresponding set membership column (e.g., "pairs" for a pair set table) now displays the count first in the cell.  This needs PO approval.
3. The /entities endpoint needed to be augmented to return the unioned set of attribute names for each type, so the entity table now uses that instead of calculating it client-side

A few issues remain, but I'd like to push this through to un-break the UI in dev.
1. Importing behavior needs to be checked.  There is logic so that when data import completes (even while the dialog is still open) the table refreshes, and it goes to the tab of the imported type.  I suspect that logic may be broken.
2. The launch analysis table refreshes when you select an item.  This isn't harmful but it's unnecessary.
3. For tables where (entity-count \* attribute-count) is large, the SSP call can take ~1 second.  We should add a spinner to this, but make it so it only appears after a delay (suggestion: 350ms) to avoid a flicker if the response is fast.
4. Error handling.  Currently the AJAX call silently fails.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -25,9 +25,23 @@
 ;; Table component with specifiable style and column behaviors.
 ;;
 ;; Properties:
-;;   :cell-padding-left (optional, default 16px)
-;;     A CSS padding-left value to apply to each cell
-;;   :paginator-space (optional, default 24)
+;;   :pagination (optional, default :internal)
+;;     Defines how the table is paginated.  Options are:
+;;       :internal -- data is given via the :data property and client-side pagination is provided
+;;       (fn [query-params callback] ...) -- paginate externally (e.g. server-side) by providing a function
+;;           that takes the query parameters and sets the data via a callback.
+;;           query-params structure:
+;;             {:current-page <1-indexed int>
+;;              :rows-per-page <positive int>
+;;              :filter-text <possibly empty string>
+;;              :sort-column <0-indexed int>
+;;              :sort-order <:asc or :desc>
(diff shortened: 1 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -329,60 +358,61 @@
...
-                       (swap! state dissoc :loading-entities?)
-                       (when-let [func (:on-filter-change props)]
-                         (func type)))}))))})
+   :pagination
+   (fn [{:keys [props state]} columns]
+     (let [{{:keys [entity-types]} :server-response} @state]
+       (fn [{:keys [current-page rows-per-page filter-text sort-column sort-order filter-group-index]} callback]
+         (let [type (nth entity-types filter-group-index)
+               sort-field (if (zero? sort-column)
+                            "name"
```
davidangb: is this the only place where the UI provides a default value? Otherwise, we use the defaults provided by the API? I'd like to make it consistent - all API defaults or all UI defaults, not mixed.

davidangb:
One comment inline, otherwise 👍

MattPutnam:
Good catch on the client-supplied default.  Rawls was already using the default of "name" when not specified.

dmohs:
👍  Code looks great and I love the functionality. Nicely done.

PR #454, opened 2016-06-13
Don't navigate the browser during sign-in

dmohs:
Fixes https://broadinstitute.atlassian.net/browse/GAWB-764.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -252,14 +252,40 @@
       [:div {}
        [comps/Button
         {:text "Sign In"
-         :href (str (config/api-url-root) "/login?callback="
-                    (js/encodeURIComponent
-                     (str js/window.location.protocol "//" js/window.location.hostname))
-                    (let [hash (nav/get-hash-value)]
-                      (if (clojure.string/blank? hash) "" (str "&path=" hash))))}]]
+         :onClick (fn [e]
+                    (.. js/window
+                        (open
+                         (str (config/api-url-root) "/login?callback="
+                              (js/encodeURIComponent (.. js/window -location -origin)))
+                         "Authentication"
+                         "menubar=no,toolbar=no,width=500,height=500")))}]]
       [:div {:style {:marginTop "1em"}} [RegisterLink]]
       [:div {:style {:maxWidth 600 :paddingTop "2em" :fontSize "small"}}
-        [Policy]]])})
+        [Policy]]])
+   :component-did-mount
+   (fn [{:keys [props]}]
+     (aset js/window (name ::handle-auth-message)
+           (fn [message] ((:on-login props) (get message "access_token")))))
+   :component-will-unmount
+   (fn [{:keys [locals]}]
+     (js-delete js/window (name ::handle-auth-message)))})
+
+
+(defn- parse-hash-into-map []
+  ;; Transforms the browser navigation hash into a map to make it easier to pass around. Technique
+  ;; stolen from Swagger's authentication flow.
+  (let [hash (nav/get-hash-value)
+        hash (clojure.string/replace hash #"^[?]" "")
+        parts (clojure.string/split hash #"[&=]")
+        parts (map js/decodeURIComponent parts)]
+    (apply assoc {} parts)))
+
+
+(defn- attempt-auth [token on-done]
+  ;; Use basic ajax call here to bypass default ajax-orch failure behavior.
+  (utils/ajax {:url (str (config/api-url-root) "/me")
+               :headers {"Authorization" (str "Bearer " token)}
+               :on-done on-done}))
 
 
 (defn- server-unavailable-dialog [dismiss]
@@ -285,7 +311,7 @@
    (fn []
      {:root-nav-context (nav/create-nav-context)})
    :render
-   (fn [{:keys [state]}]
+   (fn [{:keys [this state]}]
      [:div {}
       (when (:show-server-down-message? @state)
         (server-unavailable-dialog #(swap! state dissoc :show-server-down-message?)))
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -333,31 +361,14 @@
...
-                                   ;; 404 just means not registered
-                                   (or success? (= status-code 404))
-                                   (do (utils/set-access-token-cookie token)
-                                       (react/call :handle-successful-auth this token))
-                                   :else
-                                   (swap! state assoc :user-status :error)))))
+     (if-let [parent (.-opener js/window)]
+       ;; This window was used for login. Send the token to the parent. No need to render the UI.
+       (let [parsed (parse-hash-into-map)]
+         (.postMessage parent (clj->js parsed) (.. js/window -location -origin))
```
dmohs: Another option would be to call a method on the parent like

``` clojure
(.handleSignIn parent parsed)
```

which would require setting `handleSignIn` as a global function on `js/window`. I think this is cleaner at an identical level of security.

davidangb: can you use window.opener instead of postMessage, eliminating the use of postMessage entirely? Since window.opener has inherent same-origin policy enforced by the JavaScript VM, it's the stronger security option.

dmohs: Notice that `parent` is `window.opener`. I believe our only option is to choose between these two techniques: postMessage vs. a global function defined on the window object. The tradeoff is that a global function is messier (in my opinion) and could get accidentally clobbered by any other code running on the page with the benefit that you do not have to specify an origin to have same-origin enforced (as you said). I judge that no developer will replace `(.. js/window -location -origin)` with `"*"` either intentionally or accidentally, so the cleaner solution is better. I'm open to changing my mind about that judgement.

davidangb: I did mean using a global function, or even something like just setting parent.location, or messing with something inside the parent.document DOM - my point is eliminating postMessage entirely to remove any possibility of phishing/XSS.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -333,31 +371,16 @@
...
-                                   ;; 404 just means not registered
-                                   (or success? (= status-code 404))
-                                   (do (utils/set-access-token-cookie token)
-                                       (react/call :handle-successful-auth this token))
-                                   :else
-                                   (swap! state assoc :user-status :error)))))
+     (if-let [parent (.-opener js/window)]
+       ;; This window was used for login. Send the token to the parent. No need to render the UI.
+       (let [parsed (parse-hash-into-map)]
+         ;; (.postMessage parent (clj->js parsed) (.. js/window -location -origin))
```
davidangb: delete instead of comment

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -251,14 +254,48 @@
...
+                        (open
+                         (str (config/api-url-root) "/login?callback="
+                              (js/encodeURIComponent (.. js/window -location -origin)))
+                         "Authentication"
+                         "menubar=no,toolbar=no,width=500,height=500")))}]]
       [:div {:style {:marginTop "1em"}} [RegisterLink]]
       [:div {:style {:maxWidth 600 :paddingTop "2em" :fontSize "small"}}
-        [Policy]]])})
+        [Policy]]])
+   :component-did-mount-post-message
```
davidangb: delete component-did-mount-post-message and component-will-unmount-post-message? Those aren't used now, right?

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -251,14 +254,48 @@
...
+   :component-will-unmount-post-message
+   (fn [{:keys [locals]}]
+     (.removeEventListener js/window "message" (:message-listener @locals)))
+   :component-did-mount
+   (fn [{:keys [props locals]}]
+     (let [handle-message (fn [message] ((:on-login props) message))]
+       (aset js/window sign-in-handler-fn-name handle-message)))
+   :component-will-unmount
+   (fn [{:keys [locals]}]
+     (aset js/window sign-in-handler-fn-name nil))})
```
davidangb: (aset ... nil) or (js-delete ...)? It doesn't really matter, but delete is theoretically more correct?

davidangb:
A few syntax-level comments inline, otherwise looking good. Back to you, @dmohs.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: `js-delete` is exactly what I was looking for. Thanks!

dmohs:
Okay, tried another version using `postMessage`. I honestly believe our development team can be trusted with this usage, given the warning, and the code is objectively cleaner.

davidangb:
👎 
Between 33610b23855f913e8bf7a5ad57a020e93cb4a05b and 589150cbc414ca4bbf1a68615ef3d7f0135939d0 - the last two commits - this reverted from using a direct function call to using postMessage again. We had settled on a direct function call. Can you justify?

Edit: I see you addressed this issue above; sorry, missed the comment. Still, team decision was against postMessage. This feels unilateral.

dmohs:
I was not aware of a team decision. What I heard from the team was, "we don't understand this enough to weigh-in."

I feel I have addressed your concern, which I understood as, "a developer could naively copy the usage of this function and then modify it so that it allows for a phishing attack." If that understanding is correct, do you honestly believe it is still a hazard given the new function and comments? Or did I understand your concern incorrectly?

dmohs:
Discussed via video. I had understood your concern incorrectly. Sorry about that. New commit should address it better.

davidangb:
working for me! 👍 squash and merge.

PR #455, opened 2016-06-14
GAWB-735: Show dialog when anything returns 503

MattPutnam:
Right now there isn't any useful information in the 503, so the message is sorta generic.  The load balancing work will address this, at which point we can make the message nicer.

Note: I tested this locally by changing "503" to other things.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/dialog.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/dialog.cljs
@@ -16,26 +16,26 @@
       :cycle-focus? false})
    :render
    (fn [{:keys [props state]}]
-     (let [content (:content props)
-           anchored? (not (nil? (:get-anchor-dom-node props)))]
+     (let [{:keys [content width blocking? dismiss-self get-anchor-dom-node]} props
+           anchored? (not (nil? get-anchor-dom-node))]
        (assert (react/valid-element? content)
          (subs (str "Not a react element: " content) 0 200))
        (when (or (not anchored?) (:position @state))
-         [:div {:style {:backgroundColor (if (:blocking? props)
+         [:div {:style {:backgroundColor (if blocking?
                                            "rgba(110, 110, 110, 0.4)"
                                            "rgba(210, 210, 210, 0.4)")
                         :position "absolute" :zIndex 8888
                         :top 0 :left 0 :right 0 :height (.. js/document -body -offsetHeight)}
-                :onKeyDown (common/create-key-handler [:esc] #((:dismiss-self props)))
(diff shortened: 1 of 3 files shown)


davidangb:
I would love to see, as a general rule, more comments :)

`(fn [_ _ _ ns]` is reminiscent of:

```
@P=split//,".URRUU\c8R";@d=split//,"\nrekcah xinU / lreP rehtona tsuJ";sub p{
@p{"r$p","u$p"}=(P,P);pipe"r$p","u$p";++$p;($q*=2)+=$f=!fork;map{$P=$P[$f^ord
($p{$_})&6];$p{$_}=/ ^$P/ix?$P:close$_}keys%p}p;p;p;p;p;map{$p{$_}=~/^[P.]/&&
close$_}%p;wait until$?;map{/^r/&&<$_>}%p;$_=$d[$q];sleep rand(2)if/\S/;print
```

other than that, 👍

rushtong:
+1 on comments.

src/cljs/org/broadinstitute/firecloud_ui/common/dialog.cljs
```diff
@@ -16,26 +16,26 @@
       :cycle-focus? false})
    :render
    (fn [{:keys [props state]}]
-     (let [content (:content props)
-           anchored? (not (nil? (:get-anchor-dom-node props)))]
+     (let [{:keys [content width blocking? dismiss-self get-anchor-dom-node]} props
```
dmohs, line 19: Really loving this pattern.

dmohs:
The underscore is a common idiom (same as Scala) that just means, "I don't care about these parameters." I've seen this usage of `add-watch` in Clojure's docs:

``` clojure
(add-watch my-atom :some-key (fn [k r os ns] ...))
```

so the usage in this PR makes sense (since we don't care about the first three parameters).

However, I disagree with Clojure's docs in terms of variable naming here. The names `os` and `ns` are not descriptive enough and `ns` is a built-in for declaring a namespace, making this even worse. In our calls of `add-watch`, I'd suggest we use a more descriptive pattern:

``` clojure
(fn [_ _ _ new-value]
  (when new-value
    (swap! state assoc :show-server-down-message? true))))
```

No need to hold up this PR, but it'd be cool to see a small refactoring PR that fixes both calls to `add-watch` in our codebase. With that said, 👍

davidangb:
I'm not at all debating the proper use of syntax (or underscores) here ... I also ask for comments within Scala code!

davidangb:
👍

PR #456, opened 2016-06-17
Add a pending state for when a user has NIH linked but not yet grante…

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-695
This has a companion PR in Orchestration: https://github.com/broadinstitute/firecloud-orchestration/pull/226

This PR adds another pending state to the user profile. I am particularly interested in opinions on if this captures the case in the story correctly and completely.

Testing:
- See the companion PR in Orchestration
- Change line 19 in profile.cljs to point to the orchestration mock endpoint: `(config/api-url-root)` to mimic the effect of waiting for the results of an NIH account link process
# 
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

…d access by rawls

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/profile.cljs
@@ -22,6 +22,8 @@
         (let [loc (.-location js/window)]
           (str (.-protocol loc) "//" (.-host loc) "/#profile/nih-username-token={token}")))))
 
+(defn is-within-24-hours? [time]
+    (.isBefore (js/moment.) (.add time 24 "hours")))
 
 (react/defc NihLink
   {:render
@@ -32,7 +34,9 @@
            expired? (.isBefore expire-time (js/moment.))
            _24-hours-from-now (.add (js/moment.) 24 "hours")
            expiring-soon? (.isBefore expire-time _24-hours-from-now)
-           authorized? (get status "isDbgapAuthorized")]
+           authorized? (get status "isDbgapAuthorized")
+           linked-recently? (is-within-24-hours? (-> (get status "lastLinkTime") (* 1000) (js/moment.)))
+           pending? (and username (not authorized?) (not expired?) linked-recently?)]
        [:div {}
         [:h3 {} "Linked NIH Account"]
         (cond
@@ -59,9 +63,11 @@
            [:div {:style {:display "flex" :marginTop "1em"}}
             [:div {:style {:flex "0 0 20ex"}} "dbGaP Authorization:"]
             [:div {:style {:flex "0 0 auto"}}
-             (if authorized?
-               [:span {:style {:color (:success-green style/colors)}} "Authorized"]
-               [:span {:style {:color (:text-gray style/colors)}} "Not Authorized"])]]])]))
+              (cond
+                authorized? [:span {:style {:color (:success-green style/colors)}} "Authorized"]
+                pending? [:span {:style {:color (:success-green style/colors)}}
+                           "Your link was successful; you will be granted access shortly."]
+                :else [:span {:style {:color (:text-gray style/colors)}} "Not Authorized"])]]])]))
    :component-did-mount
    (fn [{:keys [this props state refs after-update]}]
      (let [nav-context (nav/parse-segment (:parent-nav-context props))


davidangb:
Nice simple change. I am worried about the expired use case, though ... if a user has previously successfully linked, but their link has timed out, I don't _believe_ we remove their linkedNihUsername - the user can log in, see how they were previously linked, but notice the link is no longer valid. In that case, won't this code show them as pending?

rushtong:
Good catch. I think I need to make sure the pending case does not include `expiring?` or `expiring-soon?`

davidangb:
I understand the inclusion of `expiring?` - maybe that's all we need?

I don't understand the inclusion of `expiring-soon?`. What's the reasoning behind including that? Granted, from reading the current UI code I don't see that this would cause a problem in what we display to the user - but I still don't understand the logic.

rushtong:
I included `expiring-soon?` because that looks like another form of `authorized?` state and didn't want to show the pending state language when that was also true.

davidangb:
Hm - maybe let's talk in person. If you're already checking `(and username (not authorized?) (not expired?))` I don't think `expiring-soon?` needs to be involved at all.

If the user is authorized and expiring soon, we won't show the pending language, because `authorized`.

If the user:
- linked with NIH
- waited 24 hours+ to log back in to firecloud (say, over a weekend)
- logged back in to Firecloud, went to their profile page

They would be in the condition of `expiring-soon?`. If they're still not authorized, what should we show?

Maybe this opens a bigger question: if the user links with NIH, but for some reason is not authorized within the next few minutes, what should we show?

dmohs:
Looking at the latest changes, check seems good to me now. 👍

rushtong:
@dmohs - I think @davidangb has a good point. We're not catching the related error case that David describes above. I think we should show an error message of some kind, but since there is no app-level process for sorting out what went wrong, I propose that we ask the user to follow up with a help ticket request.

davidangb:
(sorry about all the back and forth on this one; it's a small change, but just trying to get it right)

what if we added a new condition, something like `linked-recently?` that inspects the status' `lastLinkTime` and checks if the user linked within the last 24 hours? Add that condition to `pending?` (where `expiring-soon?` used to be). Then, if the user links but isn't authorized within 24 hours, we call it a "real" auth failure, and simply display "Unauthorized".

rushtong:
I think the let block might look something like this (caveat: untested):

```
[status (:nih-status @state)
           username (get status "linkedNihUsername")
           expire-time (-> (get status "linkExpireTime") (* 1000) (js/moment.))
           expired? (.isBefore expire-time (js/moment.))
           _24-hours-from-now (.add (js/moment.) 24 "hours")
           expiring-soon? (.isBefore expire-time _24-hours-from-now)
           authorized? (get status "isDbgapAuthorized")
           link-time (-> (get status "lastLinkTime") (* 1000) (js/moment.))
           _24-hours-from-last-link (.add link-time 24 "hours")
           linked-recently? (.isBefore (js/moment.) _24-hours-from-last-link)
           pending? (and username (not authorized?) (not expired?) (not linked-recently?))]
```

Am I the only one that finds this insanely ugly and complex?

davidangb:
that's definitely long and ugly. Some of that could be collapsed, e.g.

```
linked-recently? (.isBefore (js/moment.) (.add link-time 24 "hours"))
```

or, create a new function outside of this one to define linked-recently?

rushtong:
The `linked-recently?` logic isn't working in combination with pending. With this logic, there is never a pending state (and maybe this PR is unnecessary???)
- user is not linked
- user links
  - system updates user info and white list at the same time
- if the link worked, user is now authorized (pending is not true)
- if the link doesn't work, user is still unauthorized (pending is still not true)

If we want to have a linked-recently case, maybe there should be a separate error message?
Is linked-recently really the pending case?

rushtong:
@davidangb - I've updated the logic that seems to make the most sense to me at the moment. I'm considering pending to be the combination of `linked user` && `not authorized` && `not expired` && `linked recently`. After the time period of linked-recently is no longer in effect, the message drops back to Unauthorized.

I'm open to different approaches, but this seems like the right thing to me.

davidangb:
This is working great for me. I mucked with my values in Thurloe and was able to see the right combinations of messaging. Squash and merge! 👍

PR #462, opened 2016-06-27
GAWB-782: Add blocker and error handling to the paginated table

MattPutnam:
I recommend reviewing by commit.  Code was restructured so that all tables have the loading spinner, but it won't show up if the row reload is fast, which should always be the case for client-side paginated tables.  I set the time at 200ms (default; it's a parameter to the spinner), which was slow enough to not appear on a data table with ~50 entities, but fast enough to appear briefly for a table with ~10,000 entities.

We now have 2 blocker components, so we might want to move all existing ones to the new one in some future story.

Also left for a future story is the ability to style the error message or specify its content at the table level.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -139,6 +139,30 @@
          [Spinner {:text (:banner props)}]]]))})
 
 
+(react/defc SafeBlocker
+  {:show
+   (fn [{:keys [props state]}]
+     (swap! state assoc :show-requested? true)
+     (js/setTimeout #(when (:show-requested? @state)
+                      (swap! state assoc :showing? true))
+                    (:delay-time props)))
+   :hide
+   (fn [{:keys [state]}]
+     (swap! state dissoc :showing? :show-requested?))
+   :get-default-props
+   (fn []
+     {:delay-time 200})
+   :render
+   (fn [{:keys [props state]}]
+     (when (:showing? @state)
+       [:div {:style {:backgroundColor "rgba(210, 210, 210, 0.4)"
+                      :position "absolute" :top 0 :bottom 0 :right 0 :left 0 :zIndex 9999}}
+        [:div {:style {:position "absolute" :top "50%" :left "50%"
+                       :transform "translate(-50%, -50%)"
+                       :backgroundColor "#fff" :padding "2em"}}
+         [Spinner {:text (:banner props)}]]]))})
+
+
 (react/defc StatusLabel
   {:render
    (fn [{:keys [props]}]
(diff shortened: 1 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -139,6 +139,30 @@
...
+   :get-default-props
+   (fn []
+     {:delay-time 200})
+   :render
+   (fn [{:keys [props state]}]
+     (when (:showing? @state)
+       [:div {:style {:backgroundColor "rgba(210, 210, 210, 0.4)"
+                      :position "absolute" :top 0 :bottom 0 :right 0 :left 0 :zIndex 9999}}
+        [:div {:style {:position "absolute" :top "50%" :left "50%"
+                       :transform "translate(-50%, -50%)"
```
davidangb: can you explain this? looks like the top/left and transform cancel each other out.

MattPutnam: The top/left are relative to the parent, so they move the top left corner to the center of the parent.  The transform is relative to the child, so it moves it back up and to the left by half its size, putting the center at the center.

Although, a better way to do this may be with flex.

davidangb:
👍

src/cljs/org/broadinstitute/firecloud_ui/common/entity_table.cljs
```diff
@@ -0,0 +1,121 @@
+(ns org.broadinstitute.firecloud-ui.common.entity-table
```
dmohs, line 1: Moving this to a separate file was a good call.

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -139,6 +139,29 @@
          [Spinner {:text (:banner props)}]]]))})
 
 
+(react/defc SafeBlocker
```
dmohs: I'm not sure I see how this is "safer" than the other blocker. Maybe `DelayedBlocker`? Also, is there a way this could share more code with `Blocker` or even the new modal (or could the modal be modified to include this use case)?

Edit: Scratch the modal suggestion. I see this blocks the table only, so it wouldn't make sense.

Edit: I see your PR comments. I'm not going to push for merging the two blocker components, but I do worry that the cleanup activity will never happen.

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -139,6 +139,29 @@
...
+     (swap! state assoc :show-requested? true)
+     (js/setTimeout #(when (:show-requested? @state)
+                      (swap! state assoc :showing? true))
+                    (:delay-time props)))
+   :hide
+   (fn [{:keys [state]}]
+     (swap! state dissoc :showing? :show-requested?))
+   :get-default-props
+   (fn []
+     {:delay-time 200})
```
dmohs: Since the units (milliseconds) are not obvious, suggest the key be `:delay-time-ms`.

dmohs:
Code looks good. UI feels good. 👍

PR #467, opened 2016-07-01
fix create-link to allow other attributes

MattPutnam:
The `:title` key wasn't working because it wasn't a recognized parameter in create-link.  The right way is to provide default href and style using deep-merge.


MatthewBemis:
:+1: thanks for this!

PR #468, opened 2016-07-05
GAWB-798: Check refresh token before launching job

MattPutnam:
Dependent on Orchestration PR.  Will collide with the GAWB-741 PR, let that land first.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/method_config_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/method_config_editor.cljs
@@ -255,6 +255,7 @@
      (endpoints/call-ajax-orch
        {:endpoint (endpoints/get-refresh-token-date)
         :on-done (fn [{:keys [success?]}]
+                   ;; login checks validity of the token, so just check for existence
                    (swap! state assoc :has-refresh-token? success?))})
      (set! (.-onScrollHandler this)
            (fn []
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/method_config_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/method_config_editor.cljs
@@ -70,7 +70,7 @@
 
 (react/defc MethodDetailsViewer
   {:render
-   (fn [{:keys [props state]}]
+   (fn [{:keys [state]}]
      (cond
        (:loaded-method @state) [comps/EntityDetails {:entity (:loaded-method @state)}]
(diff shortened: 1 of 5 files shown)


davidangb:
Code/structure looks great (I haven't tried to run it, fyi).

Functionality-wise, I _think_ this is good enough, given the use cases/repro steps we know about. In the big picture, just checking that the refreshTokenDate API returns a 200 is not enough; for safety, we also check that the token is recent enough. See https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/service/OAuthService.scala#L124 for the implementation of that during login.

But, since the login phase handles that, and this check is only for weird corner cases, I think we're okay.

I mention all this just to make sure we're all on the same page. If we are, 👍

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -250,6 +252,10 @@
                    (if success?
                      (swap! state assoc :locked? (get-in (get-parsed-response) ["workspace" "isLocked"]))
                      (swap! state assoc :error status-text)))})
+     (endpoints/call-ajax-orch
+       {:endpoint (endpoints/get-refresh-token-date)
+        :on-done (fn [{:keys [success?]}]
+                   (swap! state assoc :has-refresh-token? success?))})
```
dmohs, line 259: To @davidangb's point, I think this is exactly the right check here, since the real check is done in the API on login, so verifying the date would be duplicated code. However, since I'm naturally inclined to ask the question, "Why is this only checking for success?" a comment like:

```
;; Only check for the existence of this key, since the validity is checked by the API on login
```

would be helpful.

dmohs:
@davidangb Does this mean that if a user is activated and then attempts to launch a job by calling the API instead of going through the UI, they'll still hit the error? To put it another way, would it be more appropriate to do this check during the "launch a submission" API call so it can be caught in all cases?

davidangb:
@dmohs yes, I can see a case where a user registers/activates via the UI, and then uses only the API; that user would be at risk of a stale refresh token. BUT, the 6-month check during login is a possibly overkill - as far as we know refresh tokens don't expire. so the error is unlikely.

I think the appropriate fix is to have the refreshTokenDate endpoint actually validate the token; then any callers that use refreshTokenDate will have the latest info. This is filed as GAWB-847.

davidangb:
👍

dmohs:
👍 Maybe file another story as we discussed.

PR #474, opened 2016-07-15
GAWB-808 added maintenance mode message for when we get a 502 

skwalker:
For [GAWB-808](https://broadinstitute.atlassian.net/browse/GAWB-808) -- this PR displays a maintenance mode message when we receive a 502 error, and any other 5XX error will produce the temporarily unavailable message. 
- [ ] **Submitter**: Tell tech lead that the PR exists if s/he wants to look at it
- [x] **Tech lead**: sign off
- [ ] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
@@ -20,20 +20,19 @@
       {:text "Sign In"
        :onClick (fn [e]
                   (u/ajax {:url (str (config/api-url-root) "/me")
-                                  :on-done (fn [{:keys [success? status-code]}]
-                                    (cond
-                                      (= status-code 502 )
+                           :on-done (fn [{:keys [success? status-code]}]
+                                      (cond
+                                        (= status-code 502 )
                                         (reset! u/maintenance-mode? true)
-                                      (contains? (set (range 500 600)) status-code)
+                                        (contains? (set (range 500 600)) status-code)
                                         (reset! u/server-down? (atom true))
-                                      :else
+                                        :else
                                         (.. js/window
-                                          (open
-                                           (str (config/api-url-root) "/login?callback="
-                                                (js/encodeURIComponent (.. js/window -location -origin)))
-                                           "Authentication"
-                                           "menubar=no,toolbar=no,width=500,height=500")))
-                                  )}))}])
+                                            (open
+                                              (str (config/api-url-root) "/login?callback="
+                                                   (js/encodeURIComponent (.. js/window -location -origin)))
+                                              "Authentication"
+                                              "menubar=no,toolbar=no,width=500,height=500"))))}))}])
    :component-did-mount
    (fn [{:keys [props]}]
      (aset js/window handler-fn-name
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
@@ -19,12 +19,21 @@
      [comps/Button
       {:text "Sign In"
        :onClick (fn [e]
-                  (.. js/window
-                      (open
-                       (str (config/api-url-root) "/login?callback="
-                            (js/encodeURIComponent (.. js/window -location -origin)))
-                       "Authentication"
-                       "menubar=no,toolbar=no,width=500,height=500")))}])
+                  (u/ajax {:url (str (config/api-url-root) "/me")
+                                  :on-done (fn [{:keys [success? status-code]}]
+                                    (cond
+                                      (= status-code 502 )
+                                        (reset! u/maintenance-mode? true)
+                                      (contains? (set (range 500 600)) status-code)
+                                        (reset! u/server-down? (atom true))
+                                      :else
+                                        (.. js/window
+                                          (open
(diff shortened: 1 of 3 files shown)


MattPutnam:
I think we should do this a little differently.  Right now, we could get a 503 and a 502 in close succession and get both dialogs.  I think the logic should be `if (502) show-maintenance else if (5xx) show-server-down`, and then we can also reuse a single flag and dialog.

davidangb:
I assume you've coordinated with @rushtong to know that 502 is the appropriate maintenance code returned by the orchestration load balancer? If so, 👍 as tech lead

rushtong:
This isn't working for me on login. When I set orchestration to return nothing but 502 errors, I do not get the dialog box on sign in, but instead get the wrong error dialog.

davidangb:
@rushtong I bet that's specific to the /me call, which doesn't use the same ajax routing as other calls. Though to the end user, there's no difference.

rushtong:
@davidangb - Agreed. I'm back at my desk and Sarah and I are both looking at it. I think that call might need some extra handling for this case.

rushtong:
One more piece of feedback. After clicking the Maintenance Mode OK dialog button, the user is dropped back to the page with a generic "Bad Gateway" error message displayed in the banner. All links still work, but with that error displayed. I think this deserves some discussion as to the correct behavior to present to the user. 

Ideas: 
- Removing the "OK" button so the user isn't brought back to a broken site.
- Changing the banner message to match the "Maintenance Mode" dialog.
- Redirecting user to the main page (perhaps a form of logout?)

davidangb:
👍 . I think Greg makes some good comments about the UX of the "OK" button, but I'm fine pushing this PR through as-is and making that a follow-on story.

rushtong:
👍 - please do add follow-on stories as necessary

skwalker:
@rushtong created a new story for better error handling [GAWB-886](https://broadinstitute.atlassian.net/browse/GAWB-886)

src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
```diff
@@ -19,12 +19,21 @@
...
       {:text "Sign In"
        :onClick (fn [e]
-                  (.. js/window
-                      (open
-                       (str (config/api-url-root) "/login?callback="
-                            (js/encodeURIComponent (.. js/window -location -origin)))
-                       "Authentication"
-                       "menubar=no,toolbar=no,width=500,height=500")))}])
+                  (u/ajax {:url (str (config/api-url-root) "/me")
+                                  :on-done (fn [{:keys [success? status-code]}]
```
dmohs: We generally follow this style guide: https://github.com/bbatsov/clojure-style-guide

Please do a style pass through this PR. If your editor of choice doesn't automatically format the code correctly, I'm happy to come help configure it.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -275,21 +275,33 @@
...
+       :ok-button [comps/Button {:text "OK" :onClick dismiss}]
+       :content
+       [:div {}
+          "FireCloud service is temporarily unavailable.  If this problem persists, check "
+          [:a {:href "http://status.firecloud.org/" :target "_blank"}
+           "http://status.firecloud.org/"]
+          " for more information."]})))
+
+
+  (defn- show-maintenance-mode-dialog [dismiss]
```
dmohs: This function duplicates a bit too much code from the previous. Maybe these two can be combined into:

``` clojure
(defn- show-system-status-dialog [maintenance-mode? dismiss]
...)
```

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -407,6 +433,10 @@
    (fn [{:keys [this state]} token]
      (attempt-auth token (fn [{:keys [success? status-code]}]
                            (cond
+                             (= status-code 502)
+                               (swap! utils/maintenance-mode? not)
```
dmohs: This says, "If you see a 502 and utils/maintenance-mode? is false, make it true, but if it is true, make it false." I don't think that is quite right.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -152,8 +153,10 @@
            :headers (merge {"Authorization" (str "Bearer " @access-token)}
                            (:headers arg-map))
            :on-done (fn [{:keys [status-code] :as m}]
-                      (when (not= @server-down? (contains? #{500 503} status-code))
-                        (swap! server-down? not))
+                      (when (and (not @server-down?)  (not @maintenance-mode?))
+                        (cond
+                          (= status-code 502) (swap! maintenance-mode? not)
```
skwalker: @dmohs I left the `swap maintenance-mode? not` here because Matt said he had issues with the modal popping up multiple times (apparently the server-watcher wasn't working as expected, it was getting triggered even if you were just keeping the value at true). But I got rid of it in the other places, and it seems to be working fine.

skwalker:
I changed the modals so there's no "ok" button -- seems like it's what Alex wants, given his comment on [GAWB-886](https://broadinstitute.atlassian.net/browse/GAWB-886)

PR #479, opened 2016-08-03
Suggestions for 810

dmohs:


MattPutnam:
Will QA be able to uniquely identify the button like this?  Just on the workspaces list, the create button and the filter button are both a.button now.

dmohs:
Yes. Because class names are not globally-unique, the accessors available in things like Selenium return a list. If this turns out to be less robust than QA needs, we can add additional class names to either the button or somewhere higher up in the chain.

PR #482, opened 2016-08-05
cleanup

MattPutnam:
I'm having trouble finding good test cases, so this is mostly untested.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -6,7 +6,6 @@
     [org.broadinstitute.firecloud-ui.common.codemirror :refer [CodeMirror]]
     [org.broadinstitute.firecloud-ui.common.icons :as icons]
     [org.broadinstitute.firecloud-ui.common.style :as style]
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -211,21 +210,16 @@
...
+             [:div {}
+              [:span {:style {:fontWeight 500 :width 100 :display "inline-block" :paddingBottom "0.3em"}} label]
+              (if (and editing? dropdown?)
+                (style/create-identity-select {:ref key
+                                               :style {:width "100px"}
+                                               :defaultValue (entity key)
+                                               :onChange (when-let [f (:onSnapshotIdChange props)]
+                                                           #(f (int (common/get-text refs "snapshotId"))))}
+                                              (:snapshots props))
+                [:span {} ((or render identity) (entity key))])])
```
MattPutnam, line 222: Factored the common stuff out.

I don't quite like the way this works--it seems like `:editing?` and `:onSnapshotIdChange` effectively duplicate each other.  But it's okay.

MatthewBemis:
Changes make sense to me. I also tested this out and everything still worked.

thanks for cleaning it up! i'll merge

PR #486, opened 2016-08-10
Improve token expiration behavior

rushtong:
Fixes https://broadinstitute.atlassian.net/browse/GAWB-852

This PR changes the expiration behavior. When your account expires, a modal pops up that asks you to sign in again. After a successful sign in, the modal goes away and whatever you were doing on the page is uninterrupted.
### Testing:

To test this, you need to expire your account manually and then make sure our proxies are not caching the results. Make sure you can see the network tab in your browser and copy the bearer token value. Enter that here:
- https://accounts.google.com/o/oauth2/revoke?token= {insert bearer token here}

Next, restart the proxies where the testing is happening. When running a local UI against develop, this restarts both of the proxies behind the load balancer:

```
ssh -t firecloud-orchestration01.dsde-dev.broadinstitute.org sudo docker restart firecloudorchestration_proxy_1 && ssh -t firecloud-orchestration02.dsde-dev.broadinstitute.org sudo docker restart firecloudorchestration_proxy_1
```

In the UI, do any kind of navigation which should now trigger a 401 error. After sign-in, the page should be unchanged. 

---
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
@@ -71,9 +71,14 @@
                                   (let [token (get message "access_token")]
                                     (reset! u/access-token token)
                                     (u/set-access-token-cookie token)
-                                    (swap! state assoc :success? true)
-                                    (when-let [callback (:callback props)]
-                                      (callback))))}]]])]]))
+                                    (swap! state assoc :success? true)))}]]])]]))
+   :component-did-update
+   (fn [{:keys [state props]}]
+     (let [{:keys [success?]} @state]
+       (when success?
+         (modal/pop-modal)
+         (when-let [callback (:callback props)]
+           (callback)))))
    :component-will-mount
    (fn []
      (set! showing-dialog? true))
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
@@ -60,18 +60,14 @@
                        :fontSize "137%" :fontWeight 400 :lineHeight 1}}
          (:title props)]
         [:div {:style {:padding "22px 48px 40px" :backgroundColor (:background-gray style/colors)}}
-         (if success?
-           [:div {:style {:color (:success-green style/colors)}} "Sign-in complete."
-            [:div {:style {:marginTop 20 :textAlign "center"}}
-             [comps/Button {:text "Close" :onClick modal/pop-modal}]]]
-           [:div {}
+          [:div {}
             (:message props)
             [:div {:style {:marginTop 20 :textAlign "center"}}
-             [Button {:on-login (fn [message]
-                                  (let [token (get message "access_token")]
-                                    (reset! u/access-token token)
-                                    (u/set-access-token-cookie token)
-                                    (swap! state assoc :success? true)))}]]])]]))
+              [Button {:on-login (fn [message]
+                                   (let [token (get message "access_token")]
+                                     (reset! u/access-token token)
+                                     (u/set-access-token-cookie token)
+                                     (swap! state assoc :success? true)))}]]]]]))
    :component-did-update
    (fn [{:keys [state props]}]
      (let [{:keys [success?]} @state]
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
```diff
@@ -71,9 +71,14 @@
...
-                                    (when-let [callback (:callback props)]
-                                      (callback))))}]]])]]))
+                                    (swap! state assoc :success? true)))}]]])]]))
+   :component-did-update
+   (fn [{:keys [state props]}]
+     (let [{:keys [success?]} @state]
+       (when success?
+         (modal/pop-modal)
+         (when-let [callback (:callback props)]
+           (callback)))))
```
MattPutnam, line 77: Why is this in `:component-did-update`?  Looks like you'd achieve the same thing by just putting `(modal/pop-modal)` at line 74.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -158,9 +159,16 @@
...
                           (contains? (set (range 500 600)) status-code) (reset! server-down? true)))
                       ;; Handle auth token expiration
-                      (when (and (= status-code 401) (not ignore-auth-expiration?))
-                        (auth-expiration-handler))
-                      (on-done m))))))
+                      (if (and (= status-code 401) (not ignore-auth-expiration?))
+                        (do
+                          (swap! pending-calls conj [path arg-map
+                                                     :service-prefix service-prefix
+                                                     :ignore-auth-expiration true])
```
MattPutnam: Is this the right signature to be storing?  Shouldn't it be:

``` clojure
[path arg-map
 {:service-prefix service-prefix
  :ignore-auth-expiration true}]
```

?

rushtong: Both work. My original version used the braces, David Mohs' update uses this one. I lean towards using the brace syntax to be consistent with other code in this area.

src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
rushtong, line 77: This is an artifact of trying different ways to execute the callbacks at the right time (which was very frustrating). I'm fine with moving it back to the button callback.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
dmohs: I don't believe you :)

In fact, both don't work. The only one that works is no brackets and spelling `:ignore-auth-expiration?` (with the question mark) correctly.

src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
dmohs, line 77: I'm not sure React would be okay with making UI changes during a render pass. This might be safer.

However, there is now an `if` statement on 63 that doesn't make sense anymore, along with the close button.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -386,7 +386,7 @@
       [modal/Component {:ref "modal"}]])
    :component-did-mount
    (fn [{:keys [this state refs]}]
-     (set! utils/auth-expiration-handler #(sign-in/show-sign-in-dialog :expired))
+     (set! utils/auth-expiration-handler (partial sign-in/show-sign-in-dialog :expired))
```
dmohs, line 389: This was a pain-in-the-ass to debug. I'd like us to do a quick huddle on how we could improve the situation. There may be no answer, but I'd at least like to have the conversation.

src/cljs/org/broadinstitute/firecloud_ui/common/sign_in.cljs
rushtong, line 77: Yes, that condition can go away.

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
rushtong: Don't believe anything I say :)
Will fix.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
rushtong, line 389: If we had a logout API call in orchestration, it would be easier.

dmohs:
One more note: In screwing around with this I found a less-nuclear option for invalidating the authentication token. In the REPL, do:

``` clojure
(require '[org.broadinstitute.firecloud-ui.utils :as u])
(swap! u/access-token str "a")
```

PR #489, opened 2016-08-12
GAWB-947: billing project dropdown should only display project names

MatthewBemis:
- [x] **Tech lead**: sign off
- [ ] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/create.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/create.cljs
@@ -98,20 +98,11 @@
                      :no-billing (str "You must have a billing project associated with your account"
                                       " to create a new workspace.")
                      "Project billing data failed to load.")
-        :onClick #(modal/push-modal [CreateDialog {:billing-projects (:billing-projects @state)
+        :onClick #(modal/push-modal [CreateDialog {:billing-projects (:billing-projects props)
                                                    :nav-context (:nav-context props)}])}]])
    :component-did-mount
-   (fn [{:keys [state]}]
-     (endpoints/call-ajax-orch
-      {:endpoint (endpoints/get-billing-projects)
-       :on-done (fn [{:keys [success? get-parsed-response]}]
-                  (if success?
-                    (let [billing-projects (flatten (->> (get-parsed-response)
-                                                         (map utils/keywordize-keys)
-                                                         (map #(select-keys % [:projectName]))
-                                                         (map #(vals %))))]
-                      (swap! state assoc :billing-projects billing-projects)
-                      (if (empty? billing-projects)
-                        (swap! state assoc :disabled-reason :no-billing)
-                        (swap! state dissoc :disabled-reason)))
-                    (swap! state assoc :disabled-reason :error)))}))})
+   (fn [{:keys [props state]}]
+     (let [disabled-reason (:disabled-reason props)]
+       (if (some? disabled-reason)
+         (swap! state assoc :disabled-reason disabled-reason)
+         (swap! state dissoc :disabled-reason))))})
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/create.cljs
```diff
@@ -106,7 +106,10 @@
       {:endpoint (endpoints/get-billing-projects)
        :on-done (fn [{:keys [success? get-parsed-response]}]
                   (if success?
-                    (let [billing-projects (get-parsed-response)]
+                    (let [billing-projects (flatten (->> (get-parsed-response)
+                                                         (map utils/keywordize-keys)
+                                                         (map #(select-keys % [:projectName]))
+                                                         (map #(vals %))))]
```
MattPutnam: This is way overkill.  All you need is

``` clojure
(let [billing-projects (map #(% "projectName") (get-parsed-response))]
  ...)
```

MattPutnam:
Hmm... I noticed that the get-billling-projects endpoint is being called twice.  Looks like WorkspacesList calls it and passes it to WorkspaceTable, but WorkspaceTable doesn't pass it to create/Button and create/Button hits it again.  Can you update this so that create/Button takes the projects as a prop and makes no ajax call on load?

MatthewBemis:
@MattPutnam back to you

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/create.cljs
```diff
@@ -98,17 +98,11 @@
...
-                      (swap! state assoc :billing-projects billing-projects)
-                      (if (empty? billing-projects)
-                        (swap! state assoc :disabled-reason :no-billing)
-                        (swap! state dissoc :disabled-reason)))
-                    (swap! state assoc :disabled-reason :error)))}))})
+   (fn [{:keys [props state]}]
+     (let [disabled-reason (:disabled-reason props)]
+       (if (some? disabled-reason)
+         (swap! state assoc :disabled-reason disabled-reason)
+         (swap! state dissoc :disabled-reason))))})
```
MattPutnam: This chunk doesn't seem quite right.  Now that all of this is passed in there shouldn't be a need for `:component-did-mount` at all.  This check should be done where the button is rendered.

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -194,7 +196,7 @@
          :else
          [:div {:style {:margin "0 2em"}}
           [WorkspaceTable
-           (assoc props :workspaces workspaces :billing-projects billing-projects)]])))
+           (assoc props :workspaces workspaces :billing-projects billing-projects :disabled-reason error-message)]])))
```
MattPutnam: I don't think this is right.  The error message is kind of an infodump; this flag was designed to be a keyword (`:no-billing` etc.)

MatthewBemis:
@MattPutnam back to you with latest fixes

MattPutnam:
@MatthewBemis Check out what I just pushed and let me know if you have any questions.  WorkspaceList is using David Mohs's pattern of putting all ajax responses in a `:server-response` within state, so the disabled reason wasn't hooked up right.  Also, you weren't detecting the empty case.  Then I moved the map so that we're never holding onto more info than we need.

MatthewBemis:
@MattPutnam nice catch with the empty case- totally slipped my mind to re-implement that bit. the server-response pattern was a bit unclear to me at first but i see with your commit how to work with it now. thanks for the help!

MattPutnam:
👍

davidangb:
👍 as TL

PR #492, opened 2016-08-15
status code 0 means we're in maintenance mode

skwalker:
When the orch load balancer is down, we don't get any cors headers, so the maintenance mode modal doesn't pop up. Instead, we can respond to a status code of 0 for maintenance mode. I don't think there's any way to make sure that it's actually a 502, not a 0 for some other reason, but I could be wrong
- [ ] **Submitter**: Rebase to develop. DO NOT SQUASH
- [ ] **Submitter**: Make sure Swagger is updated if API changes
- [ ] **Submitter**: Make sure documentation for code is complete
- [ ] **Submitter**: Review code comments; remove done TODOs, create stories for remaining TODOs
- [ ] **Submitter**: Add description or comments on the PR explaining the hows/whys (if not obvious)
- [ ] **Submitter**: Tell tech lead that the PR exists if s/he wants to look at it
- [ ] **Submitter**: Anoint a lead reviewer (LR). **Assign PR to LR**
- [x] **Tech lead**: sign off
- [ ] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -466,7 +466,7 @@
       (fn [{:keys [success? status-code get-parsed-response]}]
         (swap! state assoc :user-email (get-in (get-parsed-response) ["userInfo" "userEmail"]))
         (cond
-          (= status-code 0)
+          (contains #{0 502} status-code)
           (do (reset! utils/maintenance-mode? true) (swap! state assoc :user-status :error))
           (contains? (set (range 500 600)) status-code)
           (do (reset! utils/server-down? true) (swap! state assoc :user-status :error))
--- a/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/main.cljs
@@ -466,7 +466,7 @@
       (fn [{:keys [success? status-code get-parsed-response]}]
         (swap! state assoc :user-email (get-in (get-parsed-response) ["userInfo" "userEmail"]))
         (cond
-          (= status-code 502)
+          (= status-code 0)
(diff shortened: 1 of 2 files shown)


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -466,7 +466,7 @@
       (fn [{:keys [success? status-code get-parsed-response]}]
         (swap! state assoc :user-email (get-in (get-parsed-response) ["userInfo" "userEmail"]))
         (cond
-          (= status-code 502)
+          (= status-code 0)
```
rushtong: I would suggest we handle both cases here (and below), 502 or 0.

davidangb: agree on checking for both.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -466,7 +466,7 @@
       (fn [{:keys [success? status-code get-parsed-response]}]
         (swap! state assoc :user-email (get-in (get-parsed-response) ["userInfo" "userEmail"]))
         (cond
-          (= status-code 502)
+          (contains #{0 502} status-code)
```
MattPutnam: Should be `contains?`, right?  Also, an idiomatic way to do a set membership check is to just use the set itself:

```
(#{0 502} status-code)
```

But explicitly doing a `contains?` is fine too.

skwalker: sorry, that's what I meant.. I think I'll leave the `contains?` just because it looks a little clearer to me

MattPutnam:
👍 I'm a little iffy on how the load balancer works and everything, but I think this is correct.

davidangb:
👍 as TL

PR #495, opened 2016-08-16
GAWB-177: Make workspace attribute fields bigger

MattPutnam:
We'll need a split pane for the IGV stuff, according to Alex's mockup.  Proof of concept is a side-by-side markdown editor/preview.

This split pane implementation is probably incomplete; the scrolling/overflow behavior has
been tailored to work specifically in this case.  It will likely need to expose style
hooks to be used with other types of child components.
- [x] **Tech lead**: sign off
- [x] **LR**: sign off
  working for ANY reason!

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
@@ -15,48 +15,51 @@
   [:div {:style {:padding "1em 0 2em 0"}} children])
 
 (defn view-attributes [state refs]
-  [:div {:style {:display "inline-block"}}
+  [:div {}
    (style/create-section-header "Workspace Attributes")
    (create-section
      (when (or (:saving? @state) (:deleting? @state))
        [comps/Blocker {:banner "Updating..."}])
      [:div {}
       (map-indexed
-        (fn [i a]
-          [:div {}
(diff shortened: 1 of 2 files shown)


skwalker:
👍  the sidebar sliding thing is cool!

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
```diff
@@ -15,48 +15,51 @@
...
      (when (or (:saving? @state) (:deleting? @state))
        [comps/Blocker {:banner "Updating..."}])
      [:div {}
       (map-indexed
-        (fn [i a]
-          [:div {}
-           [:div {:style {:float "left" :marginRight "0.5em"}}
+        (fn [i [attr-key attr-value]]
+          [:div {:style {:display "flex" :alignItems "baseline"}}
+           [:div {:style {:flex "1 0 auto" :marginRight "0.5em"}}
```
davidangb: maybe go for something like `:flex-basis "35%"` here, instead of making both columns equal? I like the end result of this PR much better than previous state, but after this PR the fields can almost be TOO big. I argue that the attr key needs less space than the attr value, so something like 35%:65% might be good. Thoughts?

MattPutnam:
Good idea with the 35/65.

davidangb:
👍 as TL!

PR #499, opened 2016-08-23
IGV tweaking

davidangb:


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/analysis/igv.cljs
```diff
@@ -18,6 +18,8 @@
      :tracks (map-indexed (fn [index track]
                             {:name (str "Track " (inc index))
                              :url track
+                             :indexUrl (clojure.string/replace track #".bam" ".bai")
+                             :type "bam"
```
davidangb, line 22: I'm not sure if the syntax for replace() in the indexUrl field is correct, but you can guess at what I'm trying to do here ... 

Also, hardcoding type=bam obviously won't work in the long run, but it seems IGV wants this ...

src/static/index.html
```diff
@@ -19,7 +19,7 @@
     <link rel="stylesheet" type="text/css" href="https://igv.org/web/release/1.0.1/igv-1.0.1.css">
     <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.1/jquery.min.js"></script>
     <script type="text/javascript" src="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.2/jquery-ui.min.js"></script>
-    <script type="text/javascript" src="https://igv.org/web/release/1.0.1/igv-1.0.1.js"></script>
+    <script type="text/javascript" src="assets/js/igv-1.0.1.js"></script>
```
davidangb, line 22: I had to pull IGV locally so I could tweak it. See the second commit in this PR to see what I changed.

PR #502, opened 2016-08-26
GAWB-993: Cache get-workspace to reduce repeated calls

MattPutnam:
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/tab.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/tab.cljs
@@ -13,6 +13,7 @@
     [org.broadinstitute.firecloud-ui.page.method-repo.method-config-importer :refer [MethodConfigImporter]]
     [org.broadinstitute.firecloud-ui.page.workspace.method-configs.method-config-editor :refer [MethodConfigEditor]]
     [org.broadinstitute.firecloud-ui.utils :as utils]
+    [org.broadinstitute.firecloud-ui.workspace-cache :refer [get-workspace]]
     ))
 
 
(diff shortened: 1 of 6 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/tab.cljs
```diff
@@ -70,26 +71,19 @@
    :component-did-mount
    (fn [{:keys [this]}]
      (react/call :load this))
-   :component-did-update
-   (fn [{:keys [this state]}]
-     (when (nil? (:server-response @state))
-       (react/call :load this)))
```
MattPutnam, line 76: Not sure why this was here, it might have been a holdover from the old load-counter implementation (see deletions below).  This was causing an infinite loop of loading, and removing it doesn't seem to break anything.

src/cljs/org/broadinstitute/firecloud_ui/workspace_cache.cljs
```diff
@@ -0,0 +1,33 @@
...
+    (do
+      (swap! cache assoc workspace-id :pending)
+      (swap! awaiting assoc workspace-id [on-done])
+      (endpoints/call-ajax-orch
+        {:endpoint (endpoints/get-workspace workspace-id)
+         :on-done (fn [response]
+                    (swap! cache assoc workspace-id response)
+                    (js/setTimeout #(swap! cache dissoc workspace-id) 1000)
+                    (doseq [on-done (get @awaiting workspace-id)]
+                      (on-done response))
```
davidangb: if one of the on-done functions throws an error, what happens? Do the others never get called? Do we never `dissoc` any of them out of the awaiting atom?

MattPutnam: Yeah, errors would be bad.  That can be try/catched, but @dmohs brought up the concern that this is actually exposing a problem with our API.  Several places call this endpoint because they need just a small bit of info, for example checking if the workspace is locked.  Perhaps there needs to be an internal endpoint just for those kinds of checks.

src/cljs/org/broadinstitute/firecloud_ui/workspace_cache.cljs
```diff
@@ -0,0 +1,33 @@
...
+      (on-done workspace-response))
+    (do
+      (swap! cache assoc workspace-id :pending)
+      (swap! awaiting assoc workspace-id [on-done])
+      (endpoints/call-ajax-orch
+        {:endpoint (endpoints/get-workspace workspace-id)
+         :on-done (fn [response]
+                    (swap! cache assoc workspace-id response)
+                    (js/setTimeout #(swap! cache dissoc workspace-id) 1000)
+                    (doseq [on-done (get @awaiting workspace-id)]
```
davidangb: rename `on-done` here to avoid confusion? Or am I misreading the code, and you're actually referring to the on-done passed in to the outer get-workspace function?

MattPutnam: You're right.  I'll rename.

davidangb:
I like the concept of this, but caches are so hard. I think there's a corner-case race condition here. If multiple locations in the rest of the codebase call `get-workspace` over a long-ish (> 1 second) period of time, the setTimeout function could trigger while there are callbacks on the `awaiting` stack. If another caller calls `get-workspace` just after this - and before their ajax call resolves - this new caller will clear `awaiting`, meaning there are callbacks that never get processed.

This is unlikely to happen, but it looks theoretically possible - stems from having two different atoms that get written/cleared at different times.

PR #521, opened 2016-09-20
GAWB-987 (partial)

MattPutnam:
This is not complete, this is a checkpoint.
- Swapped out icons across the application.  Please be on the lookout for places where icons are not showing, or are styled poorly.
- Changed how the main get-workspace call parses the response.  It now separates out description, workspace attributes, and library attributes, and also converts keys to keywords.  This is to help with separation of concerns and let library.cljs know as little as possible.
- Library attribute form and display are partly working.  Theoretically, if a workspace were to have library attributes they would display correctly.  This is commented out until actually ready.  Find the 3 "TODO" comments in summary/tab.cljs to see for yourself.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -65,12 +63,10 @@
...
-        (or text (icons/icon-text icon))
-        (when (= type :add)
-          [:span {:style {:display "inline-block" :height "1em" :width "1em" :marginLeft "1em"
-                          :position "relative"}}
-           [:span {:style {:position "absolute" :top "-55%" :fontSize "200%" :fontWeight "normal"}}
-            "+"]])]))})
+        text
+        (some->> icon (icons/icon {:style (if text
+                                            {:fontSize 24 :margin "-0.5em -0.3em -0.5em 0.5em"}
+                                            {:fontSize 20})}))]))})
```
MattPutnam, line 69: Changed the button to be able to take text and an icon key.  Got rid of ":type :add" and just made it the plus icon.

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -327,16 +323,16 @@
              ;; method redact is responding with "code" for 401.  TODO: standardize and remove this extra logic
              status-code (or status-code code)]
          (if-let [expected-msg (get-in props [:expect status-code])]
-           [:div {}
+           [:div {:style {:display "flex" :alignItems "center"}}
```
MattPutnam: A lot of little alignment things popped up when I changed the icons.  This is more robust anyway.

src/cljs/org/broadinstitute/firecloud_ui/common/icons.cljs
```diff
@@ -5,30 +5,33 @@
...
+   :remove "remove_circle"
+   :done "done"
+   :cancel "cancel"
+   :warning-triangle "warning"
+   :error "error"
+   :search "search"
+   :share "share"
+   :library "import_contacts"
+   :publish "publish"
+   :catalog "local_offer"})
```
MattPutnam, line 28: Now all you have to do to introduce a new icon is find the one you like and register it here.

src/cljs/org/broadinstitute/firecloud_ui/common/table_utils.cljs
```diff
@@ -24,65 +24,73 @@
...
                "rows per page"]]
-          [:div {:style {:fontSize 13 :lineHeight 1.5 :padding "0px 48px"}}
-           view-component page-component rows-component
-           (common/clear-both)])]))})
+          [:div {:style {:fontSize 13 :lineHeight 1.5 :padding "0 48px"
+                         :display "flex" :flexDirection (if narrow? "column" "row")
+                         :alignItems (if narrow? "center" "baseline")}}
+           (container view-component "left")
+           (container page-component "center")
+           (container rows-component "right")])]))})
```
MattPutnam, line 93: The layout of the paginator went to shit surprisingly hard with the icon swap.  So I had to completely redo the layout.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/details.cljs
```diff
@@ -73,6 +73,20 @@
...
+        library-attributes (->> attributes
+                                (keep (fn [[k v]] (when (.startsWith k "library:") [(keyword (subs k 8)) v])))
+                                (into {}))
+        workspace-attributes (utils/keywordize-keys
+                               (apply dissoc attributes "description" (keys library-attributes)))]
+    (-> (utils/keywordize-keys raw-workspace)
+        (update-in [:workspace] dissoc :attributes)
+        (assoc-in [:workspace :description] (attributes "description"))
+        (assoc-in [:workspace :workspace-attributes] workspace-attributes)
+        (assoc-in [:workspace :library-attributes] library-attributes))))
```
MattPutnam, line 89: Translate attributes into their 3 buckets so it's easier to work with.

src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
```diff
@@ -711,6 +711,120 @@
   {:path (str "/billing/" project-id "/" role "/" user-email)
    :method :delete})
 
+(defn get-library-attributes [on-done]
+  (utils/ajax-orch
+    "/library-attributedefinitions-v1"
+    {:method :get
+     :on-done on-done
+     :canned-response
```
davidangb, line 719: if PO's use case holds, and the schema changes regularly, this is going to be hard to maintain correctly.

in other news: do you actually use the canned-response/mock data feature?

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -327,16 +323,16 @@
...
             [:span {:style {:paddingRight "1ex"}}
-             (icons/font-icon {:style {:color (:exception-red style/colors)}}
-               :status-warning-triangle)]
+             (icons/icon {:style {:color (:exception-red style/colors)}}
+                         :warning-triangle)]
             (str "Error: " expected-msg)]
            [:div {:style {:textAlign "initial"}}
-            [:div {}
+            [:div {:style {:display "flex" :alignItems "center"}}
```
davidangb: there's a lot of `:display "flex" :alignItems "center"` copy-paste ... can this/should this be centralized? Maybe this is a question about inline styles vs. classes :)

src/cljs/org/broadinstitute/firecloud_ui/common/icons.cljs
```diff
@@ -77,5 +80,4 @@
    (fn [{:keys [props]}]
      [:span {:style {:display "inline-block" :position "relative" :verticalAlign "middle"
                      :height (:size props) :width (:size props)}}
-      (style/center {}
-        (font-icon {:style {:color "#fff" :fontSize (:size props)}} :status-warning))])})
+      (style/center {} [:span {:style {:color "white" :fontSize (:size props) :fontWeight "500"}} "!"])])})
```
davidangb, line 83: not a warning icon?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/details.cljs
```diff
@@ -73,6 +73,20 @@
     (= tab-string MONITOR) 4
     :else 0))
 
+(defn- process-workspace [raw-workspace]
+  (let [attributes (get-in raw-workspace ["workspace" "attributes"])
+        library-attributes (->> attributes
+                                (keep (fn [[k v]] (when (.startsWith k "library:") [(keyword (subs k 8)) v])))
```
davidangb: split on the ":" character instead of doing a substring based on length?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
```diff
@@ -73,17 +68,17 @@
       (when (and (not (:editing? @state))
               (empty? (:attrs-list @state))
               (not (:saving? @state)))
-        (style/create-paragraph [:em {} "There are no attributes to display"]))])])
+        [:em {} "There are no attributes to display"])])])
 
 (defn save-attributes [state props this description]
```
davidangb, line 73: hmmm ... might want a dedicated orch endpoint for this too, so the UI doesn't have to determine all the RemoveAttribute and AddUpdateAttribute ops itself.

src/static/index.html
```diff
@@ -13,6 +13,8 @@
           rel="stylesheet" type="text/css">
     <link href="assets/css/markdown.css"
           rel="stylesheet" type="text/css">
+    <link href="https://fonts.googleapis.com/icon?family=Material+Icons"
+          rel="stylesheet">
```
davidangb, line 17: before merging, let's talk internally to DSDE to see what we need to do for licensing.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -0,0 +1,155 @@
...
+    [org.broadinstitute.firecloud-ui.common.components :as comps]
+    [org.broadinstitute.firecloud-ui.common.input :as input]
+    [org.broadinstitute.firecloud-ui.common.modal :as modal]
+    [org.broadinstitute.firecloud-ui.common.style :as style]
+    [org.broadinstitute.firecloud-ui.utils :as utils]))
+
+
+(defn- calculate-display-properties [library-schema]
+  (filter (comp not :hidden #(get-in library-schema [:properties %]))
+          (:propertyOrder library-schema)))
```
davidangb: we could make it a business requirement that hidden properties are not included in propertyOrder, if that simplifies code.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -0,0 +1,155 @@
...
+                               (let [{:keys [hidden enum]} (get-in library-schema [:properties property-key])
+                                     pk-str (name property-key)
+                                     required? (contains? (:required-props @state) property-key)]
+                                 [property-key
+                                  {:required? required?
+                                   :value (not-empty
+                                            (cond hidden (resolve-hidden property-key (:workspace props))
+                                                  enum (common/get-text refs pk-str)
+                                                  required? (do (input/validate refs pk-str)
+                                                              (input/get-text refs pk-str))
```
davidangb, line 115: I'm slightly worried about any differences between common/get-text and input/get-text ... if a PO changes the `required` field in the schema, could the end result of the attribute value be different?

src/static/index.html
davidangb: update: talked to Kristian, linking at runtime should be fine, no action needed.

davidangb (commented):
as discussed in person, I don't really like merging the library-related code in a half-finished state. That said, I'm okay with all the syntax in those areas; the functionality will need work.

A few other specific comments inline.

src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
MattPutnam: Not often.  I stuck this here mainly as a reference while I was working out the UI.

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
MattPutnam: Yeah, a lot of these cases are just for sticking an icon next to text.  There could be a util for this.

src/cljs/org/broadinstitute/firecloud_ui/common/icons.cljs
MattPutnam: There wasn't a bare exclamation mark, but there are two versions of one in a circle.  Actually, one of them is used for `:error`.  We could use that.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/attribute_editor.cljs
MattPutnam: In my work for 1080 it's much simpler.  This logic is way overblown.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
MattPutnam: Either way, it doesn't really matter.  In other places I'm assuming propertyOrder is a full list of all properties, which may or may not be a good assumption.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
MattPutnam: I'm not following.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
davidangb: Unless we enforce it through tests/validation, that's probably not a good assumption (even though I believe it is true right now)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
davidangb: in other words - could common/get-text and input/get-text ever return different values, when passed the same refs/id? If they _could_ return different values, then if a PO changes any given field from required to optional, this code would calculate a different value.

common/get-text and input/get-text have duplicated code internally - but not complete copies of each other. Could they be more centralized?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
MattPutnam: Okay, I'll refactor.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
MattPutnam: They shouldn't return different values, they're just getting the text and trimming it.  I could have input/get-text use common/get-text internally.

This is for the input form, so this wouldn't change how the properties display.

src/cljs/org/broadinstitute/firecloud_ui/endpoints.cljs
dmohs: I don't love the idea of having to run another service, but it may be time to use Mock Server and move away from our own implementation.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -0,0 +1,155 @@
...
+       (if missing-fields
+         (swap! state assoc :validation-error ["Please fill out all required fields"])
+         (let [field-data-map (->> field-data
+                                   (keep (fn [[property-key {:keys [value]}]]
+                                           (when value
+                                             [(name property-key) value])))
+                                   (into {}))]
+           (swap! state dissoc :validation-error)
+           ;; TODO: update properties
+           (utils/cljslog field-data-map)))))})
```
dmohs, line 131: As I'm sure you know, these get annoying when you're working on something else and want zero noise while doing your own debugging. If this doesn't get triggered because the code is commented-out anyway, no problem.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/tab.cljs
```diff
@@ -60,70 +61,78 @@
...
-(defn- render-sidebar [state props refs this ws billing-projects owner? writer?]
-  (let [locked? (get-in ws ["workspace" "isLocked"])
+(defn- render-sidebar [state props refs this ws billing-projects owner? writer? curator?]
+  (let [locked? (get-in ws [:workspace :isLocked])
         status (common/compute-status ws)]
     [:div {:style {:flex "0 0 270px" :paddingRight 30}}
      [comps/StatusLabel {:text (str status
                                  (when (= status "Running")
-                                   (str " (" (get-in ws ["workspaceSubmissionStats" "runningSubmissionsCount"]) ")")))
+                                   (str " (" (get-in ws [:workspaceSubmissionStats :runningSubmissionsCount]) ")")))
```
dmohs, line 70: Not in scope for this, but something to think about. I feel like we want our own `js->clj` function that performs an "everything else" -> "idiomatic naming" translation. For example, `"fooBar"` becomes `:foo-bar`, `"i_hate_this_one"` becomes `:i-hate-this-one`, etc.

dmohs (approved):
In-person review was helpful. 👍

davidangb:
👍 only remaining issue is potentially using a graphical icon instead of the character "!" within ExceptionIcon - or explicitly saying you'd rather not.

When the library-related code gets turned back on and completed, let's do a full code review of that, even if the code hasn't changed from this PR.

PR #525, opened 2016-09-27
GAWB-640: Save workspace table state

MattPutnam:
Added an optional property `:state-key` to the table.  When present, the table saves its state to local storage whenever it changes, and tries to load it again on creation.  This required a restructuring of the table code to not persist the sorting function, since you can't serialize/deserialize functions this way.  So please double-check that sorting still works correctly.

Main now also stores the user's subjectId as part of generating unique but consistent keys.

To implement this for other tables, it should be as easy as specifying a path-unique key for a table.  For example, "[workspaceNamespace]:[workspaceName]:monitor-tab".
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
@@ -355,7 +355,7 @@
        (case (count crumbs)
          0 [:div {}]
          1 [:div {} (:text (first crumbs))]
-         [:div {:style {:display "flex" :alignItems "center"}}
+         [:div {:style {:display "flex" :alignItems "center" :flexWrap "wrap"}}
           (interpose sep
             (map
               (fn [{:keys [text onClick href] :as link-props}]
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
@@ -288,10 +288,9 @@
                rows (map ->row filtered-data)
                sorted-rows (if sort-column
                              (let [column (nth (:columns props) sort-column)
-                                   key-fn (or (:sort-by column) identity)]
-                               (if (= key-fn :text)
-                                 (sort-by (fn [row] ((:as-text column) (nth row sort-column))) rows)
-                                 (sort-by (fn [row] (key-fn (nth row sort-column))) rows)))
+                                   key-fn (or (:sort-by column) identity)
+                                   key-fn (if (= key-fn :text) (:as-text column) key-fn)]
+                               (sort-by (fn [row] (key-fn (nth row sort-column))) rows))
                              rows)
                ordered-rows (if (= :desc sort-order) (reverse sorted-rows) sorted-rows)
                ;; realize this sequence so errors can be caught early:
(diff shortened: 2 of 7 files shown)


MattPutnam:
The latest commit generalizes persistence, and hopefully shows that it's pretty easy to apply to any component.  I also included the ability to exclude keys if necessary (for transient things).

src/cljs/org/broadinstitute/firecloud_ui/common/components.cljs
```diff
@@ -355,7 +355,7 @@
        (case (count crumbs)
          0 [:div {}]
          1 [:div {} (:text (first crumbs))]
-         [:div {:style {:display "flex" :alignItems "center"}}
+         [:div {:style {:display "flex" :alignItems "center" :flexWrap "wrap"}}
```
MattPutnam, line 358: Shoehorning this in because I noticed it looking dumb

src/cljs/org/broadinstitute/firecloud_ui/persistence.cljs
```diff
@@ -0,0 +1,16 @@
+(ns org.broadinstitute.firecloud-ui.persistence
+  (:require
+    [org.broadinstitute.firecloud-ui.utils :as utils]
+    ))
+
+
+(defn- generate-persistence-key [key]
+  (keyword (str "state:" @utils/current-user ":" key)))
```
MattPutnam, line 8: I decided on this format because I want to be able to tell what's what.

src/cljs/org/broadinstitute/firecloud_ui/common/table.cljs
```diff
@@ -284,7 +286,13 @@
                                (table-utils/filter-data grouped-data ->row (:columns props) txt)
                                grouped-data)
                rows (map ->row filtered-data)
-               sorted-rows (if key-fn (sort-by key-fn rows) rows)
+               sorted-rows (if sort-column
+                             (let [column (nth (:columns props) sort-column)
+                                   key-fn (or (:sort-by column) identity)]
+                               (if (= key-fn :text)
+                                 (sort-by (fn [row] ((:as-text column) (nth row sort-column))) rows)
+                                 (sort-by (fn [row] (key-fn (nth row sort-column))) rows)))
```
davidangb: these last three lines say "if key-fn is 'text', let key-fn be (:as-text column)", right? Could this be rewritten as such, so it resolves `key-fn`, then calls `sort-by`, and eliminates the duplication in the last two lines?

davidangb (commented):
Code looks good overall. Once minor syntax comment inline.

Functionally, this works. But ... if you look at the value that is persisted to local storage, it includes the `:display-rows` value. Is this necessary? Could that be excluded? I'd like to exclude it for two reasons:
1) it takes up the majority of the packet. In the extreme case where we persist a whole lot of local storage - for instance, if we enable this for all tables inside the workspace - we could eventually hit the localstorage limit. Removing this bulk would help avoid that.
2) It's slightly leaky. Even though we're not leaking any genomic data or metadata, we leak workspace names, and this persists while the user is not logged in. A malicious user could read this info from the browser without any kind of sign-in.

skwalker:
The code looks fine from my very basic understanding -- so 👍 pending David An's suggested fix
.

davidangb:
:+1: works for me!

PR #527, opened 2016-10-11
serve ui with figwheel

zarsky-broad:
JIRA Issue: N/A

Corresponds to https://github.com/broadinstitute/firecloud-develop/pull/156.

@dmohs and I had a conversation, and it occurred to us that it was confusing and unnecessary to be serving the UI app with apache and hot reloading with figwheel when figwheel has a server built-in.
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -18,7 +18,6 @@
 (defn- get-figwheel-server-opts []
   (when (inside-container?)
     {:http-server-root ""
-    :server-hostname "local.broadinstitute.org"
     :server-port 80
     :hawk-options {:watcher :polling}}))
 
--- a/project.clj
+++ b/project.clj
@@ -12,12 +12,15 @@
                       "*** You did not specify a FIGWHEEL_HOST environment variable.\n"
                       "*** Using the default: " host "\n"
                       "***")))
-      {:websocket-url (str "ws://" host ":3449/figwheel-ws")})))
+      {:websocket-url (str "ws://" host ":80/figwheel-ws")})))
 
 
 (defn- get-figwheel-server-opts []
   (when (inside-container?)
-    {:hawk-options {:watcher :polling}}))
+    {:http-server-root ""
+    :server-hostname "local.broadinstitute.org"
+    :server-port 80
+    :hawk-options {:watcher :polling}}))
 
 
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -12,12 +12,14 @@
...
-      {:websocket-url (str "ws://" host ":3449/figwheel-ws")})))
+      {:websocket-url (str "ws://" host ":80/figwheel-ws")})))
 
 
 (defn- get-figwheel-server-opts []
   (when (inside-container?)
-    {:hawk-options {:watcher :polling}}))
+    {:http-server-root ""
+    :server-port 80
+    :hawk-options {:watcher :polling}}))
```
MattPutnam: nit: indentation

MattPutnam:
👍 seems legit

project.clj
zarsky-broad: Fixed.

davidangb:
I'm open to debate on this, but I disagree with the principle. Because we serve the UI via apache in prod (and dev, alpha, etc), I think we should serve the UI via apache in our dev envs. This ensures that during development, we see the exact same behavior that our end users will see, including various apache configs.

The counter-argument is that we want to make our dev envs as frictionless/lightweight as possible, so we accept the potential differences. There are other ways to run via apache in dev envs.

Does this (coupled with broadinstitute/firecloud-develop#156) change any behavior when running via the full-stack firecloud-docker-compose.yaml? I can't imagine it does ...

dmohs:
@davidangb You raise a good point and it spurred a discussion. I'll summarize our thoughts as best I can.

There is value both in making development lightweight and in ensuring we are committing code that is as likely as possible to work in all of our deployment environments.

How about we say that developers should generally do a final round of testing by building the container before deploying so they can catch issues related to how the software will work in the deployed environment?

davidangb:
sounds good to me. As long as the full-stack docker-compose still serves via container, I think we have a good set of options.

If it turns out that UI-served-by-figwheel and UI-served-by-apache have nontrivial differences, then we can always roll this back.

PR #532, opened 2016-10-18
Iz state save gawb 1147

zarsky-broad:
  - Authentication
  - Authorization
  - Encryption
  - Audit trails
- Review cycle:
  - LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/persistence.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/persistence.cljs
@@ -8,7 +8,11 @@
   (keyword (str "state:" @utils/current-user ":" key)))
 
 (defn save [{:keys [key state except only]}]
-  (utils/local-storage-write (generate-persistence-key key) (apply dissoc (if only (select-keys @state only) @state) except)))
+  (assert (not (and except only)) "Specify EITHER except OR only")
+  (utils/local-storage-write (generate-persistence-key key)
+    (cond except (apply dissoc @state except)
+          only (select-keys @state only)
+          :else @state)))
 
 (defn try-restore [{:keys [key initial]}]
   (if-let [saved-state (some-> key generate-persistence-key utils/local-storage-read cljs.reader/read-string)]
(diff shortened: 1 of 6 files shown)


src/cljs/org/broadinstitute/firecloud_ui/persistence.cljs
```diff
@@ -7,8 +7,8 @@
 (defn- generate-persistence-key [key]
   (keyword (str "state:" @utils/current-user ":" key)))
 
-(defn save [{:keys [key state except]}]
-  (utils/local-storage-write (generate-persistence-key key) (apply dissoc @state except)))
+(defn save [{:keys [key state except only]}]
+  (utils/local-storage-write (generate-persistence-key key) (apply dissoc (if only (select-keys @state only) @state) except)))
```
MattPutnam: Specifying both `:except` and `:only` seems nonsensical, so it might be good to guard against that with an assert (or a precondition?).  And we could clean up this line a bit too, something like:

``` clojure
(utils/local-storage-write (generate-persistence-key key)
  (cond except (apply dissoc @state except)
        only (select-keys @state only)
        :else @state))
```

MattPutnam (approved):
👍 with one suggestion

davidangb:
Code looks good syntax-wise. I noticed two functional problems:
- if I change rows-per page, my choice is persisted when I reload - but the `<select>` control doesn't default correctly and will show the `<option>` for 10.
- my choice of entity type in the top button bar isn't persisted. I'd like to be able to return to the Data tab and default to seeing participants, for instance. I realize this isn't specified as an explicit requirement in the Jira ticket, but it feels like it belongs. Thoughts?

zarsky-broad:
@davidangb First comment should be addressed now. As to the second, I agree that it's a desirable behavior, but is it far enough removed from the ticket that it should get its own? I'm still learning the DSDE preference on these things.

davidangb:
:+1: works for me! Just need the PO input on the entity-type selection, but otherwise good.

zarsky-broad:
@davidangb Tiffany says that would be out of scope for this ticket.

MattPutnam:
👍

PR #534, opened 2016-10-19
Figwheel always on 3449

dmohs:
This change causes Figwheel to always run on its default port, 3449. We use docker flags (in our scripts) to choose which port to use on the host.


jmthibault79:
Hot reload and `build-once` both work for me in the local Docker stack environment.  Thanks!

Update the README to use the correct hot-reload script.

Are the other scripts in `/script` still valid?

davidangb:
also needs a rebase.

davidangb:
+1 to Joel's comments. When you update the README, also remove reference to FIGWHEEL_HOST.

dmohs:
@jmthibault79 Thanks for mentioning the other scripts. I made a commit that makes them more robust, which I had meant to do but forgot.

@jmthibault79 @davidangb README updated.

jmthibault79:
👍

davidangb (approved):
👍

PR #548, opened 2016-10-31
Add a logout button and refactor authentication

dmohs:
Requires: https://github.com/broadinstitute/firecloud-develop/pull/170 and https://github.com/broadinstitute/firecloud-orchestration/pull/292

This new flow uses Google's JavaScript library to sign-in and acquire an access token. This process also gives a code that can be exchanged for a refresh token, so this is what is sent to orchestration to complete the process.

Since the refresh token can become stale or otherwise lost, the UI sends a request to orchestration after a successful sign-in to check the refresh token. If it is expired, it gives the user a chance to update it.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -327,66 +336,130 @@
...
+   (fn [{:keys [props state]}]
+     (utils/ajax-orch "/me"
+                      {:on-done (fn [{:keys [success? status-code]}]
+                                  (if success?
+                                    ((:on-success props))
+                                    (case status-code
+                                      403 (swap! state assoc :error :not-active)
+                                      ;; 404 means "not yet registered"
+                                      404 ((:on-success props))
+                                      (swap! state assoc :error true))))}
```
dmohs, line 389: Pretty sure this is wrong. Stand by.

dmohs: Nope this is right. Basically, if successful or 404, proceed, since that either means the user is good to go or they are not yet registered. Suggestions for improving clarity welcome.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -275,35 +276,43 @@
 
 (react/defc LoggedOut
   {:render
-   (fn [{:keys [props]}]
-     [:div {}
+   (fn [{:keys [this props]}]
+     ;; Google's code complains if the sign-in button goes missing, so we hide this component rather
+     ;; than removing it from the page.
+     [:div {:style {:display (when (:hidden? props) "none")}}
```
davidangb, line 297: is this comment still true?

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -275,35 +276,43 @@
...
-               :headers {"Authorization" (str "Bearer " token)}
-               :on-done on-done}))
+       [Policy]]])
+   :component-did-mount
+   (fn [{:keys [props locals]}]
+     (swap! locals assoc :refresh-token-saved? true)
+     (let [{:keys [auth2 on-change]} props]
+       (-> auth2 (.-currentUser)
+           (.listen (fn [u]
+                      (on-change (.isSignedIn u) (:refresh-token-saved? @locals)))))))
```
davidangb, line 313: can we use this hook to detect when the user's access token has expired, instead of the awkward detection of 40x inside of our ajax function(s)? (would certainly be a separate story/PR, don't want to scope-creep this one)

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
```diff
@@ -321,66 +330,130 @@
...
+   :component-did-mount
+   (fn [{:keys [props state]}]
+     ;; Use basic ajax call here to bypass authentication.
+     (utils/ajax {:url "/config.json"
+                  :on-done (fn [{:keys [success? get-parsed-response]}]
+                             (if success?
+                               (do
+                                 (reset! config/config (get-parsed-response false))
+                                 ((:on-success props)))
+                               (swap! state assoc :error true)))}))})
```
davidangb, line 365: an enhancement: have the ConfigLoader component also check the contents of config.json to ensure all required keys are present (as defined in config.cljs). This enhancement would handle the case, for instance, where somebody is using this branch of firecloud-ui but hasn't incorporated the associated PR from firecloud-develop.

src/cljs/org/broadinstitute/firecloud_ui/page/billing/create_project.cljs
```diff
@@ -18,20 +17,14 @@
...
-                                                       "Authentication"
-                                                       "menubar=no,toolbar=no,width=500,height=500"))}]
-                    [comps/ErrorViewer {:error billing-acct-error}])
-                  (catch js/Object _ [comps/ErrorViewer {:error billing-acct-error}]))
+                (if (= billing-acct-error 403)
+                  [:div {:style {:textAlign "center"}}
+                   "Billing permissions are not enabled." [:br] [:br]
+                   [comps/Button {:text "Enable Billing Permissions"
+                                  :onClick #(react/call :.enable-billing-permissions this)}]]
+                  [comps/ErrorViewer {:error (get-error-details)}])
```
davidangb: how do you differentiate the 403 of scope-not-enabled vs. the 403 when the oauth scope is enabled, but the user doesn't own the relevant billing project?

src/cljs/org/broadinstitute/firecloud_ui/page/billing/create_project.cljs
```diff
@@ -70,29 +63,31 @@
...
+   (fn [{:keys [this]}]
+     (-> @utils/google-auth2-instance
+         (.grantOfflineAccess (clj->js {:redirect_uri "postmessage"
+                                        :scope "https://www.googleapis.com/auth/cloud-billing"}))
+         (.then (fn [response]
+                  ;; At some point, the access token gets replaced with a new one, but it doesn't
+                  ;; happen right away, so this dummy delay is necessary.
+                  (js/setTimeout
+                   #(react/call :get-billing-accounts this)
+                   1000)))))
```
davidangb: 1) by granting offline access to a new scope, oauth is generating a new refresh token, right? Does that mean the refresh token previously saved for the user is no longer valid, in which case analyses could fail?

2) where does the delay happen? Is there a more graceful way to handle this than a hardcoded 1-second sleep?

src/cljs/org/broadinstitute/firecloud_ui/common/gcs_file_preview.cljs
```diff
@@ -34,6 +34,7 @@
                       (common/format-filesize data-size)
                       [:span {:style {:marginLeft "1em"}}
                        [:a {:href (common/gcs-object->download-url (:bucket-name props) (:object props))
+                            :onClick #(utils/set-access-token-cookie (utils/get-access-token))
```
davidangb, line 37: moving the cookie-writing here has a very minor implication: previously, developers/users could log in to FireCloud via the UI (to get the cookie), and then manually hit the orchestration-layer cookie-authed endpoints in their browser - say, via swagger - to Do Stuff. This will no longer work - they need to click a download url *in the UI* to get the cookie now.

I don't object to this - but we may see developers get tripped up and need to be aware. I have no idea if end users ever rely on the old behavior.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
```diff
@@ -136,11 +135,12 @@
...
-                    :disabled? (:disabled? props)
-                    :onClick #(if (:force-login? props)
-                               (sign-in/show-sign-in-dialog :refresh-token (:after-login props))
-                               (modal/push-modal [Form (select-keys props [:config-id :workspace-id :root-entity-type :on-success :cromwell-version])]))}])})
+     [comps/Button
+      {:text "Launch Analysis..."
+       :disabled? (:disabled? props)
+       :onClick (modal/push-modal
+                 [Form (select-keys props [:config-id :workspace-id
+                                           :root-entity-type :on-success :cromwell-version])])}])})
```
davidangb, line 143: why was the :force-login option removed here?

My understanding is this option handled the (corner) case where the user's refresh token expired, based on our definition of expire, within the user's current session. Before launching an analysis, we check the refresh token to see if it's still valid, and if not, we go get a new one. In this PR, we only check the refresh token date upon login, correct?

davidangb (commented):
comments inline, back to you @dmohs

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Yes. I didn't copy this—I had to re-add it :(

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Google's code regenerates the access token automatically, so there is no expiration anymore.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: That is a fantastic idea. To keep this PR clean, I'll do that as a follow-up (GAWB-1334).

src/cljs/org/broadinstitute/firecloud_ui/page/billing/create_project.cljs
dmohs: I believe this is a request for **all** billing accounts, not a specific one. As far as I know, I didn't change the behavior of this, only the mechanics.

src/cljs/org/broadinstitute/firecloud_ui/page/billing/create_project.cljs
dmohs: 1. According to [this documentation](https://developers.google.com/identity/protocols/OAuth2WebServer#incrementalAuth), adding the new scope does not invalidate the existing tokens. I don't believe it is generating a new token. It is giving me a code that could be used for that, but I don't see a reason to re-create the refresh token to include the new scope.
2. Google issues a request perhaps to update its token, but it appears that the `then` is fired before that request comes back. I might be able to save the previous value of the access token and loop until it has changed. I'll give that a try.

src/cljs/org/broadinstitute/firecloud_ui/common/gcs_file_preview.cljs
dmohs: Thank you. Good to keep that in mind.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
dmohs: We check whenever the page is first loaded as well as on login. I can't imagine needing to do the check here.

src/cljs/org/broadinstitute/firecloud_ui/page/billing/create_project.cljs
dmohs: Implemented a loop which should be much more reliable.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
davidangb: relevant issue is GAWB-798.

dmohs: Thank you! Reading that, I believe it would be impossible for the user to launch a job without seeing the refresh token warning.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
davidangb: ✅

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
davidangb: that's a huge - and positive - change! Needs to get documented somewhere, at minimum so QA understands.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
davidangb: have you tried the repro case from 798 against this code? just from reading it, I think this PR makes us vulnerable again.

src/cljs/org/broadinstitute/firecloud_ui/main.cljs
dmohs: Added comments to "QA Requirements" section in the ticket per Gary.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/method_configs/launch_analysis.cljs
dmohs: To go from the "account is inactive" screen to anything else, you would have no choice but to refresh the page. That action would cause a status check on the refresh token, so you couldn't possibly launch a job without seeing the "Your offline credentials are missing or out-of-date." warning.

davidangb:
👍 approve!

PR #555, opened 2016-11-04
Dm gawb 544 review

dmohs:


jlmwhitt:
I updated the format as suggested, but left the version in the modal near the launch button, as requested.

PR #562, opened 2016-11-09
Install devtools correctly

dmohs:
ClojureScript supports preloads in 1.9. We're on 1.7. This change also adds the "hints" feature which is very helpful for "null" and "undefined" error messages.


zarsky-broad (approved):
LGTM

PR #573, opened 2016-11-18
GAWB-1236: Prevent single-owner acl from removing only owner permissions

rushtong:
PR for https://broadinstitute.atlassian.net/browse/GAWB-1236

This addresses, but does not complete GAWB-1236
Please see https://github.com/broadinstitute/agora/pull/171 for the Agora side of this PR. These can be tested separately.

Add a check for deletion of permissions to prevent removing all managers and therefore leaving the entity in an un-manageable state.

Looking for feedback on the UI handling. The one thing I don't like is the following condition: 
* if there is a single owner ACL, then that ACL is appropriately disabled
* upon adding new ACLs, or changing ACLs, that disabled state should toggle off if another owner is added.

---
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/methods_configs_acl.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/methods_configs_acl.cljs
@@ -47,7 +47,7 @@
            (map-indexed
             (fn [i acl-entry]
               (let [disabled? (and (< (count-owners(:acl-vec @state)) 2) (= owner-level (:role acl-entry)))
-                    background (if disabled? (:disabled-state style/colors) (:input-background style/colors))]
+                    background-color (if disabled? (:disabled-state style/colors) (:input-background style/colors))]
                 [:div {}
                  [input/TextField
                   {:ref (str "acl-key" i)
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/method_repo/methods_configs_acl.cljs
```diff
@@ -25,6 +25,8 @@
 
 (def ^:private column-width "calc(50% - 4px)")
 
+(defn count-owners [acl-vec]
+  (count (filter #(= (:role %) owner-level) acl-vec)))
```
MattPutnam: Another option:

```
(defn count-owners [acl-vec]
  (count (filter (comp #{owner-level} :role) acl-vec)))
```

MattPutnam (commented):
Ugh... there are a bunch of things that are annoying me in this file.  Let me make a cleanup branch off this.

MattPutnam:
👍 
Up to you to decide if the thing I suggested above is too arcane or not.

rushtong:
It makes sense - I'll push an update out.

davidangb:
There's a clickpath where the UI protection doesn't work:

1. Start editing an ACL
2. Add a new row. Leave the new row's user empty, but set the new row to OWNER.
3. Remove all previous owners.
4. Save.

This clickpath should still be protected on the API side, so the data is safe; but if we're going to add UI protection it shouldn't be this easy to defeat. I also don't know how to get around this ... it would be just as easy to add a new row and fill in a dummy email address, and without sending to the back end we can't know if that email address is valid or not.

rushtong:
With the new requirements, there is no need to update the UI. Current errors from the API are correctly percolated up to the editing modal.

PR #577, opened 2016-11-18
GAWB-1126

skwalker:
Orch PR: https://github.com/broadinstitute/firecloud-orchestration/pull/313

I think that select2 matches our UI better than the twitter typeahead, but both use jquery, and Isaac had used select2 before and suggested using that, which is why I did. I don't think there's anything that twitter typeahead could give us that is missing from select2, but if we think we should switch to that, that's okay.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -177,4 +177,5 @@
           :else 1)))
 
 (defn url-to-doid [url]
-   (str "DOID: " (second (clojure.string/split url #"DOID_"))))
+  ;; TODO when we support multiple ontologies, we'll need to look for some string other than "DOID"
+  (str "DOID: " (second (clojure.string/split url #"DOID_"))))
--- a/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common.cljs
@@ -135,7 +135,7 @@
   (let [unparsed-values (get unparsed-profile "keyValuePairs")]
     (into {} (map (fn [m] [(keyword (m "key")) (m "value")]) unparsed-values))))
 
-(defn workspace-id->string [workspace-id] 
+(defn workspace-id->string [workspace-id]
   (str (:namespace workspace-id) "/" (:name workspace-id)))
 
 (defn get-id-from-nav-segment [segment]
@@ -175,3 +175,6 @@
           (count (get-in entity ["attributes" (set-type->membership-attribute entity-type)]))
           ;; something nonsensical has been selected, submission will probably fail anyway:
           :else 1)))
+
+(defn url-to-doid [url]
+   (str "DOID: " (second (clojure.string/split url #"DOID_"))))
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
@@ -126,4 +126,6 @@
 (defn render-entity [namespace name snapshot-id]
   (render-name-id (str namespace "/" name) snapshot-id))
 
+(defn url-to-doid [url]
+  (second (clojure.string/split url #"DOID_")))
 
--- a/src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
@@ -126,6 +126,3 @@
 (defn render-entity [namespace name snapshot-id]
   (render-name-id (str namespace "/" name) snapshot-id))
 
-(defn url-to-doid [url]
-  (second (clojure.string/split url #"DOID_")))
-
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
@@ -6,6 +6,7 @@
     [org.broadinstitute.firecloud-ui.common.input :as input]
     [org.broadinstitute.firecloud-ui.common.modal :as modal]
     [org.broadinstitute.firecloud-ui.common.style :as style]
+    [org.broadinstitute.firecloud-ui.config :as config]
     [org.broadinstitute.firecloud-ui.endpoints :as endpoints]
     [org.broadinstitute.firecloud-ui.utils :as utils]
     ))
@@ -59,6 +60,33 @@
                       parsed))))
     (do (utils/log "unknown type: " type)
         value)))
+
+(defn select2-typeahead [placeholder r on-item-selected]
+  (.select2 (js/$ r)
+            (clj->js { :ajax { :url (fn [params]
+                                      (str (config/api-url-root) "/duos-autocomplete/" (aget params "term")))
+                               :dataType "json"
+                               :delay 250
+                               :processResults (fn [data]
+                                                 (clj->js {:results data}))
+                               :data (fn [params] ;; don't actually want to send it as a query
+                                       (clj->js {}))}
+                       :minimumInputLength 1
+                       :cache true
+                       :templateResult (fn [obj]
+                                         (if (aget obj "loading")
+                                           (aget obj "text")
+                                           (str "<div style='font-weight: bold'>" (aget obj "label")
+                                                "<div style='float: right; font-weight: normal'> DOID: "
+                                                (style/url-to-doid (aget obj "id")) "</div></div>"
+                                                "<div style='font-size: small'> "
+                                                (if (not (nil? (aget obj "definition")))
+                                                  (aget obj "definition") "</div>"))))
+                       :escapeMarkup (fn [markup] markup)
+                       :templateSelection (fn [obj]
+                                            (on-item-selected (aget obj "id") (aget obj "label"))
+                                            (aget obj "label"))
+                       :placeholder (clj->js placeholder)})))
 
 (react/defc LibraryAttributeForm
   {:get-initial-state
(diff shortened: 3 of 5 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -96,18 +126,20 @@
...
-                                         :predicates [(when required?
-                                                        (input/nonempty pk-str))
-                                                      (when (= type "integer")
-                                                        (input/integer pk-str :min minimum))]}])])))
+                     (cond
+                       enum (style/create-identity-select {:ref pk-str
+                                                             :defaultValue (get existing property-key ENUM_EMPTY_CHOICE)}
+                                                            (cons ENUM_EMPTY_CHOICE enum))
+                       (= typeahead "ontology") [:select {:ref "duos-typeahead"
+                                                          :style {:width "100%"}}]
```
davidangb: what happens if there are two attributes with typeahead? They'd both be looking for the ref `duos-typeahead`.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -25,18 +26,20 @@
...
 (defn- render-property [library-schema library-attributes property-key]
   [:div {:style {:display "flex" :padding "0.5em 0" :borderBottom (str "2px solid " (:line-default style/colors))}}
-   [:div {:style {:flexBasis "33%" :fontWeight "bold" :paddingRight "2em"}}
+   [:div {:style {:flexBasis "40%" :fontWeight "bold" :paddingRight "2em"}}
     (get-in library-schema [:properties property-key :title])]
-   [:div {:style {:flexBasis "67%"}}
-    (render-value (get library-attributes property-key))]])
+   [:div {:style {:flexBasis "60%"}}
+    (render-value (get library-attributes property-key))
+    (if (= property-key :library:diseaseOntology) [:span {:style {:fontStyle "italic"}} (str " (DOID: " (style/url-to-doid (get library-attributes :library:diseaseOntologyDOID)) ")")])]])
```
davidangb: this shouldn't look specifically for a property named "diseaseOntology" - it should be looking for any property enabled for ontology typeahead. Similarly, once we use another ontology, the "DOID_" prefix will be invalid.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -137,15 +170,22 @@
...
+                          (fn [id label]
+                            (swap! state assoc :ontologyText label :ontologyDOID id)))))})
 
 (react/defc LibraryAttributeViewer
   {:render
    (fn [{:keys [props state]}]
      (let [{:keys [library-attributes library-schema]} props
            form-properties (select-keys props [:library-schema :workspace :workspace-id :request-refresh])
-           primary-properties [:library:indication :library:numSubjects :library:datatype :library:dataUseRestriction] ;; TODO replace with new field from schema
+           primary-properties [:library:numSubjects :library:datatype :library:dataUseRestriction :library:diseaseOntology] ;; TODO replace with new field from schema
```
davidangb: arggg these really should be in the schema, not cljs (see GAWB-1234)

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -58,6 +61,33 @@
...
+                       :cache true
+                       :templateResult (fn [obj]
+                                         (if (aget obj "loading")
+                                           (aget obj "text")
+                                           (str "<div style='font-weight: bold'>" (aget obj "label")
+                                                "<div style='float: right; font-weight: normal'> DOID: "
+                                                (style/url-to-doid (aget obj "id")) "</div></div>"
+                                                "<div style='font-size: small'> "
+                                                (if (not (nil? (aget obj "definition")))
+                                                  (aget obj "definition") "</div>"))))
```
davidangb: this might be the only place we're using HTML directly, instead of cljs. Could this be refactored into cljs?

davidangb (changes requested):
I have a few comments inline, but my biggest concern is that this code is written very specifically for the diseaseOntology attribute, the DOID ontology, and will only work with one instance of typehead. This should be made more generic and schema-driven instead of tightly coupled to that specific property.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -58,6 +61,33 @@
...
+                       :escapeMarkup (fn [markup] markup)
```
MattPutnam: `(fn [markup] markup)` -> `identity`

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -25,18 +26,20 @@
...
 
-(defn- resolve-hidden [property-key workspace]
+(defn- resolve-hidden [property-key workspace doid]
   (case property-key
     :workspaceId (get-in workspace [:workspace :workspaceId])
     :workspaceNamespace (get-in workspace [:workspace :namespace])
     :workspaceName (get-in workspace [:workspace :name])
-    nil))
+    nil)
+  :diseaseOntologyDOID doid)
```
MattPutnam: This needs to be inside the `case`, otherwise you're dropping everything on the floor and just returning `doid` no matter what.

src/cljs/org/broadinstitute/firecloud_ui/common/style.cljs
```diff
@@ -126,4 +126,6 @@
 (defn render-entity [namespace name snapshot-id]
   (render-name-id (str namespace "/" name) snapshot-id))
 
+(defn url-to-doid [url]
+  (second (clojure.string/split url #"DOID_")))
```
MattPutnam: This shouldn't be in style, maybe common?

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
MattPutnam: I'll end up taking care of that in GAWB-1130.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
skwalker: Uhh I don't think that's possible... The select2 doesn't know how to go from cljs to html and I know this is kinda gross but I'm not sure how to change it..

src/cljs/org/broadinstitute/firecloud_ui/common.cljs
```diff
@@ -175,3 +175,6 @@
           (count (get-in entity ["attributes" (set-type->membership-attribute entity-type)]))
           ;; something nonsensical has been selected, submission will probably fail anyway:
           :else 1)))
+
+(defn url-to-doid [url]
+   (str "DOID: " (second (clojure.string/split url #"DOID_"))))
```
davidangb: add a TODO comment: when we support multiple ontologies, we'll need to look for some string other than "DOID".

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -25,17 +26,21 @@
...
-   [:div {:style {:flexBasis "33%" :fontWeight "bold" :paddingRight "2em"}}
+   [:div {:style {:flexBasis "40%" :fontWeight "bold" :paddingRight "2em"}}
     (get-in library-schema [:properties property-key :title])]
-   [:div {:style {:flexBasis "67%"}}
-    (render-value (get library-attributes property-key))]])
+   [:div {:style {:flexBasis "60%"}}
+    (render-value (get library-attributes property-key))
+    (if (= (get-in library-schema [:properties property-key :typeahead]) "ontology")
+      ;; check if it's a doid ontology?
+      [:span {:style {:fontStyle "italic"}} (str " (" (common/url-to-doid (get library-attributes :library:diseaseOntologyDOID)) ")")])]])
```
davidangb: add a TODO comment: when we support multiple ontology-driven properties, can't hardcode the property name here.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -58,6 +63,32 @@
     (do (utils/log "unknown type: " type)
         value)))
 
+(defn select2-typeahead [placeholder r on-item-selected]
+  (.select2 (js/$ r)
+            (clj->js { :ajax { :url (fn [params]
+                                      (str (config/api-url-root) "/duos/autocomplete/" (aget params "term")))
```
davidangb, line 71: add a TODO comment: when we support multiple ontologies, may need a different endpoint.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -116,10 +149,11 @@
      (swap! state dissoc :validation-error :server-error)
      (let [validation-errors (atom [])
            field-data (->> props :library-schema :properties
-                           (keep (fn [[property-key {:keys [hidden enum] :as property}]]
+                           (keep (fn [[property-key {:keys [hidden enum typeahead] :as property}]]
                                    (let [pk-str (name property-key)
-                                         value (cond hidden (resolve-hidden property-key (:workspace props))
+                                         value (cond hidden (resolve-hidden property-key (:workspace props) (:ontologyDOID @state))
                                                      enum (resolve-enum (common/get-text refs pk-str))
+                                                     (= typeahead "ontology") (:ontologyText @state)
```
davidangb: add a TODO comment: when we support multiple ontology-driven properties, can't rely on hardcoded values in state.

src/cljs/org/broadinstitute/firecloud_ui/page/workspace/summary/library.cljs
```diff
@@ -137,15 +171,27 @@
...
+       [attributes (get-in props [:workspace :workspace :library-attributes])
+        properties (get-in props [:library-schema :properties])]
+       (doseq [[property-key {:keys [typeahead] :as property}] properties]
+                (if (= typeahead "ontology")
+                  (let [pk-str (name property-key)]
+                    (select2-typeahead {:id  (get attributes :library:diseaseOntologyDOID -1 )
+                                        :label (get attributes :library:diseaseOntology "Select a disease ontology.")}
+                                       (@refs pk-str)
+                                       (fn [id label]
+                                         (swap! state assoc :ontologyText label :ontologyDOID id))))))))})
```
davidangb: add a TODO comment: when we support multiple ontology-driven properties, can't hardcode the keys in state or attributes

davidangb (commented):
PR works for me. I'm requesting addition of a few TODO comments in relevant places of the code, to make it easier later. Also, anywhere we're not talking specifically about the phenotype attribute specifically, try to use generic variable names like "id" instead of something like "doid".

Other than those - hopefully quick - changes, 👍

davidangb:
👍 haven't tested functionally but code looks correct.

PR #594, opened 2016-12-06
GAWB-1302: split workspace list filter into access/realm conditions

davidangb:
For GAWB-1302.

This does *not* handle the acceptance criteria in GAWB-1302 to change the message shown to the user when clicking on a dataset in Library search results. I will handle that in a separate PR.

This PR splits the workspace list's access level filter into two sections: permissions and realm. We operate with OR inside each section, but AND between the two sections.

We also add a new realm option "non-TCGA". There's extra code in here to handle defaulting that option to true/on/checked, in case the user has nothing in local storage (since all users will have nothing in local storage, because it's a new option)

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
@@ -116,13 +116,14 @@
        [:div {}
         (when (:show-access-level-select? @state)
           (let [checkbox (fn [label]
-                           [:div {}
-                            [:label {:style {:cursor "pointer"}}
-                             [:input {:type "checkbox"
-                                      :checked (get (:selected-types @state) label)
-                                      :onChange #(swap! state update-in [:selected-types label] not)
-                                      :style {:cursor "pointer"}}]
-                             [:span {:style {:marginLeft "0.5ex"}} label]]])]
+                           (let [stateval (get (:selected-types @state) label)]
+                             [:div {}
+                              [:label {:style {:cursor "pointer"}}
+                               [:input {:type "checkbox"
+                                        :checked (or (nil? stateval) stateval)
+                                        :onChange #(swap! state update-in [:selected-types label] not)
+                                        :style {:cursor "pointer"}}]
+                               [:span {:style {:marginLeft "0.5ex"}} label]]]))]
             [overlay/Overlay {:get-anchor-dom-node #(react/find-dom-node (@refs "anchor"))
                               :anchor-x :right :anchor-y :bottom
                               :dismiss-self #(swap! state dissoc :show-access-level-select?)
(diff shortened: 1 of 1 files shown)


src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -109,20 +115,23 @@
...
-                             [:input {:type "checkbox"
-                                      :checked (get (:selected-types @state) label)
-                                      :onChange #(swap! state update-in [:selected-types label] not)
-                                      :style {:cursor "pointer"}}]
-                             [:span {:style {:marginLeft "0.5ex"}} label]]])]
+                           (let [stateval (get (:selected-types @state) label)]
+                             [:div {}
+                              [:label {:style {:cursor "pointer"}}
+                               [:input {:type "checkbox"
+                                        :checked (or (nil? stateval) stateval)
```
davidangb: if state does not have a value, default to `checked="true"`

MattPutnam: Alternate: `(not (false? stateval))`

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
```diff
@@ -109,20 +115,23 @@
...
+                                        :onChange #(swap! state update-in [:selected-types label] false?)
```
MattPutnam, line 123: using `not` to flip the value might be clearer

MattPutnam (approved):
👍 With two small suggestions

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
davidangb: good call!

src/cljs/org/broadinstitute/firecloud_ui/page/workspaces_list.cljs
davidangb: I explicitly changed this from `not` to `false?`. In the case where the user does not have any saved state for a given label, we default the checkbox to checked and enable the filter. However, when the user clicks the checkbox and `:onChange` fires, we would attempt to `(not nil)` which is `true`. End result is that it would take two clicks for the user to uncheck the checkbox. `false?` does the right thing here for the user.

MattPutnam: Ah, I see.

PR #609, opened 2016-12-13
Fixes race condition in sign-in

dmohs:
Better use of the Google JavaScript listener.

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
@@ -28,7 +28,8 @@
 
 (defn call-external-object-method
   "Call an external object's method by name, since a normal call will get renamed during
-   advanced compilation and cause an error."
+   advanced compilation and cause an error.
+   DEPRECATED: this is what js-invoke is for"
   [obj method-name & args]
   (apply (.bind (aget obj (name method-name)) obj) args))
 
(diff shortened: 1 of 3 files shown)


src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
```diff
@@ -84,10 +85,26 @@
...
+      (aget "currentUser")
+      (js-invoke
+       "listen" (fn [u]
+                  (doseq [[_ on-change] @user-listeners]
+                    (on-change u))))))
 
 (defn get-access-token []
-  (-> @google-auth2-instance (.-currentUser) (.get) (.getAuthResponse) (.-access_token)))
+  (-> @google-auth2-instance
+      (aget "currentUser") (js-invoke "get") (js-invoke "getAuthResponse") (aget "access_token")))
```
zarsky-broad, line 107: Is this really better? It seems a lot more complicated.

dvoet: the old way may be been simpler but it did not work... working is better.

dvoet (approved):
👍

davidangb:
Does this fix GAWB-1394?

src/cljs/org/broadinstitute/firecloud_ui/utils.cljs
dmohs: This pattern is necessary to support advanced compilation. I had to turn that off a while back, but I'd like to collect more data before we decide to abandon it.

dmohs:
My suspicion is that this does fix https://broadinstitute.atlassian.net/browse/GAWB-1394 but I'll want to test that independently.

PR #614, opened 2016-12-19
GAWB-1443: use appropriate auth token for IGV

davidangb:
This is fallout from the login/logout/oauth refactor. In IGV, we were attempting to use an auth token from your cookie - but we no longer write that cookie in all cases. Fix is to use the auth token from the Google library instead.

I've also removed the `get-access-token-cookie` function, so we won't try to call it again in the future.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/analysis/igv.cljs
+++ b/src/cljs/org/broadinstitute/firecloud_ui/page/workspace/analysis/igv.cljs
@@ -17,7 +17,7 @@
                               {:name (str "Track " (inc index))
                                :url (common/gcs-uri->google-url track-url)
                                :indexURL (when (string? @index-url) (common/gcs-uri->google-url @index-url))
-                               :headers {:Authorization (str "Bearer " (utils/get-access-token-cookie))}
+                               :headers {:Authorization (str "Bearer " (utils/get-access-token))}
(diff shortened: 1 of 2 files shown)


davidangb:
Dunno who's working today - @ansingh7115 , @jmthibault79 , @dmohs could I get a review? Who's tech-leading for analysis team - anyone claim TL for this too?

dmohs (approved):
Definitely the right fix. Great to remove the offending `get-access-token-cookie` function too.

breilly2:
There's still `set-access-token-cookie` and `delete-access-token-cookie`. `delete-` is only called from `set-`, but `set-` is called from a few places in code. It's probably worth auditing those to make sure they can be removed as well.

davidangb:
I don't think `set-access-token-cookie` (and, therefore, `delete-access-token-cookie`) should be removed. It's still used for downloading GCS objects in the `/cookie-authed/download/b/{bucket}/o/{object}` endpoint.

breilly2:
Right, and FCtoken is read in orchestration. Got it. On the surface, it seemed weird to leave calls to a setter for a something that isn't read, but the read is elsewhere.

PR #631, opened 2017-01-12
GAWB-1458: never cache index.html in browser

davidangb:
Configure Apache so that it aggressively tells the browser not to cache index.html (and only index.html).

Doing this from the Apache side is more reliable than using meta tags inside index.html - and I believe those meta tags go away with HTML5 anyway.

I don't know the best way to test this. The `/config/start-server.sh` script and `/config/docker-rsync-local-ui.sh` run outside apache, so they won't see these changes. You could manually build a local docker image with these changes and run that, I suppose. I tested by adding a volume to /config/docker-compose.yaml:
```
    - ../src/docker/site.conf:/etc/apache2/sites-available/site.conf
```
and then starting up via that docker-compose ...

Once this hits dev (and beyond), it *should* just work.

In Chrome Dev Tools, compare the response headers for index.html vs. other assets, such as compiled.js. For index.html, and only index.html, you should see the Cache-Control, Pragma, and Expires: 0 headers.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/docker/site-https-only.conf
+++ b/src/docker/site-https-only.conf
@@ -26,11 +26,6 @@ Header always set X-Frame-Options SAMEORIGIN
         Order allow,deny
         Allow from all
         Require all granted
-        <Files "index.html">
-          Header set Cache-Control "no-store, no-cache, must-revalidate"
-          Header set Pragma "no-cache"
-          Header set Expires 0
-        </Files>
     </Directory>
 
     Alias /config.json /var/www/html/config.json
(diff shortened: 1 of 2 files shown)


dmohs (approved):
I'm mystified as to why this basic mechanism (HTTP caching) has broken down in our infrastructure, but I am confident (after reading the bug) that the problem is real and that this fix is reasonable. Although less of a problem, it may make sense to apply the same technique to `config.json`.

davidangb:
I've also set it to disallow browser caching for config.json. This required using `<FilesMatch>` instead of `<Files>`, and moving the directive to the top server level instead of inside a Directory/VirtualHost, since config.json is controlled by an `Alias` and served out of /var/www/html. Re-review, @dmohs?

src/docker/site-https-only.conf
```diff
@@ -55,3 +55,9 @@ Header always set X-Frame-Options SAMEORIGIN
 
     Alias /config.json /var/www/html/config.json
 </VirtualHost>
+
+<FilesMatch "(^index\.html|config\.json)">
```
dmohs, line 59: Thank you for being careful about your regexing.

dmohs (approved):

PR #637, opened 2017-01-18
Assert that config.json is complete

dmohs:
https://broadinstitute.atlassian.net/browse/GAWB-1334

Corresponding launch script change: https://github.com/broadinstitute/firecloud-develop/pull/270

Developer usability improvement suggested by David An. We validate config.json, which is especially helpful when we make changes to it but sometimes forget to propagate those changes to our deployment environments.

This PR also introduces unit testing for UI code. I have gathered feedback from several UI developers on the usability of running these tests (see README). This initial version is not yet ready for integration into our PR process, but it sets the stage.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/cljs/org/broadinstitute/firecloud_ui/testrunner.cljs
```diff
@@ -0,0 +1,10 @@
+(ns org.broadinstitute.firecloud-ui.testrunner
+  (:require
+   cljs.test
+   org.broadinstitute.firecloud-ui.configtest
+   org.broadinstitute.firecloud-ui.main
+   org.broadinstitute.firecloud-ui.testutils
+   ))
+
+(defn run-all-tests []
+  (cljs.test/run-tests 'org.broadinstitute.firecloud-ui.configtest))
```
breilly2: Do all tests have to be manually added here? Is there any way to auto-discover tests?

dmohs: It doesn't exist by default, though we could write it. I'd prefer to not do that for this initial version.

README.md
```diff
@@ -40,10 +40,10 @@ To connect, reload the browser window. The prompt should appear less than ten se
 
 ## Testing
 
-To run all tests, enter this at the Figwheel REPL:
+To run all tests, enter this at the JavaScript console:
 
-```clojure
-(broadfcui.testrunner/run-all-tests)
+```javascript
+broadfcuitest.testrunner.run_all_tests()
```
breilly2, line 46: I like this. Brings together where you run the tests from and where to look for the output.

src/cljs/main/broadfcui/common.cljs
```diff
@@ -1,10 +1,10 @@
-(ns org.broadinstitute.firecloud-ui.common
+(ns broadfcui.common
```
breilly2, line 1: All of this file moving could cause issues for people when merging previously created branches. I think it would be a good idea to warn people before merging this to develop.

breilly2 (approved):
:+1: Thanks for listening to the feedback (here and in person). I think this is a great start at having some focused UI tests!

src/cljs/main/broadfcui/common.cljs
dmohs: Thank you. I believe I have warned everyone who cares except for David An, who is a reviewer on this PR. I've also squashed the commits to four and I'm going to merge them without squashing so the steps are clearer and the rename is on its own.

README.md
```diff
@@ -10,41 +10,49 @@ https://firecloud.dsde-prod.broadinstitute.org/
...
-Build once:
-```
-./script/build-once.sh
+```bash
+./config/docker-rsync-local-ui.sh
 ```
 
-Watch files and rebuild whenever a file is saved:
-```
-./script/start-auto-build.sh
```
davidangb, line 35: should start-auto-build.sh still exist in the source tree, then? Or any of the scripts? I wonder if some, like build-once.sh, are still useful.

Update: build-once.sh is now failing for me on this branch - so it either needs an update or it needs to be deleted.

project.clj
```diff
@@ -1,40 +1,45 @@
...
-
 (defproject org.broadinstitute/firecloud-ui "0.0.1"
   :dependencies
   [
-   [dmohs/react "0.2.11"]
-   [org.clojure/clojure "1.7.0"]
-   [org.clojure/clojurescript "1.7.228"]
+   [dmohs/react "1.0.2+15.0.2"]
+   [org.clojure/clojure "1.8.0"]
+   [org.clojure/clojurescript "1.9.293"]
```
davidangb, line 6: we're now way ahead of the web repl at http://clojurescript.net/ ... know of any others?

project.clj
```diff
@@ -15,18 +15,21 @@
...
               {:builds
                {:client
-                {:figwheel true
+                {:source-paths ["src/cljstest"]
+                 :figwheel true
                  :compiler
-                 {:optimizations :none
+                 {;; Use this namespace (which requires main) so that testing is readily available
+                  ;; in all dev builds.
+                  :main ~(with-ns "testrunner")
```
davidangb: is it worth adding `:load-tests false` for non-dev builds, just in case?

davidangb (commented):
some comments inline, too.

I wish this had been a few separate PRs. I see independent work in here on a few fronts:
1) renaming/moving namespaces
2) updating third-party libraries and restructuring build directories (these could even be broken apart)
3) adding unit tests

The third-party updates are the part that makes me the most nervous - those warrant pretty wide testing and/or reviewing lots of release notes - which would delay merging unit tests. I haven't done any of the testing/reviewing of the 3p libs or build changes.

project.clj
```diff
@@ -1,40 +1,45 @@
...
+                  :preloads [devtools.preload]
+                  :external-config {:devtools/config {:features-to-install
+                                                      [:formatters :hints]}}}}}}}
              :deploy {:cljsbuild
-                      {:builds {:client {:source-paths ["src/cljsprod"]
-                                         :compiler
-                                         {:main ~(with-ns "render")
-                                          ;; As of 10/29/15, advanced optimization triggers
+                      {:builds {:client {:compiler
+                                         {;; As of 10/29/15, advanced optimization triggers
```
davidangb, line 36: does `clean` in the `:deploy` profile still work? I'm getting errors locally, but it could be me.

project.clj
dmohs: That's the only one I know about. However, I wouldn't trust it for anything besides a trivial expression, since the versions will always be different than our own code. Instead, I'd recommend the REPL started by Figwheel, which is running our versions.

README.md
dmohs: I went with the delete. My intuition is that these aren't used often-enough to warrant keeping them current. Does that sound right to you?

project.clj
dmohs: Just tried this `lein with-profile deploy clean` which worked for me.

dmohs:
Regarding `:load-tests false`: I'm not familiar with this option and can't find any docs. I'd prefer to somehow assert that our build is correct over doubling up on a config option. I'll ponder.

Regarding renaming: I completely agree. I tried twice to cherry-pick that commit out into a separate PR, but it got messy. I think the best course of action is to just proceed and note that this is not the right practice.

Regarding third-party-libraries: I agree with you in general, but in this case I didn't want to risk implementing unit tests and have the approach I took not work with the current versions of our libraries, so I felt it necessary to do the upgrade first. That also gave me time to develop and test with them while I worked on this code.

For what it's worth, I work on around five projects using this stack (including dbLoF) and they are all running versions very close to these and have been for at least six months.

README.md
davidangb: probably folks aren't using them? But I suggest polling the team to find out.

davidangb:
this is `:load-tests`: https://github.com/clojure/clojurescript/wiki/Compiler-Options#load-tests

davidangb (approved):
👍 approving, with these additional actions:
* make sure none of the devs (or Jenkins, etc) are using the deleted scripts
* read docs on `load-tests` and ponder including it for the deploy build, either in this PR or a future one
* figure out how to inform dev team to clean up now-obsolete resources->target symlink

PR #645, opened 2017-01-25
Make deploy build also write to compiled.js

dmohs:
This commit did a bad: https://github.com/broadinstitute/firecloud-ui/commit/86606042026aa1b53f563bdd9dd8b56a55744fac

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -39,7 +39,8 @@
                                           :optimizations :simple
                                           :pretty-print false}}}}}}
   :cljsbuild {:builds {:client {:source-paths ["src/cljs/main"]
-                                :compiler {:main "broadfcui.main"}}}}
+                                :compiler {:main "broadfcui.main"
+                                           :output-to "target/compiled.js"}}}}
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -39,7 +39,8 @@
                                           :optimizations :simple
                                           :pretty-print false}}}}}}
   :cljsbuild {:builds {:client {:source-paths ["src/cljs/main"]
-                                :compiler {:main "broadfcui.main"}}}}
+                                :compiler {:main "broadfcui.main"
+                                           :output-to "target/compiled.js"}}}}
```
breilly2, line 43: About all I can do is confirm that this reintroduces `:output-to "target/compiled.js"` which certainly seems important.

I also noticed that commit removed `:output-dir` and `:asset-path`. Do those need to be revived as well?

dmohs: I really don't think so. I've tested this as best I could.

breilly2 (approved):
:+1:

PR #646, opened 2017-01-25
Add new key to config validation

dmohs:
This commit added a new config value that I didn't see when I was writing the test: https://github.com/broadinstitute/firecloud-ui/commit/3820cec8d6619e10c9f2dfc1fbc43dc4a752e3db


dmohs:
Anu says 👍

PR #673, opened 2017-02-21
Iz gawb 1577 spike attempt tooltip implementation with foundation

zarsky-broad:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/cljs/main/broadfcui/common.cljs
```diff
@@ -210,3 +210,26 @@
...
+           position (:position props)]
+       [:div {}
+        [:button {:className "has-tip" :data-toggle rand-id}
+         (:text props)]
+        [FoundationComponent
+         {:contents [:div {:className (str "tooltip " position) :id rand-id :data-dropdown ""
+                           :data-v-offset (if (or (not position) (= position "top")) 12 100)
+                           :data-h-offset (if (or (= position "left") (= position "right")) 12 100)
+                           :data-hover true :data-hover-pane true}
+                     (:tooltip props)]}]]))})
```
MattPutnam: (Comment is meant to be attached to the whole component, not just this last line)

Recently I've been unpacking `props` fully.  I find it helps document what the component's properties are.  So in that `let`, instead of just pulling out `position`, do `{:keys [position text tooltip]} props` and then use those.

It would also be good to have a default `:position` prop, both for ease of use and to document what its valid values look like.

MattPutnam (commented):
Looks pretty reasonable to me.  I'm having trouble building on this branch but I trust it works.

src/cljs/main/broadfcui/common.cljs
zarsky-broad: Default for position is blank, which implicitly means bottom. But I take your point.

dmohs:
Let's add a little padding around the edge so if the user drifts slightly off the tooltip while copying it doesn't close. Let's also see one example of showing the tooltip above the text. Maybe anchor that one to a different location so we can see the difference in the code.

Otherwise, I think this is ready for comparison. I'll put a meeting on people's calendars.

dmohs:
If I put my mouse even one pixel off the edge, the tooltip closes. Can we give a little room there? I'd prefer to see a little less padding around the text, but with this forgiveness.

There is a bit of weirdness now on the summary page. When I load it, the scrollbar is visible. After I hover over each tip, the text moves a bit and the scrollbar goes away. Can you fix please?

zarsky-broad:
Re: the lack of hover-padding: I'm afraid that's baked into foundation. If this is a deal-breaker for you, that does this solution in. That said, Foundation has gone through 6 versions and is widely used, and they consider it sufficient.

Here's a thought, though:
We're treating tooltips as the way to show _any_ additional information regarding a hovered object. I propose that we think more granularly.
- A tooltip is for short explanatory text, that won't ever have to be copied. For example, hovering over an icon might open a tooltip that describes it, i.e. "in progress".
- We need a separate solution for larger quantities of information. I'm going to propose an info icon that goes next to text. When clicked, it opens a dropdown (see example: http://foundation.zurb.com/sites/docs/dropdown.html) that has the additional information. This is only dismissed on another click outside of it.

The purpose of this bifurcation is to avoid overloading one type of control. For the latter scenario, where there's more information that the user may want, and may wish to be able to select, a click is a more
conclusive interaction, which lends itself to a larger information display like that.

Thoughts?

MattPutnam:
I actually like my tooltips to go away quickly and easily.  Too often there's a tooltip blocking what I want and I have to move the mouse horribly far away in order to dismiss it, and then carefully navigate around whatever tripped the tooltip when I try again.

dmohs:
@zarsky-broad I can get on board with your arguments, and I think we're hitting on the core issue, which is why this spike is so much more valuable than just tooltips.

I have enough experience that my decisions on usability and code are more often right than wrong, but that leaves an enormous amount of room for mistakes and errors in judgement. I have no choice but to assume that library developers are in the same boat.

That means that we will encounter scenarios where they made a choice with which we do not agree or that does not work for our specific use case. What then? It's up to chance whether the adjustment we need is even possible.

To Matt's point (@MattPutnam), I noticed that the foundation tip has a slight delay before disappearing, which almost subconsciously makes me less confident that moving my mouse away actually worked to dismiss it. That begs the million dollar question: Did I get lucky and this is easy to fix in the library?

I think this should be one focus of our discussion in the meeting. If I can be talked down from this concern, then I can pretty much support Foundation 100%.

In the meantime, I really do want a fix for the scrollbar/text jumping in this version (I can reproduce reliably if you can't). I would be disappointed if we had to live with that.

zarsky-broad:
Here's the thing about my implementation, here:
_I'm not actually using foundation tooltips._
I'll give you a moment to collect your jaws from the floor. But @dmohs that pause is built in to Foundation's hover-dropdowns, not its tooltips.

I think my style overloading might be causing the scrollbar issue, I'll look into it.

zarsky-broad:
I'm going to add to this some examples of a tooltip proper, and of a dropdown proper, as I mentioned above.

dmohs:
Take a look at this contrived style modification:
```
diff --git a/src/cljs/main/broadfcui/page/workspace/summary/tab.cljs b/src/cljs/main/broadfcui/page/workspace/summary/tab.cljs
index e2c6202..dfc228d 100644
--- a/src/cljs/main/broadfcui/page/workspace/summary/tab.cljs
+++ b/src/cljs/main/broadfcui/page/workspace/summary/tab.cljs
@@ -157,8 +157,9 @@
      [:div {:style {:display "flex" :flexWrap "wrap"}}
       (render-detail-box
         1
-        [common/FoundationTooltip {:text (str "Workspace Owner" (when (> (count owners) 1) "s"))
-                                   :tooltip "Hey look, a tooltip"}]
+        [:div {:style {:position "relative" :height "2rem" :overflow "hidden"}}
+         [common/FoundationTooltip {:text (str "Workspace Owner" (when (> (count owners) 1) "s"))
+                                    :tooltip "Hey look, a tooltip"}]]
         (style/create-paragraph
           [:div {}
            (interpose ", " owners)
```

In that case, the tooltip doesn't fully un-hide on hover. Even though the example is contrived, the styles are similar to those of a table cell. I believe it would be confusing to developers using this if we did not support that use case.

zarsky-broad:
@dmohs Fair point. If we're going to go with this solution, I think I'd prefer we use `FoundationProperTooltip` and `FoundationInfoBox`, which I've added. They are the native Foundation implementations with a little bit of sugar, and I'd expect them to be more robust.

PR #680, opened 2017-02-28
GAWB-1483: enable sorting on library columns

davidangb:
This complements and requires broadinstitute/firecloud-orchestration#385. Without that orch branch, sorting won't work.

Technical changes in this PR:
1. Turn on column sorting for the datasets table in Library.
2. Reset sort when a user performs a new text search.
3. Use keys instead of names to identify columns in the datasets table in Library (e.g. use ":library:datasetName" instead of "Cohort Name"). This was required for sort, but also helps us with local-storage persistence in the future - we are unlikely to rename a key, but likely to rename a name.
4. Bump the persistence version for the `library-table` local storage value. This was necessary because of the previous bullet point; we're now persisting keys instead of names.
---
5. Delete the `DatasetTables :set-filter-text` method - this was never used.
6. Tweak the `library-page` persistence value to exclude `:library-attributes` - we no longer store the entire Library schema in local storage. This felt dangerous because it's easy for local storage to get out of sync with the actual schema - and it totally bloated the content of this local storage key.
7. Bump the persistence version for the `library-page` local storage value, to force end users to reset the value and therefore ensure they are not storing the `:library-attributes` copy of the Library schema.

Bullet points 5-7 are totally scope creep for this PR and are not necessary to achieve the jira ticket; I realize I am adding them as riders!

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -400,7 +400,9 @@
      (:query-params @state))
    :update-query-params
    (fn [{:keys [state]} new-params]
-     (swap! state update :query-params merge new-params))
+     (let [change? (not= new-params (select-keys (:query-params @state) (keys new-params)))]
+       (swap! state update :query-params merge new-params)
+       change?))
    :get-default-props get-default-table-props
    :get-initial-state get-initial-table-state
    :render render-table
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -401,7 +401,8 @@
    :update-query-params
    (fn [{:keys [state]} new-params]
      (let [change? (not= new-params (select-keys (:query-params @state) (keys new-params)))]
-       (swap! state update :query-params merge new-params)
+       (when change?
+         (swap! state update :query-params merge new-params))
        change?))
    :get-default-props get-default-table-props
    :get-initial-state get-initial-table-state
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -395,7 +395,6 @@
 
 
 (react/defc Table
-  (utils/log-methods "Table"
   {:get-query-params
    (fn [{:keys [state]}]
      (:query-params @state))
@@ -420,4 +419,4 @@
    :component-will-unmount
    (fn [{:keys [locals]}]
      (.removeEventListener js/window "mousemove" (:mouse-move-handler @locals))
-     (.removeEventListener js/window "mouseup" (:mouse-up-handler @locals)))}))
+     (.removeEventListener js/window "mouseup" (:mouse-up-handler @locals)))})
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -400,9 +400,11 @@
      (:query-params @state))
    :update-query-params
    (fn [{:keys [state]} new-params]
-     (let [change? (not= new-params (select-keys (:query-params @state) (keys new-params)))]
+     (let [old-state (:query-params @state)
+           new-state (merge old-state new-params)
+           change? (not= old-state new-state)]
        (when change?
-         (swap! state update :query-params merge new-params))
+         (swap! state assoc :query-params new-state))
        change?))
    :get-default-props get-default-table-props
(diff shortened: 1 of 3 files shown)


rushtong (approved):
👍 LG, works great!

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -16,23 +16,20 @@
...
-     (react/call :update-query-params (@refs "table") {:filter-text new-filter-text :current-page 1}))
-   :render
+  {:render
    (fn [{:keys [state this props]}]
      (let [attributes (:library-attributes props)
            search-result-columns (:search-result-columns props)
            extra-columns (subvec search-result-columns 4)]
        [table/Table
-        {:ref "table" :state-key "library-table" :v 1
+        {:ref "table" :state-key "library-table" :v 2
```
ahaessly: can we add some constants for the name and version number similar to what we did below for library page?

src/cljs/main/broadfcui/common/table.cljs
```diff
@@ -366,13 +366,14 @@
...
+        mouse-up-handler
         #(when (:dragging? @state)
-          (common/restore-text-selection (:saved-user-select-state @state))
-          (swap! state dissoc :dragging? :drag-column :mouse-x :saved-user-select-state)))
-  (.addEventListener js/window "mouseup" (.-onMouseUpHandler this)))
+           (common/restore-text-selection (:saved-user-select-state @state))
+           (swap! state dissoc :dragging? :drag-column :mouse-x :saved-user-select-state))]
+    (.addEventListener js/window "mousemove" mouse-move-handler)
+    (.addEventListener js/window "mouseup" mouse-up-handler)
+    (swap! locals assoc :mouse-move-handler mouse-move-handler :mouse-up-handler mouse-up-handler)))
```
MattPutnam, line 385: All of the changes here are just reorganization, changing it so these functions are stored in locals rather than attached to the object directly.

src/cljs/main/broadfcui/common/table.cljs
```diff
@@ -394,17 +395,16 @@
    :get-default-props get-default-table-props
    :get-initial-state get-initial-table-state
    :render render-table
-   :execute-search (fn [{:keys [this]}] (react/call :refresh-rows this))
```
MattPutnam, line 406: Since execute-search is just calling refresh-rows, I deleted it and made callers call refresh-rows directly

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -311,6 +318,5 @@
                          (select-keys @state [:library-attributes :search-result-columns :search-text
                                               :facet-filters :aggregate-fields :expanded-aggregates]))])]])
    :component-did-update
-   (fn [{:keys [state refs]}]
-     (persistence/save {:key PERSISTENCE-KEY :state state})
-     (react/call :execute-search (@refs "dataset-table")))})
+   (fn [{:keys [state]}]
+     (persistence/save {:key PERSISTENCE-KEY :state state :except [:library-attributes]}))})
```
MattPutnam, line 333: This was the bulk of the problem when reloading.  Don't re-search on every update, explicitly do it when changing the search text or facets.

src/cljs/main/broadfcui/common/table.cljs
```diff
@@ -399,24 +400,25 @@
      (:query-params @state))
    :update-query-params
    (fn [{:keys [state]} new-params]
-     (swap! state update :query-params merge new-params))
+     (let [change? (not= new-params (select-keys (:query-params @state) (keys new-params)))]
+       (swap! state update :query-params merge new-params)
+       change?))
```
davidangb: does this need to check values as well as keys?

MattPutnam: It does.  But I see how it looks like it doesn't.  Let me see if I can rewrite that to be clearer.

skwalker:
code looks good to me, works well 👍

davidangb:
this is so much better than it used to be. Still - sort order is supposed persist across facet selections (sort order only clears on a new text search). Is that feasible? If PO agrees, we could save that for a follow-on story.

davidangb:
👍 looks good to me! (I can't approve since I technically created this PR)

PR #683, opened 2017-03-02
GAWB-1688:  Make file previews disappear for empty files

ansingh7115:
What the file previews looked like before when the file was empty:

screen shot 2017-03-02 at 4 33 45 pm

What the file previews will look like now when the files are empty: 

screen shot 2017-03-02 at 2 49 25 pm

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/gcs_file_preview.cljs
+++ b/src/cljs/main/broadfcui/common/gcs_file_preview.cljs
@@ -28,8 +28,7 @@
           (labeled "Google Bucket" (:bucket-name props))
           (labeled "Object" (:object props))
           [:div {:style {:marginTop "1em"}}
-           [:div {} (if (or (= data-size "0") (clojure.string/blank? data-size))
-                      "File empty"
+           [:div {} (when-not (or (= data-size "0") (clojure.string/blank? data-size))
                       "Previews for some filetypes may be unsupported. ")]
            (when (> data-size preview-byte-count) (str "Last " (:preview-line-count @state)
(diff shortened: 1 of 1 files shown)


src/cljs/main/broadfcui/common/gcs_file_preview.cljs
```diff
@@ -28,34 +28,40 @@
           (labeled "Google Bucket" (:bucket-name props))
           (labeled "Object" (:object props))
           [:div {:style {:marginTop "1em"}}
-           [:div {} "Previews for some filetypes may be unsupported. "]
+           [:div {} (when-not (or (= data-size "0") (clojure.string/blank? data-size))
```
MattPutnam: We're checking `(or (= data-size "0") (clojure.string/blank? data-size))` several times, so let's pull this test out into a common function.

MattPutnam (approved):
👍 with one refactoring

dmohs:
Thanks for the screenshots! +1 to what Putnam said. Also, it's weird that you have to check for both "0" and blank, so a comment about why that is would be helpful (if you know).

ansingh7115:
I tried implementing it with only checking for "0" and not blank - and the behavior I saw was that the preview stuff would load and then a split-second later would disappear. The reason for this was that the data-size was nil for that split-second and then would be populated by the actual size. 

I decided to use blank? because it handles nil, empty, or only whitespace cases.

PR #708, opened 2017-03-27
GAWB-1748: mc's with colons in their name won't be weird now

ansingh7115:
Before, there was weirdness when a method config had a colon in it's name because the nav context was parsing method config ns and names like this (ns:mcname) --> (:namespace ns :name mcname). It was using the colon as the barrier. So if you had colons in the name, the rest would get cut off, like so (ns:mcna:me) --> (:namespace ns :name mcna).

So I put a limit on the number of times to split the string.
This assumes that the rest of the url is the name of the method config, which it should be. 
The function we changed ( get-id-from-nav-segment) is also used by the workspace for it's namespace and name. However, since there is a strict bar on what characters can be used in a namespace and workspace name, this shouldn't be a problem. 
The way navigation is done will change soon (as in this or next sprint) so this way of doing things will disappear in bit of time. 

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common.cljs
+++ b/src/cljs/main/broadfcui/common.cljs
@@ -154,7 +154,7 @@
 
 (defn get-id-from-nav-segment [segment]
   (when-not (clojure.string/blank? segment)
-    (let [[ns n] (clojure.string/split segment #":")]
+    (let [[ns n] (clojure.string/split segment #":" 2)]
       {:namespace ns :name n})))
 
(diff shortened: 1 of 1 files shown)


helgridly:
beautiful 👍

dmohs (approved):
I echo Hussein.

PR #714, opened 2017-03-28
GAWB-1757: Table refactor

MattPutnam:
Implemented in the billing projects, monitor tab, and library page.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/cljs/main/broadfcui/page/billing/billing_management.cljs
```diff
@@ -69,64 +70,63 @@
...
-                         (style/create-link {:text projectName
-                                             :onClick #((:on-select props) projectName)})
-                         :else projectName)
-                       (when message
-                         [:div {:style {:float "right" :position "relative"}}
-                          [common/FoundationInfoBox
-                           {:text [:div {} [:strong {} "Message:"] [:br] message]}]])])}
-                   {:header "Role" :starting-width :remaining :resizable? false}]
+       [Table
+        {:table {:behavior {:reorderable-columns? false}
```
dmohs: Let's rename this key to `:core` or `:body` since we've established "table" to mean the entire thing including toolbars and pagination.

dmohs: Very good call to namespace these top-level keys though.

src/cljs/main/broadfcui/common/flex_utils.cljs
```diff
@@ -4,12 +4,12 @@
...
+(defn box [attrs & children]
   [:div (utils/deep-merge {:style {:display "flex"}} attrs)
    children])
 
-(defn flex-strut [size]
+(defn strut [size]
   [:div {:style {:flexGrow 0 :flexShrink 0 :flexBasis size}}])
 
-(def flex-spacer
+(def spring
```
dmohs, line 14: Worth noting: If your parent says `{:justifyContent "space-between"}` you get a very similar effect to what this is used for. However, this might actually be cleaner.

src/cljs/main/broadfcui/common/table/utils.cljs
```diff
@@ -0,0 +1,74 @@
+(ns broadfcui.common.table.utils
+  (:require
+    [broadfcui.utils :as utils]
+    ))
+
+
+(defn resolve-id [{:keys [header id]}]
+  (assert (or (not id) (string? id)) "Column id must be a string, if present")
+  (assert (or id (string? header)) "Every column must have a string header or id")
```
dmohs, line 9: These assertions make me happy.

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -22,76 +24,73 @@
...
+                           " Dataset"
+                           (when-not (= 1 total) "s")
+                           " found"))]]
+                  flex/spring]
+          :style {:alignItems "flex-start" :marginBottom 7}
+          :column-edit-button {:style {:order 1 :marginRight nil}
+                               :anchor :right}}}]))
    :execute-search
    (fn [{:keys [refs]} reset-sort?]
      (let [query-params (merge {:current-page 1} (when reset-sort? {:sort-column nil :sort-order nil}))]
```
dmohs: `:current-page` doesn't seem to work. I think you meant `:page-number`.

To prevent this type of mistake, I think you can probably assert that `:update-query-params` may only get certain keys and anything else is an error. The logic in config.cljs may be helpful.

src/cljs/main/broadfcui/common/table/column_editor.cljs
```diff
@@ -0,0 +1,130 @@
...
+                           (dismiss))}
+          (icons/icon {:style {:fontSize 18 :lineHeight "inherit"}} :reset)
+          [:span {:style {:fontSize 14 :flexGrow 1 :textAlign "center"}}
+           "Reset Columns"]]
+         "Show:"
+         (let [style {:width "4rem" :padding "4px 8px" :marginRight "0.5rem" :borderRadius 5}]
+           [:div {:style {:padding "0.5em 0"}}
+            [comps/Button {:style style :onClick #(set-all props true) :text "All"}]
+            [comps/Button {:style style :onClick #(set-all props false) :text "None"}]])
+         (map-indexed
```
dmohs: I think fixed columns would be clearer to users if they showed up as immovable in this list.

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,140 @@
...
+       :rows []))
+   :render
+   (fn [{:keys [props state]}]
+     (let [props (utils/deep-merge default-props props)
+           {:keys [rows column-display total-count filtered-count query-params]} @state
+           {:keys [table toolbar paginator]} props
+           {:keys [empty-message columns behavior external-query-params]} table
+           {:keys [fixed-column-count allow-no-sort?]} behavior
+           query-params (merge query-params (select-keys props external-query-params))
+           update-column-display #(swap! state assoc :column-display %)]
```
dmohs: In lieu of documentation, this might be a good place to make some assertions about the structure of the `:columns` key and fields within. Same goes for any other keys that are crucial.

dmohs (commented):
Overall this is excellent and a joy to read through. My comments are few and minor considering the size of this PR. I'd love to take one more look to see how you addressed this issues.

davidangb (commented):
I don't have much to add - looks good, overall!

One testing note: we seem to have lost tooltips for each cell.

src/cljs/main/broadfcui/common/flex_utils.cljs
MattPutnam: With this you can specify exactly where you want to have extra space:  `[spring comp1 (strut 10) comp2 spring]` to center two components with a fixed gap between them.

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+    [broadfcui.common.icons :as icons]
+    [broadfcui.common.table.utils :as table-utils]
+    [broadfcui.utils :as utils]
+    ))
+
+
+(defn- flex-params [width]
+  (if (= width :auto)
+    {:flexBasis "auto" :flexGrow 1 :flexShrink 1}
+    {:flexBasis width :flexGrow 0 :flexShrink 0}))
```
zarsky-broad: This can be written as:
```cljs
{:flex "1 auto"}
{:flex "none" :width width}
```

If you think this is unclear, we can certainly leave it as is, but this is more idiomatic to CSS.

src/cljs/main/broadfcui/common/icons.cljs
```diff
@@ -33,6 +33,8 @@
    :settings "fa-cog"
    :share "fa-share-alt"
    :shield "fa-shield"
+   :sort-asc "fa-sort-asc"
+   :sort-desc "fa-sort-desc"
```
zarsky-broad: In reviewing this, I think I actually prefer [fa-sort-amount-asc](http://fontawesome.io/icon/sort-amount-asc/), it seems clearer

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+
+
+(defn- header [{:keys [joined-columns sort-column sort-order set-sort style
+                       start-column-drag column-reset allow-no-sort?]}]
+  [:div {:style (merge {:display "flex"} (:row style) (:header-row style))}
+   (map-indexed
+    (fn [index {:keys [id width initial-width header visible? resizable? sortable?]}]
+      (when visible?
+        [:div {:style (merge (flex-params width)
+                             {:position "relative" :cursor (when sortable? "pointer")}
```
zarsky-broad, line 29: Man, I wish we had a better solution than making this a position relative, just so we can make the resize thingy position absolute...

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+                                                  {:borderRight "1px solid" :marginRight -1})))}
+         [:div {:style (merge {:display "flex" :width width}
+                              (:cell style)
+                              (:header-cell style))
+                :onClick (when sortable?
+                           #(cond (not= sort-column id) (set-sort id :asc)
+                                  (= sort-order :asc) (set-sort id :desc)
+                                  allow-no-sort? (set-sort nil nil)
+                                  :else (set-sort id :asc)))}
+          [:div {:style {:flex "1 1 auto" :overflow "hidden" :textOverflow "ellipsis"}}
```
zarsky-broad, line 40: This can be `:flex "1 auto"`, see previous comment.

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+           header]
+          flex/spring
+          (when (= id sort-column)
+            (icons/icon {:style {:margin "0 0.4rem 0 0.1rem"}}
+                        (if (= :asc sort-order) :sort-asc :sort-desc)))]
+         (when resizable?
+           [:div {:style {:position "absolute" :cursor "col-resize"
+                          :right -11 :top 0 :width 21 :height "100%" :zIndex 1}
```
zarsky-broad: How were these numbers arrived at? [Magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) make me nervous...

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+                     :title (cond as-text (as-text column-value)
+                                  (string? rendered) rendered)}
+               rendered])))
+        joined-columns)])
+    rows)])
+
+
+(defn- resolve-column-props [{:keys [initial-width as-text] :as props} behavior]
+  (merge {:resizable? (and (not= initial-width :auto) (:resizable-columns? behavior))
+          :sortable? (:sortable-columns? behavior)
```
zarsky-broad, line 77: I could **absolutely** be confused, but does this mean that individual columns can't be marked as non-sortable?

src/cljs/main/broadfcui/common/table/body.cljs
```diff
@@ -0,0 +1,122 @@
...
+            start-column-drag
+            (fn [{:keys [e width index]}]
+              (swap! locals assoc :start-mouse-x (.-clientX e) :start-width width)
+              (swap! state assoc
+                     :dragging? true :drag-column index
+                     :saved-user-select-state (common/disable-text-selection)))
+            column-reset
+            (fn [{:keys [index initial-width]}]
+              (update-column-display (assoc-in column-display [index :width] initial-width)))
+            +props (merge props (utils/restructure joined-columns start-column-drag column-reset))]
```
zarsky-broad: I presume leading with a plus like this is idiomatic, but it makes me feel squidgy. Can we change this name?

src/cljs/main/broadfcui/common/table/column_editor.cljs
```diff
@@ -0,0 +1,148 @@
...
+    [broadfcui.common.icons :as icons]
+    [broadfcui.common.overlay :as overlay]
+    [broadfcui.common.table.utils :as table-utils]
+    [broadfcui.common.style :as style]
+    [broadfcui.utils :as utils]
+    ))
+
+
+(defn- set-all [{:keys [column-display update-column-display fixed-column-count]} visible?]
+  (->> (drop fixed-column-count column-display)
```
zarsky-broad, line 15: I'm not sure why this is happening like this, but then I'm not _entirely_ sure what this is doing. Why are we removing members from a collection…?

src/cljs/main/broadfcui/common/flex_utils.cljs
```diff
@@ -4,12 +4,12 @@
     ))
 
 
-(defn flex-box [attrs & children]
+(defn box [attrs & children]
```
zarsky-broad, line 7: This is only used in one place, since we often seem to merge multiple props together. Do we really need it as a util?

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+  (cond
+    (<= total-pages 5) (range 1 (inc total-pages))
+    (<= current-page 3) (range 1 6)
+    (>= current-page (- total-pages 2)) (range (- total-pages 4) (inc total-pages))
+    :else (range (- current-page 2) (+ current-page 3))))
+
+
+(defn- container [narrow? child align]
+  (if narrow?
+    [:div {:style {:margin "0.25em 0"}} child]
```
zarsky-broad: `rem` plz (probably about 0.2)

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+    [:div {:style {:width "33%" :textAlign align}} child]))
+
+
+(defn- view-component [{:keys [page-number rows-per-page filtered-count total-count]}]
+  (let [right-num (min filtered-count (* page-number rows-per-page))
+        left-num (if (zero? right-num) 0 (inc (* (dec page-number) rows-per-page)))]
+    [:div {:style {:display "inline-flex"}}
+     [:b {:style {:marginRight "1ex"}} (str left-num " - " right-num)]
```
zarsky-broad: `rem` here too, and consider changing this from a `:b` to a `:strong`

src/cljs/main/broadfcui/common/table/body.cljs
zarsky-broad: Also, can we intercept double/triple clicks so they don't highlight the text? It's super distracting to me.

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+
+
+(defn- page-component [{:keys [filtered-count rows-per-page page-number page-selected]}]
+  (let [num-pages (js/Math.ceil (/ filtered-count rows-per-page))
+        allow-prev (> page-number 1)
+        allow-next (< page-number num-pages)]
+    (style/create-unselectable
+     :div {:style {:display "inline-flex" :alignItems "baseline"}}
+     [:div {:style {:display "inline-flex" :alignItems "baseline"
+                    :padding "0em 0.9em"
```
zarsky-broad: can be `:padding "0 0.75rem"` or thereabouts

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+                             (:border-light style/colors))
+                    :cursor (when allow-prev "pointer")}
+            :onClick (when allow-prev #(page-selected (dec page-number)))}
+      (icons/icon {:style {:alignSelf "center" :paddingRight "0.5rem"}} :angle-left)
+      "Prev"]
+     [:span {:style {:whiteSpace "nowrap"}}
+      (map (fn [n]
+             (let [selected? (= n page-number)]
+               [:div {:style {:textAlign "center"
+                              :paddingTop 5 :display "inline-block" :width 29 :height 24
```
zarsky-broad: Aaaagh magic numbers

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+                              :backgroundColor (when selected? (:button-primary style/colors))
+                              :color (if selected? "white" (:button-primary style/colors))
+                              :borderRadius (when selected? "100%")
+                              :cursor (when-not selected? "pointer")}
+                      :onClick (when-not selected? #(page-selected n))}
+                n]))
+           (create-page-range page-number num-pages))]
+     [:div {:style {:display "inline-flex" :alignItems "baseline"
+                    :padding "0em 0.9em"
```
zarsky-broad: `:padding "0 0.75rem"` again

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+                    :color (if allow-next
+                             (:button-primary style/colors)
+                             (:border-light style/colors))
+                    :cursor (when allow-next "pointer")}
+            :onClick (when allow-next #(page-selected (inc page-number)))}
+      "Next"
+      (icons/icon {:style {:alignSelf "center" :paddingLeft "0.5rem"}} :angle-right)])))
```
zarsky-broad, line 70: Do I recall you mentioning something about having first/last page buttons, maybe only if the page count is higher than 5? (perhaps `fa-angle-double-right`?)

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+
+
+(defn- rows-component [{:keys [rows-per-page per-page-options per-page-selected]}]
+  [:div {:style {:display "inline-flex" :alignItems "baseline"}}
+   "Display"
+   (style/create-identity-select
+    {:value rows-per-page
+     :style {:width 60 :margin "0em 1em"}
```
zarsky-broad: no unit on 0 and `rem`, thanks (I know you didn't originally write much of this, thanks for tolerating my cleanup demands…)

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+     :onChange #(per-page-selected (-> % .-target .-value js/parseInt))}
```
zarsky-broad: I am stunned that clojure has no built-in int parsing from strings. Well, perhaps not that stunned.

src/cljs/main/broadfcui/common/table/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+    [broadfcui.common.style :as style]
+    ))
+
+
+(def table-icon-size 16)
+
+(def clip-text {:whiteSpace "nowrap" :overflow "hidden" :textOverflow "ellipsis"})
+
+(defn tab [color]
+  {:borderRight (str "1px solid " (if (keyword? color) (color style/colors) color))
```
zarsky-broad: Do we want to put a default in here, too?

src/cljs/main/broadfcui/common/table/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+   :marginRight -1})
+
+(def table-light
+  {:table {:fontSize "0.8rem"}
+   :cell (merge clip-text {:padding "0.6rem 0 0.6rem 16px"})
+   :header-row {:color (:text-light style/colors)}
+   :body {:fontWeight 500}
+   :body-row (constantly {:borderTop style/standard-line})
+   :resize-tab {:borderRight style/standard-line :marginRight -1}})
```
zarsky-broad: Why not just use `tab` here?

src/cljs/main/broadfcui/common/table/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+
+(def clip-text {:whiteSpace "nowrap" :overflow "hidden" :textOverflow "ellipsis"})
+
+(defn tab [color]
+  {:borderRight (str "1px solid " (if (keyword? color) (color style/colors) color))
+   :marginRight -1})
+
+(def table-light
+  {:table {:fontSize "0.8rem"}
+   :cell (merge clip-text {:padding "0.6rem 0 0.6rem 16px"})
```
zarsky-broad: Where's 16px come from?

src/cljs/main/broadfcui/common/table/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+   :header-row {:color (:text-light style/colors)}
+   :body {:fontWeight 500}
+   :body-row (constantly {:borderTop style/standard-line})
+   :resize-tab {:borderRight style/standard-line :marginRight -1}})
+
+(def table-heavy
+  {:table {:fontSize "0.8rem" :fontWeight 500}
+   :cell clip-text
+   :header-row {:backgroundColor (:background-dark style/colors) :color "white"}
+   :header-cell {:padding "0.6rem 0 0.6rem 16px"}
```
zarsky-broad: Same question on 16px

src/cljs/main/broadfcui/common/table/style.cljs
```diff
@@ -0,0 +1,31 @@
...
+   :resize-tab (tab "#777")
+   :body-row (fn [{:keys [index]}]
+               {:backgroundColor (if (odd? index) "white" (:background-light style/colors))})
+   :body-cell {:padding "0.5rem 0 0.4rem 16px"}})
```
zarsky-broad: Same question on 16px

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+(def ^:private default-props
+  {:body {:empty-message "There are no rows to display."
+          :external-query-params #{}
+          :behavior {:reorderable-columns? true
+                     :fixed-column-count 0
+                     :sortable-columns? true
+                     :allow-no-sort? false
+                     :resizable-columns? true
+                     :filterable? true}}
+   :toolbar {:style {:display "flex" :alignItems "baseline" :marginBottom "1rem"}
```
zarsky-broad, line 27: Do we really want to allow overriding the toolbar's base style?

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+             :column-edit-button {:style {:marginRight "1rem"}
```
zarsky-broad, line 28: Same question: how much overriding do we want to allow here?

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
+(ns broadfcui.common.table.table
```
zarsky-broad, line 1: I presume this will take over at `broadfcui.common.table` before this is merged?

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+(def ^:private all-query-params #{:page-number :rows-per-page :filter-text :sort-column :sort-order})
+
+
+(react/defc Table
+  {:update-query-params
+   (fn [{:keys [state]} new-params]
+     (assert (subset? (set (keys new-params)) all-query-params) "Unknown key passed to :update-query-params")
+     (let [old-state (:query-params @state)
+           new-state (merge old-state new-params)
+           different? (not= old-state new-state)]
```
zarsky-broad, line 45: Would `changed?` or `modified?` be more precise here? Took me a sec to parse this out...

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+       (when different?
+         (swap! state assoc :query-params new-state))
+       different?))
+   :refresh-rows
+   (fn [{:keys [props state]}]
+     (swap! state assoc :loading? true)
+     ((-> props :body :data-source) {:columns (-> props :body :columns)
```
zarsky-broad, line 52: So, I presume `:data-source` is a function that transforms the data. If that's the case, I think it's an unclear name. The source of the data is the ajax, right? Wouldn't this be more like `:data-renderer`?

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+     (let [props (utils/deep-merge default-props props)
+           {:keys [rows column-display total-count filtered-count query-params]} @state
+           {:keys [body toolbar paginator]} props
+           {:keys [empty-message columns behavior external-query-params]} body
+           {:keys [fixed-column-count allow-no-sort?]} behavior
+           query-params (merge query-params (select-keys props external-query-params))
+           update-column-display #(swap! state assoc :column-display %)]
+       [:div {}
+        (when (:loading? @state)
+          [comps/DelayedBlocker {:banner "Loading..."}])
```
zarsky-broad: I'd love it if we didn't block the entire page while the data's loading. What if I realize I navigated to the wrong tab, etc.?

src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -0,0 +1,141 @@
...
+             [:div {:style (:style button-props)}
+              [ColumnEditButton
+               (assoc (utils/restructure columns column-display update-column-display fixed-column-count)
+                 :reorder-anchor (:anchor button-props)
+                 :button (:button button-props))]]))
+         (when (and (:filterable? behavior) (not (contains? external-query-params :filter-text)))
+           (let [filter-bar-props (:filter-bar toolbar)]
+             [:div {:style (:style filter-bar-props)}
+              [comps/TextFilter {:on-filter #(swap! state update :query-params assoc :filter-text %)}]]))
+         (list* (:items toolbar))]
```
zarsky-broad, line 108: Why do we need `list*` here? I'm a bit confused.

src/cljs/main/broadfcui/common/table/paginator.cljs
```diff
@@ -0,0 +1,92 @@
...
+    per-page-options)
+   "rows per page"])
+
+
+(defn paginator [props]
```
zarsky-broad, line 84: Why not make this a `defc` to match the other table parts?

dmohs (approved):
Love the new fixed column display.

src/cljs/main/broadfcui/common/table/utils.cljs
```diff
@@ -0,0 +1,82 @@
...
+  (or id header))
+
+
+(defn canonical-name [{:keys [header id]}]
```
zarsky-broad, line 13: When would this be used that `resolve-canonical-name ` couldn't?

src/cljs/main/broadfcui/common/table_utils.cljs
```diff
@@ -5,6 +5,7 @@
    [broadfcui.common :as common]
    [broadfcui.common.gcs-file-preview :refer [GCSFilePreviewLink]]
    [broadfcui.common.components :as comps]
+   [broadfcui.common.flex-utils :as flex]
```
zarsky-broad, line 8: `:refer`?

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -22,76 +24,73 @@
...
-                    extra-columns))
-         :pagination (react/call :pagination this)
-         :->row (fn [data]
-                  (->> extra-columns
-                       (concat [:library:indication :library:dataUseRestriction :library:numSubjects])
-                       (map data)
-                       (cons data)
-                       (cons data)
-                       vec))}]))
+         {:items [[:div {:style {:fontSize "112%"}}
```
zarsky-broad: 112%?

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -22,76 +24,73 @@
...
-                            :borderBottom (str "2px solid " (:border-light style/colors))}
-         :header-style {:padding "0.5em 0 0.5em 1em"}
-         :resizable-columns? true
-         :sortable-columns? true
-         :filterable? false
-         :resize-tab-color (:border-light style/colors)
-         :reorder-anchor :right
-         :reorder-style {:width "300px" :whiteSpace "nowrap" :overflow "hidden" :textOverflow "ellipsis"}
-         :reorder-prefix "Columns"
+       [Table
```
zarsky-broad: After clicking to reorder based on a column on this table, it's a moment before the view is refreshed, but there's no evidence that anything is happening during this time. Can we add something?

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -22,76 +24,73 @@
...
+                   [:span {:style {:fontWeight 700 :color (:text-light style/colors) :marginRight "0.5rem"}}
+                    "Matching Cohorts"]
+                   [:span {:style {:fontSize "80%"}}
+                    (let [total (or (:total @state) 0)]
+                      (str total
+                           " Dataset"
+                           (when-not (= 1 total) "s")
+                           " found"))]]
+                  flex/spring]
+          :style {:alignItems "flex-start" :marginBottom 7}
```
zarsky-broad: why 7px?

src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
```diff
@@ -185,30 +185,31 @@
...
+          [comps/Button {:text "Previous"
+                         :onClick (fn [_]
+                                    (if-let [prev-page (peek (:pages-stack @state))]
+                                      (swap! state #(-> %
+                                                        (assoc :page-num prev-page)
+                                                        (update :pages-stack pop)
+                                                        (dissoc :validation-error)))))
+                         :style {:width 80}
+                         :disabled? (zero? page-num)}]
+          (flex/strut 27)
```
zarsky-broad, line 203: 27?

src/webpack-deps.js
```diff
@@ -28,3 +28,4 @@ require('./styles/markdown.css');
 require('./styles/twitter-typeahead.css');
 require('./styles/foundation.scss');
 require('./styles/select2.scss');
+require('./styles/fixes.scss');
```
zarsky-broad: I think these are more shims for react than fixes, yes?

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: Something's not calculating right when I do that.

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: Any ideas about the double clicks? I don't see a way to block highlighting without blocking all text selection. Which people apparently want to be able to do.

src/cljs/main/broadfcui/common/table/body.cljs
zarsky-broad: In that case, `:flex (str "1 " width "px")`?

src/cljs/main/broadfcui/common/table/body.cljs
zarsky-broad: `(.preventDefault event)` perhaps?

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: I fished around for what felt right and ended up with 11px, and then 21 is what made it symmetrical (double the edge distance... minus one for some boundary condition reason)

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: Column sortability overrides this global default.  Check out the library page table, the left column isn't sortable.

src/cljs/main/broadfcui/common/table/body.cljs
zarsky-broad: Is there some way to canonize this behind a named variable?
(Also, any reason to keep `:top` here?)

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: Yeah I was trying that but I couldn't find the right div or something.

src/cljs/main/broadfcui/common/table/body.cljs
MattPutnam: Dunno why, but `:top` really needs to be there.  I'll pull the 11 into a private variable.

src/cljs/main/broadfcui/common/table/column_editor.cljs
MattPutnam: Mohs and I decided that it only makes sense for unmovable columns to be the leftmost, so there's a `:fixed-column-count` property on the table now.  This is dropping the fixed columns from the logic that sets All/None visibility.

src/cljs/main/broadfcui/common/flex_utils.cljs
MattPutnam: I once had Big Plans for these utilities.  Modeled after a combination of Java Swing's BoxLayout and Apple's Interface Builder.  I think there's other places that could be replaced by `flex/box`es.

src/cljs/main/broadfcui/common/table/paginator.cljs
MattPutnam: Pixel pushing to make those divs square.  I would love to know a better way to do this though!

src/cljs/main/broadfcui/common/table/body.cljs
zarsky-broad: Weird.

It'd be cool to have some sort of `(def offset-right -11)` and then width as 2*offset-1

src/cljs/main/broadfcui/common/table/paginator.cljs
MattPutnam: Let's leave that for a separate story.

src/cljs/main/broadfcui/common/table/paginator.cljs
MattPutnam: There totally is (`int`), we're just being dumb.  Fixed.

src/cljs/main/broadfcui/common/table/paginator.cljs
MattPutnam: Because it's stateless and has no need for any React lifecycle stuff.

src/cljs/main/broadfcui/common/table/style.cljs
MattPutnam: One of the old PixelMedia designs, I think. I'll clean it up.

src/cljs/main/broadfcui/common/table/table.cljs
MattPutnam: We're not going to convert every table as part of this PR, but it will move over once we finish migrating all of the tables.

src/cljs/main/broadfcui/common/table/table.cljs
MattPutnam: This base style, maybe not.  But the individual components, yes.  Check out how the library table gets the column reorder button to be on the right.

src/cljs/main/broadfcui/common/table/table.cljs
MattPutnam: `:data-source` is a function that gets the query params and returns the appropriate rows, either via ajax or via client-side manipulation.

src/cljs/main/broadfcui/common/table/table.cljs
MattPutnam: So the user can specify the additional toolbar items as a vector.  `list*` converts to a list so that rendering doesn't try to treat the toolbar items array as a single DOM element.

src/cljs/main/broadfcui/page/library/library_page.cljs
MattPutnam: Hmm... seems like the blocker isn't showing up.  Will investigate.  I'll also see about getting it only on the table body.

src/cljs/main/broadfcui/page/library/library_page.cljs
MattPutnam: Magic number to make it match something else... I forget what.

src/cljs/main/broadfcui/page/library/library_page.cljs
MattPutnam: Magic number to make the line under the table header line up with the line under the filter.

src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
MattPutnam: Ugh, yeah.  That matches the existing OKCancelForm (which in turn comes from PixelMedia, again), which I unfortunately had to copy/paste instead of reusing.  Major target for future refactoring.

src/webpack-deps.js
MattPutnam: So... `shims.scss`?  `react-shims.scss`?

src/cljs/main/broadfcui/common/table/utils.cljs
MattPutnam: I'll rename and add some docstrings, but `canonical-name` operates on a "resolved column" (i.e. the full column from the definitions), and `resolve-canonical-name` works from a column ID and the list of column definitions.

src/cljs/main/broadfcui/common/table/paginator.cljs
zarsky-broad: I thought that only worked for parsing from doubles etc, not from strings?

src/webpack-deps.js
zarsky-broad: I'd vote for `react-shims.scss`, yeah

src/cljs/main/broadfcui/page/library/library_page.cljs
MattPutnam: Done

src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
zarsky-broad: Want to put like a `;; REFACTOR` around these magic numbers when they appear?

zarsky-broad (approved):
If you'd be so kind as to leave a FIXME or similar by the magic numbers, I'm quite happy

PR #717, opened 2017-03-31
expand build script and add documentation

jacmrob:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


scripts/build.sh
```diff
@@ -1,12 +1,58 @@
...
-  broadinstitute/clojure-node \
-  bash -c 'lein with-profile deploy do clean, resource, cljsbuild once && npm install && npm run webpack -- -p'
+function clj_build() {
+    docker run --rm \
+      -w /work \
+      -v "$PWD":/work \
+      -v maven-cache:/root/.m2 \
+      -e npm_config_unsafe_perm="true" \
+      broadinstitute/clojure-node \
+      lein with-profile deploy do clean, resource, npm install, cljsbuild once
```
zarsky-broad: @dmohs teaches that this can mess up, which is the reason for the `bash -c` form.

scripts/build.sh
```diff
@@ -1,12 +1,58 @@
...
+
+}
+
+function docker_cmd()
+{
+    if [ $DOCKER_CMD = "build" ] || [ $DOCKER_CMD = "push" ]; then
+        echo "building docker image..."
+        GIT_SHA=$(git rev-parse ${GIT_BRANCH})
+        echo GIT_SHA=$GIT_SHA > env.properties  # for jenkins jobs
```
zarsky-broad: Is there a way to verify that we're in a Jenkins build before doing this? Or does it not matter?

scripts/build.sh
```diff
@@ -1,12 +1,58 @@
 #!/bin/bash
+# Single source of truth for building Firecloud-Ui.
+# @ Jackie Roberti and Isaac Zarsky
+#
+# Provide command line options to do one or several things:
+#   compile: compile clojure proj into /target
+#   -d | --docker : provide arg either "build" or "push", to build and push docker image
+# Jenkins build job should run with all options, for example,
+#   ./docker/build.sh compile -d push
```
zarsky-broad, line 9: 1. Will we ever _not_ want to compile? Why?
2. Why do we need the `-d` flag before `push` or `build`?
3. Should we set some help text to echo out if there are no arguments?

scripts/build.sh
```diff
@@ -1,12 +1,58 @@
...
+        fi
+    else
+        echo "Not a valid docker option!  Choose either build or push (which includes build)"
+    fi
+}
+
+# parse command line options
+DOCKER_CMD=
+GIT_BRANCH=${GIT_BRANCH:-$(git rev-parse --abbrev-ref HEAD)}  # default to current branch
+REPO=${REPO:-broadinstitute/$PROJECT}  # default to rawls docker repo
```
zarsky-broad: Why assign these to global variables? Why not just make them arguments to `docker_cmd`?

scripts/build.sh
jacmrob: 1. i suppose if you've already compiled the project, and just want to trigger the docker stuff.  i think it's good to have them separated.  the format is consistent across all the services, because some have different build steps that we may or may not want to do. 
2.  "-d" was to imply a docker command, bc we have multiple different ones supported
3.  help text would be good

scripts/build.sh
jacmrob: what is the `bash -c` form?

scripts/build.sh
jacmrob: it doesn't matter; creating a properties file is harmless

scripts/build.sh
jacmrob: because i am bad at bash :( that's a good idea

jacmrob: at least for `DOCKER_CMD` i think you're right.  for the other two i think they should be global in case we expand the script.

scripts/build.sh
zarsky-broad: As it was previously:
`bash -c 'lein with-profile deploy do clean, resource, cljsbuild once && npm install && npm run webpack -- -p'`

scripts/build.sh
zarsky-broad: 1. Cool, consistency wins.
2. Ok, if you think it's helpful/clearer.

scripts/build.sh
```diff
@@ -1,12 +1,56 @@
...
+      -e npm_config_unsafe_perm="true" \
+      broadinstitute/clojure-node \
+      lein with-profile deploy do clean, resource, npm install, cljsbuild once
+
+}
+
+function docker_cmd()
+{
+    DOCKER_CMD=$1
+    REPO=$2
```
zarsky-broad: If this is the only place repo is used, why not just build it here, rather than passing it in?

scripts/build.sh
```diff
@@ -1,12 +1,56 @@
...
+    if [ $DOCKER_CMD = "build" ] || [ $DOCKER_CMD = "push" ]; then
+        echo "building docker image..."
+        GIT_SHA=$(git rev-parse ${GIT_BRANCH})
+        echo GIT_SHA=$GIT_SHA > env.properties  # for jenkins jobs
+        docker $DOCKER_CMD  -t $REPO:${GIT_SHA:0:12} .
```
zarsky-broad: Shouldn't this be `docker build`? Otherwise we could wind up trying to push twice but never building, no?

scripts/build.sh
jacmrob: this is smart

scripts/build.sh
jacmrob: GOOD CATCH!

PR #720, opened 2017-04-03
convert discoverable by to a list

ahaessly:
the api was saving discoverable by as a list, but the UI was saving it as a single value. converting to list.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
+++ b/src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
@@ -9,8 +9,8 @@
     [broadfcui.endpoints :as endpoints]
     [broadfcui.page.workspace.summary.catalog.questions :refer [Questions]]
     [broadfcui.page.workspace.summary.library-utils :as library-utils]
-    [broadfcui.utils :as utils]
-    ))
+    [broadfcui.utils :as utils]))
+
 
 (defn- render-wizard-breadcrumbs [{:keys [library-schema page-num pages-seen]}]
   (let [pages (:wizard library-schema)]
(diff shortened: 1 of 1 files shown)


src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs
```diff
@@ -60,23 +60,27 @@
 
 (def ^:private ALL_USERS "All users")
 
+(defn- make-list [input]
+  ;;input may or maynot be a list, make it a list
+  (flatten (conj '() input)))
```
MattPutnam: `flatten` is a little dangerous, as it flattens recursively.  `(flatten [1 [2 [3] 4] 5])` -> `[1 2 3 4 5]`  Also, it looks like we don't need lists specifically, just any sequential type.  I would suggest:

```clojure
(defn- ensure-sequence [input]
  (if (sequential? input)
    input
    [input]))
```

And then replace `'()`s with `[]`s.

MattPutnam: Actually, it might make sense to make this nil-safe:
```clojure
(defn- ensure-sequence [input]
  (cond (sequential? input) input
        (nil? input) []
        :else [input]))

Comment (davidangb) [reply src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs]: Though, if the input really *was* `[1 [2 [3] 4] 5]`, `(sequential? )` would return true, and we'd attempt to pass the nested structure through ... resulting in bad data. So I agree that a generic `ensure-sequence` function should not flatten, but in this particular case we *do* want flatten.

I think it's a corner case so I don't feel strongly either way. Maybe we keep `ensure-sequence` generic, but we flatten in the location we call it from:
```
{:library:discoverableByGroups (flatten (make-list (get props :library:discoverableByGroups '())))}
```

File: src/cljs/main/broadfcui/page/workspace/summary/catalog/wizard.cljs [hunk 2]
```diff
@@ -60,23 +60,27 @@
...
+
 (react/defc DiscoverabilityPage
   {:validate (constantly nil)
    :get-initial-state
    (fn [{:keys [props]}]
-     (select-keys props [:library:discoverableByGroups :owner?]))
+     {:library:discoverableByGroups (make-list (get props :library:discoverableByGroups '()))})
```
davidangb: if `ensure-sequence` (née `make-list`) is nil-safe, you don't need the default value here.

davidangb (commented):
minor comments inline, 👍 overall!

davidangb (approved):
I notice you eliminated `flatten` - your call on that. 👍 lgtm!

davidangb:
👍

MattPutnam:
👍

PR #728, opened 2017-04-10
GAWB-1406: Workspace filter overhaul

MattPutnam:
Also fixes an untracked bug introduced in the table refactor--restoring the text of filter fields.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/workspaces_list.cljs
+++ b/src/cljs/main/broadfcui/page/workspaces_list.cljs
@@ -177,11 +177,9 @@
               :initial-width 132 :resizable? false
               :column-data :accessLevel
               :sort-by #(case % "OWNER" 0 "WRITER" 1 "READER" 2 "NO ACCESS" 3 4) :sort-initial :asc
-              :render (fn [accessLevel]
+              :render (fn [access-level]
                         [:div {:style {:paddingLeft 14}}
-                         (if (= accessLevel "PROJECT_OWNER")
-                           "Project Owner"
-                           (clojure.string/capitalize accessLevel))])}])
+                         (prettify access-level)])}])
           :behavior {:reorderable-columns? false}
           :style {:header-row {:color (:text-lighter style/colors) :fontSize "90%"}
                   :header-cell {:padding "0.4rem 0"}
--- a/src/cljs/main/broadfcui/page/workspaces_list.cljs
+++ b/src/cljs/main/broadfcui/page/workspaces_list.cljs
@@ -83,9 +83,6 @@
 (defn- get-workspace-description [ws]
   (not-empty (get-in ws [:workspace :attributes :description])))
 
-(defn- get-max-length [func workspaces]
-  (->> workspaces (map func) (map count) (apply max)))
-
 ;; An obnoxious amount of effort due to "PROJECT_OWNER" vs. "NO ACCESS"
 (defn- prettify [s]
   (as-> s $
@@ -137,9 +134,7 @@
    :render
    (fn [{:keys [props state this locals]}]
      (let [{:keys [workspaces nav-context]} props
-           {:keys [filters-expanded?]} @state
-           max-workspace-name-length (get-max-length get-workspace-name-string workspaces)
-           max-description-length (get-max-length get-workspace-description workspaces)]
+           {:keys [filters-expanded?]} @state]
        [Table
         {:ref "table" :persistence-key "workspace-table" :v 2
          :body
@@ -162,11 +157,11 @@
               :column-data column-data :as-text :status
               :render (fn [data] [StatusCell (utils/restructure data nav-context)])}
              {:id "Workspace" :header [:span {:style {:marginLeft 24}} "Workspace"]
-              :initial-width (min 500 (* max-workspace-name-length 10))
+              :initial-width 300
               :column-data column-data :as-text :name :sort-by :text
               :render (fn [data] [WorkspaceCell (utils/restructure data nav-context)])}
              {:id "Description" :header [:span {:style {:marginLeft 14}} "Description"]
-              :initial-width (max 200 (min 500 (* max-description-length 10)))
+              :initial-width 250
               :column-data get-workspace-description
               :render (fn [description]
                         [:div {:style {:paddingLeft 14}}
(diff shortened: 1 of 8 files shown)


src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -84,157 +87,175 @@
...
-                          {:text "Complete" :pred #(= "Complete" (:status %))}
-                          {:text "Running" :pred #(= "Running" (:status %))}
-                          {:text "Exception" :pred #(= "Exception" (:status %))}]
+   (fn [{:keys [props state this]}]
+     (let [{:keys [workspaces nav-context]} props
+           {:keys [filters-expanded?]} @state
+           max-workspace-name-length (get-max-length get-workspace-name-string workspaces)
+           max-description-length (get-max-length get-workspace-description workspaces)]
+       [Table
+        {:ref "table" :persistence-key "workspace-table" :v 2
```
davidangb: use `:v VERSION` instead?

src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -84,157 +87,175 @@
...
+              :render (fn [date]
+                        [:div {:style {:paddingLeft 14}} (common/format-date date)])}
+             {:id "Access Level" :header [:span {:style {:marginLeft 14}} "Access Level"]
+              :initial-width 132 :resizable? false
+              :column-data :accessLevel
+              :sort-by #(case % "OWNER" 0 "WRITER" 1 "READER" 2 "NO ACCESS" 3 4) :sort-initial :asc
+              :render (fn [accessLevel]
+                        [:div {:style {:paddingLeft 14}}
+                         (if (= accessLevel "PROJECT_OWNER")
+                           "Project Owner"
```
davidangb: use `prettify` instead?

src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -84,157 +87,175 @@
...
+        filter/area
+        {:style {:float "left" :margin "0 1rem 1rem 0" :width 175}}
+        (filter/section
+         {:title "Tags"
+          :content (react/create-element
+                    [comps/TagAutocomplete {:ref "tag-autocomplete"
+                                            :tags (filters "Tags")
+                                            :show-counts? false
+                                            :on-change #(swap! state update :filters assoc "Tags" %)}])
+          :on-clear #((@refs "tag-autocomplete") :set-tags [])})
```
davidangb: in the workspace listing, the tag autocomplete control should use the set of tags returned *on the list of workspaces for the current user*, not the set of tags returned from /api/workspaces/tags.

davidangb (changes requested):
Some comments inline. Also, from testing this myself:
* if I enter some characters into the tag control, I always get a first suggestion of "undefined"
* selecting Published=published finds 86 workspaces, even though library reindex only finds 83 workspaces. I don't know why this is happening.
* the text filter updates the pagination control to say e.g. "1 - 20 of 72 results (filtered from 2803 total)" ... should the filters do the same? The previous filter did not, but with the UX changing, PO might want this ...
* I'm surprised PO hasn't asked to add a Tags column to the table.
* if I add a tag as a filter, then reload the page - or switch to library and then back to workspaces - the tag filter _is_ properly initialized with my previous tag. However, if I use the delete key to remove that tag, instead of removing the tag, the control changes the tag "badge" into plain text.
* upon first load (or after clearing local storage), the Access Level column of the table is scrolled off the right side of the page for me.

davidangb:
I found the issue with the UI filter finding 86 published workspaces, where the backend Scala filter only finds 83. There are three workspaces - including broad-dsde-dev/GAWB-616-failedWorkflows which you are an owner of - that have a `library:published` attribute with a _string_ value of "true". The UI filter finds these, whereas Scala does not.

src/cljs/main/broadfcui/page/workspaces_list.cljs
MattPutnam: There are two versions here: the page, and the table (and I just noticed that the page persistence key is still "workspace-table-types" -- do we want to do something about that?).  We could tie them together for clarity though.

src/cljs/main/broadfcui/page/workspaces_list.cljs
MattPutnam: Is that going to be a new orch endpoint?

rushtong: David's understanding was mine as well ... sum up all of the tags in each of the workspaces returned from the workspace list and use that as the universe of autocomplete tags when filtering workspaces.

rushtong:
Generic comment on pagination:
* Navigate to any page > 1 in your workspace list
* Check some facet (published in my case) that limits the known list to less than the number displayed per page.
* The behavior is confusing. All rows go away, a message with `There are no rows to display.`, but you can navigate back to page 1 and see the rows. 

Can we drop the pagination to something that makes sense when the results get filtered?

rushtong:
**Tag-list reloading strangeness.** I don't know if this is our library working the way it does, or something else, but the effect is confusing.
* Enter a tag into the text box 
* Immediately delete without moving the mouse (using the delete key)
* I get a message in the drop down, "The results could not be loaded"

That error persists for some variable number of seconds if I don't move the mouse or click out of the text box. Sometimes it refreshes almost immediately, sometimes for many seconds. Clicking outside the box, then back in again will trigger an immediate refresh.

**Edit:** I can not reliably reproduce this. It seemed to happen more often with some facets selected.

rushtong (changes requested):
I don't have any code feedback, just functionality suggestions in other comments.

rushtong (approved):
👍 LG, WG!

davidangb:
Code looks good and works as expected. 👍 on the code level.

I have UX concerns - which really isn't my department, but here goes:
* I love the redesign of namespace/name!
* The tag control allows you to enter arbitrary tags - including tags that don't exist. This doesn't make sense in terms of the tag control.
* By default (i.e. after clearing local storage), the Workspace column is cut off, but the Access Level column has lots of extra whitespace. The Description column also has whitespace - I think this could be optimized. If you need to shrink, I bet you could drop the font size just a little.
* I think the text filter control should be moved into the grey section of the left column, same way it works in Library. I realize this complicates the hide/show filters feature. Personally, I'd be fine always showing the filters.

davidangb (commented):
added UX feedback. Attempting to change this review from "Request Changes" to "Comment".

davidangb (approved):
👍 for code, did not test

PR #746, opened 2017-04-19
GAWB-1648 Added class names for registration page tests

breilly2:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/header.cljs
+++ b/src/cljs/main/broadfcui/header.cljs
@@ -41,7 +41,8 @@
    {:label [:div {:style {:borderRadius 2
                           :backgroundColor (:background-light style/colors)
                           :color "#000" :textDecoration "none"
-                          :padding "0.5rem" :border style/standard-line}}
+                          :padding "0.5rem" :border style/standard-line}
+                  :className "accountDropdown"}
             (-> auth2 (.-currentUser) (.get) (.getBasicProfile) (.getEmail))
             [:div {:style {:display "inline-block" :marginLeft "1em" :fontSize 8}} "▼"]]
     :width :auto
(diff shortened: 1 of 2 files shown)


zarsky-broad (commented):
Putting down what we discussed: `data-` prop selectors

zarsky-broad (approved):

zarsky-broad:
Incidentally, @dmohs may have some thoughts on camel case vs snake case

src/cljs/main/broadfcui/header.cljs
```diff
@@ -41,7 +41,8 @@
    {:label [:div {:style {:borderRadius 2
                           :backgroundColor (:background-light style/colors)
                           :color "#000" :textDecoration "none"
-                          :padding "0.5rem" :border style/standard-line}}
+                          :padding "0.5rem" :border style/standard-line}
+                  :data-test-id "accountDropdown"}
```
dmohs: Isaac is right. I'd prefer "account-dropdown" unless there is a particular reason to use camel case for these.

breilly2: I can change this here.

In profile.cljs, I reused `key` which is already camelCase, e.g. "firstName", instead of adding a parameter to `:render-field` and `:render-nested-field`. Do you think it's worth adding a parameter specifically for the `data-test-id` property so that the two are decoupled? That would have the happy side-effect that the value for `data-test-id` and can be kebab-case.

dmohs: I like the concision of reusing the key. I can live with camel-case if it flows naturally from the data, as it does in that case. I'd like to see language-idiomatic names when we create them.

On another project, I convert keys to kebab-case as part of the JSON parsing and conversion, and back to the server's preferred format on transmission. I've found this to be quite beneficial in making my code more internally consistent and nicer to work with. If we decide to go this route, these keys will naturally move.

breilly2: I have reservations about converting keys, but that's a separate concern. I will kebab-case this one.

PR #738, opened 2017-04-14
GAWB-1898: Prevent XSS in catalog wizard

MattPutnam:
- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * Add notes on what you've tested
- [x] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [x] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [x] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [x] Tell the tech lead (TL) that the PR exists if they wants to look at it
- [x] Anoint a lead reviewer (LR). **Assign PR to LR**
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Product Owner** sign off
- [x] **Assign to submitter** to finalize
- [ ] **Submitter**: Verify all tests go green, including CI tests
- [x] **Submitter**: Squash commits and merge to develop
- [x] **Submitter**: Delete branch after merge
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

PR #752, opened 2017-04-20
GAWB-1926: support structured highlights returned by autosuggest API

davidangb:
requires broadinstitute/firecloud-orchestration#425. That other PR changes the response payload for Library typeahead suggestions. This PR accommodates that change, and encodes the final output to prevent XSS attacks.

I'm not super-psyched that the Library suggestions encode via `clojure.string/escape` while other typeaheads rely on `jQuery.text()`. However, the jQuery approach felt pretty awkward here, given the highlighted-term replacement that's happening. Happy to hear feedback on approach!

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/library/library_page.cljs
+++ b/src/cljs/main/broadfcui/page/library/library_page.cljs
@@ -154,6 +154,14 @@
                        ((:update-aggregates props) aggregations)))
                    (on-done {:error status-text})))}))))))})
 
+(defn encode [text]
+  (clojure.string/escape text {\" "&quot;" \& "&amp;" \< "&lt;", \> "&gt;", \\ "&#x27;" \/ "&#x2F;"}))
+
+(defn highlight-suggestion [suggestion highlight]
+  (if (some? highlight)
+    (clojure.string/replace (encode suggestion) (encode highlight) (str "<strong>" (encode highlight) "</strong>"))
+    (encode suggestion)))
+
 (react/defc SearchSection
   {:get-filters
    (fn [{:keys [props]}]
@@ -186,11 +194,15 @@
                                                :from 0
                                                :size 10}))))}
         :typeaheadDisplay (fn [result]
-                            ;; we intentionally do not encode the result value here, because it is already
-                            ;; encoded from the server.
-                            (.text (js/$ (str "<div>" (aget result "value") "</div>"))))
+                            (let [sugg-object (js->clj (aget result "value"))
+                                  suggestion (get sugg-object "suggestion")]
+                              (.text (js/$ (str "<div>" suggestion "</div>")))))
         :typeaheadSuggestionTemplate (fn [result]
-                                       (str "<div style='textOverflow: ellipsis; overflow: hidden; font-size: smaller;'>" (aget result "value") "</div>"))}]])})
+                                       (let [sugg-object (js->clj (aget result "value"))
+                                             suggestion (get sugg-object "suggestion")
+                                             highlight (get sugg-object "highlight")
+                                             display (highlight-suggestion suggestion highlight)]
+                                         (str "<div style='textOverflow: ellipsis; overflow: hidden; font-size: smaller;'>" display "</div>")))}]])})
 
 (react/defc FacetCheckboxes
   {:render
--- a/src/cljs/main/broadfcui/page/library/library_page.cljs
+++ b/src/cljs/main/broadfcui/page/library/library_page.cljs
@@ -195,8 +195,10 @@
                                                :from 0
                                                :size 10}))))}
         :typeaheadDisplay (fn [result]
-                            (let [suggestion (aget result "value" "suggestion")]
-                              (.text (js/$ (str "<div>" suggestion "</div>")))))
+                            ;; underlying typeahead library uses the result of this function
+                            ;; via $input.val(x), which is safe from xss. So we explicitly
+                            ;; do not want to encode anything here.
+                            (aget result "value" "suggestion"))
         :typeaheadSuggestionTemplate (fn [result]
                                        (let [suggestion (aget result "value" "suggestion")
                                              highlight (aget result "value" "highlight")
--- a/src/cljs/main/broadfcui/page/library/library_page.cljs
+++ b/src/cljs/main/broadfcui/page/library/library_page.cljs
@@ -155,10 +155,11 @@
                    (on-done {:error status-text})))}))))))})
 
 (defn encode [text]
+  ;; character replacements modeled after Lucene's SimpleHTMLEncoder.
   (clojure.string/escape text {\" "&quot;" \& "&amp;" \< "&lt;", \> "&gt;", \\ "&#x27;" \/ "&#x2F;"}))
 
 (defn highlight-suggestion [suggestion highlight]
-  (if (some? highlight)
+  (if (not (clojure.string/blank? highlight))
     (clojure.string/replace (encode suggestion) (encode highlight) (str "<strong>" (encode highlight) "</strong>"))
     (encode suggestion)))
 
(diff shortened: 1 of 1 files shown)


src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -154,6 +154,14 @@
                        ((:update-aggregates props) aggregations)))
                    (on-done {:error status-text})))}))))))})
 
+(defn encode [text]
+  (clojure.string/escape text {\" "&quot;" \& "&amp;" \< "&lt;", \> "&gt;", \\ "&#x27;" \/ "&#x2F;"}))
```
dmohs: Maybe a short comment above this line about why these specific characters. I don't think quotes need escaping, but if they do, I'd imagine single quotes need to be there as well.

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -186,11 +194,15 @@
                                                :from 0
                                                :size 10}))))}
         :typeaheadDisplay (fn [result]
-                            ;; we intentionally do not encode the result value here, because it is already
-                            ;; encoded from the server.
-                            (.text (js/$ (str "<div>" (aget result "value") "</div>"))))
+                            (let [sugg-object (js->clj (aget result "value"))
+                                  suggestion (get sugg-object "suggestion")]
```
dmohs: Could just be:
```cljs
suggestion (aget result "value" "suggestion")
```

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -186,11 +194,15 @@
...
+                              (.text (js/$ (str "<div>" suggestion "</div>")))))
```
dmohs: This still scares me a smidge. Can this be:
```cljs
(js/$ (str "<div>" (encode suggestion) "</div>"))
```

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -186,11 +194,15 @@
...
         :typeaheadSuggestionTemplate (fn [result]
-                                       (str "<div style='textOverflow: ellipsis; overflow: hidden; font-size: smaller;'>" (aget result "value") "</div>"))}]])})
+                                       (let [sugg-object (js->clj (aget result "value"))
```
dmohs: Similarly, if you want:
```cljs
suggestion (aget result "value" "suggestion")
highlight (aget result "value" "highlight")
```

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -154,6 +154,14 @@
                        ((:update-aggregates props) aggregations)))
                    (on-done {:error status-text})))}))))))})
 
+(defn encode [text]
+  (clojure.string/escape text {\" "&quot;" \& "&amp;" \< "&lt;", \> "&gt;", \\ "&#x27;" \/ "&#x2F;"}))
+
+(defn highlight-suggestion [suggestion highlight]
+  (if (some? highlight)
```
dmohs: Worth checking for empty string too?

src/cljs/main/broadfcui/page/library/library_page.cljs
MattPutnam: Do we even need to do all of this jquery packing/unpacking with the new result format?

davidangb: we *can* do `(js/$ (str "<div>" (encode suggestion) "</div>"))`, and it seems to work for the end user ... but it feels like we should use *either* `jQuery.text` or `encode`, and not both.

davidangb:
updated to reflect PR feedback - one open question on `jQuery.text` vs. `encode` - unfortunately that conversation is flagged as outdated ... read it here: https://github.com/broadinstitute/firecloud-ui/pull/752#discussion_r112497375

src/cljs/main/broadfcui/page/library/library_page.cljs
davidangb: @matt - we still need to do at least one of encode or jquery.

dmohs: The call to `.text()` is weird because it strips out the HTML, which means we add it with the `div` elements and then immediately remove it. I guess I would prefer either:
```cljs
(-> (str "<div>" (encode suggestion) "</div>") js/$)
```
or
```cljs
(-> suggestion js/$ (js-invoke "text"))
```
(no need to use my syntax, just giving an example)

dmohs: Actually, this came into my mind over the weekend. I recall that tag-stripping libraries are notorious for security bugs, so I would much prefer that we always encode and never attempt to strip.

davidangb: 1. jQuery, internally, does not do tag-stripping. It looks like it relies on `textContent` to do browser-native HTML-to-text conversion. (see: https://github.com/jquery/jquery/blob/2.2-stable/external/sizzle/dist/sizzle.js#L978)

2. I think you may be misreading what the jquery text line is doing?
```
(.text (js/$ (str "<div>" suggestion "</div>")))
```
should be equivalent to: 
```
$("<div>" + suggestion + "</div>").text();
```
right? That statement creates - but does not append to the DOM - a \<div\> element, and then extracts its text content and returns that text content as a string.

3. if you're uncomfortable with the `.text()` approach, we'll need to update the impl from #738 too.

In general, I'm *more* prone to use a third-party library, especially a mature one like jQuery, than write our own encoder; that's why I raised a caveat about a hand-written encoder right up front!

dmohs: I can't really tell how this particular value is used, but it seems to me the returned string needs to be sanitized. For example:
```javascript
$('<div><script>bad</script></div>').text() // -> 'bad'
```
but I can just do this:
```javascript
$('<div>&lt;script&gt;bad&lt;/script&gt;</div>').text() // -> '<script>bad</script>'
```

If the user of this value will be escaping it, then just a plain `suggestion` should be sufficient.

dmohs: And if they won't be escaping it, `(encode suggestion)` is what I would trust.

davidangb: I see the path to a vulnerability now, thank you - lemme take this back to rework.

src/cljs/main/broadfcui/page/library/library_page.cljs
```diff
@@ -186,11 +195,15 @@
                                                :from 0
                                                :size 10}))))}
         :typeaheadDisplay (fn [result]
-                            ;; we intentionally do not encode the result value here, because it is already
-                            ;; encoded from the server.
-                            (.text (js/$ (str "<div>" (aget result "value") "</div>"))))
+                            ;; underlying typeahead library uses the result of this function
+                            ;; via $input.val(x), which is safe from xss. So we explicitly
+                            ;; do not want to encode anything here.
+                            (aget result "value" "suggestion"))
```
dmohs, line 201: Perfect. Great comment.

davidangb:
@dmohs @MattPutnam ready for yet another review.

I spent a bunch of time testing and checking the source of the typehead library. What I settled on is that the typeahead library uses the result of `:typeaheadDisplay` as the argument to `val()` (see https://github.com/twitter/typeahead.js/blob/master/dist/typeahead.bundle.js#L1482 and http://api.jquery.com/val/#val2). Setting the value of an input control is natively safe from xss, so we should NOT encode anything here. I've removed the encoding for that part, and haven't changed the `:typeaheadSuggestionTemplate` part since your last review.

MattPutnam (approved):
👍 LGTM

PR #767, opened 2017-04-28
GAWB-1342: Create new method by 'duplicating'

MattPutnam:
Also quite a bit of code cleanup, converting the methods table to the new table, and redesigning the method detail view a bit to accommodate the new button.

The old table implementation had the option to add filter groups (in this case all/methods/configs).  In converting the methods table, I didn't add this to the new table component, but rather implemented it externally.  I'd like to eventually get rid of that design and do something like what the new workspace list does instead.  Keeping the filter groups external will facilitate/encourage this.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/components.cljs
+++ b/src/cljs/main/broadfcui/common/components.cljs
@@ -181,56 +181,57 @@
    (fn [{:keys [refs]}]
      {"methodVersion" (int (common/get-text refs "snapshotId"))})
    :render
-   (fn [{:keys [props refs state this]}]
+   (fn [{:keys [props state this]}]
      [:div {} (when-let [wdl-parse-error (:wdl-parse-error props)] (style/create-server-error-message wdl-parse-error))
       (let [entity (:entity props)
-            editing? (:editing? props)
-            make-field
-            (fn [entity key label dropdown? & [render]]
-              [:div {}
-               [:span {:style {:fontWeight 500 :width 100 :display "inline-block" :paddingBottom "0.3em"}} label]
+            config? (contains? entity :method)]
+        [:div {:style {:backgroundColor (:background-light style/colors)
+                       :borderRadius 8 :border style/standard-line
+                       :padding "1rem"}}
+         (this :render-details entity)
+         [:div {:style {:paddingTop "0.5rem"}}
+          [:span {:style {:fontWeight 500 :marginRight "1rem"}} (if config? "Referenced Method:" "WDL:")]
+          (style/create-link {:text (if (:payload-expanded @state) "Collapse" "Expand")
+                              :onClick #(swap! state assoc :payload-expanded (not (:payload-expanded @state)))})]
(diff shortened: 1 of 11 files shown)


src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -777,3 +777,41 @@
             :onClick #(swap! state assoc :collapsed? (not (:collapsed? @state)))})
           body]
          body)))})
+
+(react/defc FilterGroupBar
```
davidangb: could this `FilterGroupBar` be aligned with the FilterGroupBar in table_utils.cljs? There's a lot of duplicated code between the two.

davidangb (commented):
One comment on code inline, not much else to say syntax-wise.

Regarding functionality:
1. Should we support duplicating configs as well as methods? Currently, we do ... is that correct?
2. This PR doesn't match exactly the details of the Jira ticket: "Copy of {foo}" vs. "{foo} copy"; "Upload" vs. "Create New Method". Could you please either update the PR or update the Jira ticket? Otherwise you'll get rejected during QA.
3. If I select "Duplicate method," fill in a namespace, then click "Upload" ... the copy is saved, but I remain on the original method. Looks like the jira ticket specifies I should be brought to my new copy instead.

zarsky-broad:
There's a plan in the [mid-term] future to totally overhaul this table, so I wouldn't worry too much about the whole filter thing for now. Just FYI.

src/cljs/main/broadfcui/page/method_repo/method_config_importer.cljs
```diff
@@ -45,32 +48,43 @@
...
-                                                                      :on-delete (:on-delete props)}])}]]
-        (clear-both)])
-     [:div {:style {:border style/standard-line
-                    :backgroundColor (:background-light style/colors)
-                    :borderRadius 8 :padding "1em" :marginTop "1em"}}
-      [:div {:style {:fontSize "120%" :marginBottom "0.5em"}}
-       (if workspace-id "Import as:" "Export to Workspace as:")]
-      (map
+       [:div {:style {:flex "0 0 290px" :paddingRight "1rem"}}
+        (when-not config?
```
davidangb: right now, possible entities in the methods repo are "Workflow", "Task", and "Configuration". We're trying to do away with Tasks, but they still exist ... and this check allows duplication of both Tasks and Workflows. I think - and it's a PO call - we might only want to allow duplication on Workflows.

davidangb (commented):
one last functional comment inline, looking good otherwise! If PO approves as-is - including duplication of Tasks - then 👍 . If PO doesn't want duplication of Tasks, then this needs a revision.

davidangb (approved):
👍 ! Didn't test the last two commits; just read the code (and it's small code)

PR #768, opened 2017-05-02
DSDEQA-80: Adding Data Properties for automated tests

ansingh7115:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/components.cljs
+++ b/src/cljs/main/broadfcui/common/components.cljs
@@ -580,6 +580,7 @@
                            :fontSize "106%" :fontWeight 500 :textDecoration "none"
                            :color (:button-primary style/colors)}
                    :href "javascript:;"
+                   :data-test-id "cancel-button"
                    :onClick modal/pop-modal
                    :onKeyDown (common/create-key-handler [:space :enter] modal/pop-modal)}
                cancel-text])
--- a/src/cljs/main/broadfcui/common/components.cljs
+++ b/src/cljs/main/broadfcui/common/components.cljs
@@ -172,6 +172,7 @@
                       :color (if heavy? "#fff" color)
                       :border (when-not heavy? style/standard-line)
                       :borderRadius 5}
+              :data-test-id (:data-test-id props)
               :onClick (if disabled? (create-error-message disabled?) (:onClick props))}
         (icons/icon {:style {:padding "0 20px" :borderRight style/standard-line} :className "fa-fw"} (:icon props))
         [:div {:style {:textAlign "center" :margin "auto"}}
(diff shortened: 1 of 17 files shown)


src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -49,7 +49,7 @@
      {:color (:button-primary style/colors)})
    :render
    (fn [{:keys [props]}]
-     (let [{:keys [color icon href disabled? onClick text style class-name]} props]
+     (let [{:keys [color icon href disabled? onClick text style class-name data-test-id]} props]
```
breilly2, line 52: As I mentioned on Friday, I have a PR out (#786) that takes a different approach to updating the `Button` component. Conveniently, the difference is isolated here so invokers of the `Button` component will be the same no matter which approach sticks.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -147,7 +148,8 @@
                     :alignItems "center" :justifyContent "center"}}
       (:icon props)
       [:span {:style {:marginLeft "1em" :fontSize "125%" :fontWeight 400
-                      :verticalAlign "middle"}}
+                      :verticalAlign "middle"}
+              :data-test-id "submission-status"}
```
breilly2, line 152: I'm thinking that `"submission-status"` should not be hard-coded here given this component's general namespace/name of `broadfcui.common.components/StatusLabel`. Also, `StatusLabel` is used in 2 places. Maybe it's for submission status in both cases, but if there's a chance that they could both be rendered at the same time they should have different IDs.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -406,10 +409,11 @@
        [:div {:style {:display "inline-flex" :width width}}
         (style/create-search-field
           {:ref "filter-field" :autoSave "true" :results 5 :autofocus "true"
+           :data-test-id (str (:table-name props) "-search-field")
```
breilly2, line 415: Since `TextFilter` is a general component, the property used to construct `data-test-id` should probably not be specific to a table. This component could be used to filter text in other contexts.

Interesting. I see why you're building the actual ID here... so you can have both a `-field` and `-button`. Still, I think you could just take a `data-test-id` prop and apply it directly to the text input since that's the primary element for this component. If you also want an ID on the button, you could construct that with `(str (:data-test-id props) "-button")`.

As a side note, `data-test-id` could be explicitly pulled from props for the text field or this component could be modified to allow passthrough attributes (which I guess would NOT be applied to the button... or attributes to be applied to the button could be in a separate map passed to `TextFilter`.

breilly2 (commented):
Not done yet... just done for now, but I wanted to post the feedback that I have so far.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -572,6 +576,7 @@
                            :fontSize "106%" :fontWeight 500 :textDecoration "none"
                            :color (:button-primary style/colors)}
                    :href "javascript:;"
+                   :data-test-id "cancel-button"
```
breilly2, line 583: In general, we should be cautious about static test IDs in reusable components, but it should be okay in this case because it's a modal and therefore there should only be one (unless there are stacked modals, but I think we may be okay there too because only the topmost one will be accessible? probably worth experimenting with that).

src/cljs/main/broadfcui/common/input.cljs
```diff
@@ -25,6 +25,7 @@
    (fn [{:keys [state props]}]
      (style/create-text-field
       (merge {:ref "textfield"
+              :data-test-id (:data-test-id props)
```
breilly2, line 28: I didn't find this to be necessary because `props` gets merged in with this map (minus `:ref`, `:style`, and `:onChange`) and passed to `create-text-field`.

src/cljs/main/broadfcui/common/style.cljs
```diff
@@ -89,7 +89,8 @@
 
 (defn create-message-well [message]
   [:div {:style {:textAlign "center" :backgroundColor (:background-light colors)
-                 :padding "1em 0" :borderRadius 8}}
+                 :padding "1em 0" :borderRadius 8}
+         :data-test-id "empty-message"}
```
breilly2, line 110: Either `data-test-id` could be passed in to `create-message-well` or the ID here could have a more generic name... then tests can grab it and check the text contents.

PR #769, opened 2017-05-04
GAWB-1908: Upgrade to jQuery 3

rushtong:
Addresses https://broadinstitute.atlassian.net/browse/GAWB-1908

Minor code updates to deal with jquery3+react warning about nil value properties on input fields. 

Notes:
* You can add a migration plugin that will help diagnose deprecations. 
* in `package.json`, add `"jquery-migrate": "^3.0.0"` to the dependencies
* in `webpack-dep.js`, add `require('jquery-migrate');` after `window.$ = window.jQuery;`
* Usage of the plugin can be found here: https://github.com/jquery/jquery-migrate/blob/master/warnings.md
* jQuery3 upgrade notes are here:  https://jquery.com/upgrade-guide/3.0/

---

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "font-awesome": "~4.7.0",
     "foundation-sites": "~6.3.0",
     "github-markdown-css": "~2.4.1",
-    "jquery": "^3.0.0",
+    "jquery": "^3.2.1",
     "jquery-ui": "^1.12",
     "marked": "~0.3.6",
     "select2": "~4.0"
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "font-awesome": "~4.7.0",
     "foundation-sites": "~6.3.0",
     "github-markdown-css": "~2.4.1",
-    "jquery": "^2.0",
+    "jquery": "^3.0.0",
     "jquery-ui": "^1.12",
     "marked": "~0.3.6",
     "select2": "~4.0"
--- a/src/cljs/main/broadfcui/common/style.cljs
+++ b/src/cljs/main/broadfcui/common/style.cljs
@@ -62,11 +62,15 @@
 (defn create-form-label [text]
   [:div {:style {:marginBottom "0.16667em" :fontSize "88%"}} text])
 
+;; jQuery3+React warns about nil :value elements. Filter.
+(defn- filter-nil-props [props]
+   (into {} (filter #(not (nil? (val %))) props)))
+
 (defn create-text-field [props]
-  [:input (deep-merge {:type "text" :style input-text-style} props)])
+  [:input (deep-merge {:type "text" :style input-text-style} (filter-nil-props props))])
 
 (defn create-search-field [props]
-  [:input (deep-merge {:type "search" :style (assoc input-text-style :WebkitAppearance "none")} props)])
+  [:input (deep-merge {:type "search" :style (assoc input-text-style :WebkitAppearance "none")} (filter-nil-props props))])
 
 (defn create-text-area [props]
   [:textarea (deep-merge {:style input-text-style} props)])
--- a/src/cljs/main/broadfcui/common/style.cljs
+++ b/src/cljs/main/broadfcui/common/style.cljs
@@ -62,15 +62,19 @@
 (defn create-form-label [text]
   [:div {:style {:marginBottom "0.16667em" :fontSize "88%"}} text])
 
-;; jQuery3+React warns about nil :value elements. Filter.
-(defn- filter-nil-props [props]
-   (into {} (filter #(not (nil? (val %))) props)))
+;; jQuery3+React warns about nil :value elements. Replace with empty string.
+(defn- replace-nil-value-key [props]
+  (reduce-kv (fn [m k v]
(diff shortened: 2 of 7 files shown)


src/cljs/main/broadfcui/page/workspace/method_configs/launch_analysis.cljs
```diff
@@ -69,16 +83,7 @@
...
-     (style/create-text-field {:placeholder "leave blank for default"
-                               :style {:width "100%"
-                                       :backgroundColor (when disabled (:background-light style/colors))}
-                               :disabled disabled
-                               :value (if disabled
-                                        "Disabled - selected entity is of root entity type"
-                                        (:expression @state))
-                               :onChange #(let [text (-> % .-target .-value clojure.string/trim)]
-                                            (swap! state assoc :expression text))}))
+   [DefineExpressionField props state]
```
rushtong: Moved to its own component because jquery3+react complained about updates from uncontrolled to controlled components: https://facebook.github.io/react/docs/forms.html#controlled-components
Does not occur under jquery2+react

src/cljs/main/broadfcui/common/style.cljs
```diff
@@ -62,11 +62,15 @@
 (defn create-form-label [text]
   [:div {:style {:marginBottom "0.16667em" :fontSize "88%"}} text])
 
+;; jQuery3+React warns about nil :value elements. Filter.
```
rushtong: Required because jquery3+react complained about nil values being assigned to properties on an input element. Does not occur under jquery2+react:
```
`Warning: `value` prop on `input` should not be null. Consider using the empty string to clear the component or `undefined` for uncontrolled components.`
```

rushtong: See later commit. Filtering `nil` caused https://facebook.github.io/react/docs/forms.html#controlled-components errors

src/cljs/main/broadfcui/page/workspace/method_configs/launch_analysis.cljs
rushtong: Converting `nil` `:value` properties to empty string resolved this issue.

package.json
```diff
@@ -9,7 +9,7 @@
     "font-awesome": "~4.7.0",
     "foundation-sites": "~6.3.0",
     "github-markdown-css": "~2.4.1",
-    "jquery": "^2.0",
+    "jquery": "^3.0.0",
```
davidangb: this should get us 3.2.1 ... but why not specify `^3.2.1` here?

src/cljs/main/broadfcui/common/style.cljs
```diff
@@ -62,11 +62,19 @@
 (defn create-form-label [text]
   [:div {:style {:marginBottom "0.16667em" :fontSize "88%"}} text])
 
+;; jQuery3+React warns about nil :value elements. Replace with empty string.
+(defn- replace-nil-value-key [props]
+  (reduce-kv (fn [m k v]
+               (if (and (= k :value) (nil? v))
+                 (assoc m k "")
+                 (assoc m k v)))
+             {} props))
```
davidangb, line 71: does it also work to remove the `value` property altogether? i.e. instead of
```
<input placeholder="test" disabled="false" value="" />
```
we'd have
```
<input placeholder="test" disabled="false" />
```

davidangb (commented):
love the tests! Some comments inline. I haven't pulled and tested.

src/cljs/main/broadfcui/common/style.cljs
rushtong: That actually does not work. It was my first approach (if you look at the commits individually). That ends up causing a different problem. In the case where we remove a nil `:value` property, and some other component then tries to manipulate the value, react complains about changing a component from controlled to uncontrolled and vice versa: https://facebook.github.io/react/docs/forms.html#controlled-components

package.json
rushtong: No reason other than making the minimum amount of changes needed to address the story. I'll see if there are any issues with the higher version.

src/cljs/main/broadfcui/common/style.cljs
davidangb: Gotcha ... understood! I'm okay with the empty string ... I believe this is used only for `<input type="text"`, right?

rushtong: Correct, we're only using this approach in `create-text-field` and `create-search-field` so it's very limited. We *may* need to use this in other fields, but I did not find any cases like that.

The other strange thing about this that I don't understand, and one reason I tried to keep this change very isolated, is that it does not happen with jquery2+react.

davidangb (approved):
👍 for code, plus the previous smoke testing I've done. I did not give this a thorough test, since ... that's exactly what you did.

MattPutnam (approved):
👍

PR #773, opened 2017-05-05
Fix Sign-Out

dmohs:
Our current implementation of close-on-click prevents click events which breaks sign-out. Disabling for now.
```
Ran 3 tests containing 16 assertions.
0 failures, 0 errors.
All tests passed.
```


MattPutnam (approved):
Verified that clicking sign out does nothing before this change, and works again with it.

PR #783, opened 2017-05-09
Fix travis

dmohs:
https://broadinstitute.atlassian.net/browse/GAWB-2054

### DIFF
--- a/scripts/.tests-compose.yaml
+++ b/scripts/.tests-compose.yaml
@@ -31,7 +31,7 @@ services:
     volumes:
       - work:/w
     working_dir: /w/resources/public
-    command: python -m SimpleHTTPServer 80
+    command: bash -c 'npm install -g node-static && static -a 0.0.0.0 -p 80'
 volumes:
   work:
   jar-cache:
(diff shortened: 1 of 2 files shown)


scripts/.tests-compose.yaml
```diff
@@ -31,7 +31,7 @@ services:
     volumes:
       - work:/w
     working_dir: /w/resources/public
-    command: python -m SimpleHTTPServer 80
+    command: bash -c 'npm install -g node-static && static -a 0.0.0.0 -p 80'
```
zarsky-broad, line 34: Why does it need to be exposed? Is just serving at localhost insufficient?

davidangb:
I like that when the problem this PR fixes is "Travis doesn't work for PRs", I can just look at this PR and see that Travis works.

scripts/.tests-compose.yaml
dmohs: It is not exposed to the host (which is good) but it does need to be exposed within the container network so the PhantomJS container can pull down the page for running tests.

zarsky-broad (approved):

PR #786, opened 2017-05-11
DSDEQA-90 Render extra Button props on the anchor element

breilly2:
Also added data-test-id to create workspace button.

Specifying data-test-id for TextField components was simple because TextField already merged extra props into the attributes being rendered on the input element. However, not all (most?) components do this. @dmohs suggested that components should be migrated to work like TextField. This seemed preferable to adding an explicit property to components specifically for the test ID.

The goal is to be able to easily add test-data-id attributes. If you have other ideas about how this should be done, please feel free to make suggestions. Thanks.

  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/components.cljs
+++ b/src/cljs/main/broadfcui/common/components.cljs
@@ -50,21 +50,23 @@
    :render
    (fn [{:keys [props]}]
      (let [{:keys [color icon href disabled? onClick text style class-name]} props]
-       [:a {:className (or class-name "button")
-            :style (merge
-                    {:display "inline-flex" :alignItems "center" :justifyContent "center"
-                     :backgroundColor (if disabled? (:disabled-state style/colors) color)
-                     :cursor (when disabled? "default")
-                     :color (if disabled? (:text-light style/colors) "white")
-                     :fontWeight 500
-                     :minHeight 19 :minWidth 19
-                     :borderRadius 2 :padding (if text "0.7em 1em" "0.4em")
-                     :textDecoration "none"}
-                    (if (map? style) style {}))
-            :href (or href "javascript:;")
-            :onClick (if disabled? (create-error-message disabled?) onClick)
-            :onKeyDown (when (and onClick (not disabled?))
-                         (common/create-key-handler [:space] onClick))}
+       [:a (merge
+            {:className (or class-name "button")
+             :style (merge
+                     {:display "inline-flex" :alignItems "center" :justifyContent "center"
+                      :backgroundColor (if disabled? (:disabled-state style/colors) color)
+                      :cursor (when disabled? "default")
+                      :color (if disabled? (:text-light style/colors) "white")
+                      :fontWeight 500
+                      :minHeight 19 :minWidth 19
+                      :borderRadius 2 :padding (if text "0.7em 1em" "0.4em")
+                      :textDecoration "none"}
+                     (if (map? style) style {}))
+             :href (or href "javascript:;")
+             :onClick (if disabled? (create-error-message disabled?) onClick)
+             :onKeyDown (when (and onClick (not disabled?))
+                          (common/create-key-handler [:space] onClick))}
+            (dissoc props :color :icon :href :onClick :style :className))
         text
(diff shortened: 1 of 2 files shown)


src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -50,21 +50,23 @@
...
-                     :fontWeight 500
-                     :minHeight 19 :minWidth 19
-                     :borderRadius 2 :padding (if text "0.7em 1em" "0.4em")
-                     :textDecoration "none"}
-                    (if (map? style) style {}))
-            :href (or href "javascript:;")
-            :onClick (if disabled? (create-error-message disabled?) onClick)
-            :onKeyDown (when (and onClick (not disabled?))
-                         (common/create-key-handler [:space] onClick))}
+       [:a (merge
```
zarsky-broad: Is this all just to preserve `data-test-id`s? If so, why not just use that prop?

breilly2: I'm glad you asked!

I think when I first encountered (and worked around) this a while back, it was with a different component. Unlike `Button` and like `TextField`, that component used deeper composition before getting down to the actual (shadow) DOM element to apply the attributes to. Therefore, a new prop would have to be added to each layer going down. :(

Next, while we're using data-test-id at the moment, it may be useful to add additional data- props. The approach used here will not require further changes to the components to support additional data- props.

Finally, it seemed generally useful to me to be able to do this even beyond data- props. It could also be argued that the component should not allow this and instead pick and choose the props and decide how/where to apply them, in which case `TextField` should be changed. However, I was very happy with how easy it was to add the prop to `TextField` instances when I needed it.

When I asked @dmohs, he seemed to think that propagating the pattern used by `TextField` was good so that's what I did. I added you, @zarsky-broad, to this PR so you could present the counter opinion if you felt differently. Whatever we do is going to be done much more very soon, so let's make sure we make the right decision here.

Thanks.

dmohs: If we're going to ask developers to use this custom element instead of, say, `button`, we are obligated to support as much of the original control's API as feasible. Custom data properties, mouse hover events, etc. This is simply a convenient technique for doing that.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -50,21 +50,23 @@
...
+                     {:display "inline-flex" :alignItems "center" :justifyContent "center"
+                      :backgroundColor (if disabled? (:disabled-state style/colors) color)
+                      :cursor (when disabled? "default")
+                      :color (if disabled? (:text-light style/colors) "white")
+                      :fontWeight 500
+                      :minHeight 19 :minWidth 19
+                      :borderRadius 2 :padding (if text "0.7em 1em" "0.4em")
+                      :textDecoration "none"}
+                     (if (map? style) style {}))
+             :href (or href "javascript:;")
```
dmohs: This could be:
```cljs
:href "javascript:;"
```
and remove `:href` from the props dissoc below.

src/cljs/main/broadfcui/common/components.cljs
zarsky-broad: Ok, I'm sold.

src/cljs/main/broadfcui/common/components.cljs
breilly2: Cool. If I do that, I can also take `href` out of the `(let [{:keys [color icon href disabled? onClick text style class-name]} props]` line, right?

dmohs: Yep.

zarsky-broad:
Can this be closed?

breilly2:
@zarsky-broad, I'll figure that out and close or merge in the next day or so.

breilly2:
I still like this change but I'm going to abandon this PR at this point. I forgot to merge it when it was fresh. Now it's out-of-date and would require tweaks and re-review.

PR #793, opened 2017-05-16
When running tests, fail the build if it emits warnings

dmohs:
No Jira ticket. Needs only regression, which runs on all snapshots.


zarsky-broad (approved):
LGTM

MattPutnam (approved):
👍

PR #795, opened 2017-05-17
GAWB-1758 ajax pattern

UrsaStutsman:
Also included:
- GAWB-1885
- GAWB-2061
- GAWB-2062
- GAWB-2063

### DIFF
--- a/src/cljs/main/broadfcui/net.cljs
+++ b/src/cljs/main/broadfcui/net.cljs
@@ -6,31 +6,32 @@
   ))
 
 (defn handle-ajax-response [update-fn]
-  (fn [{:keys [xhr success?]}]
+  (fn [{:keys [xhr status-code success?]}]
     (if-let [parsed-response (as-> xhr x
-                                (aget x "responseText")
-                                (try (js-invoke js/JSON "parse" x) (catch js/Error e e))
-                                (when-not (instance? js/Error x)
-                                  (js->clj x :keywordize-keys true)))]
-      (if success?
-        (update-fn :parsed parsed-response)
-        (update-fn :error (select-keys parsed-response [:message :statusCode])))
-      (update-fn :error "Error parsing server response"))))
+                                   (aget x "responseText")
+                                   (try (js-invoke js/JSON "parse" x) (catch js/Error e e))
+                                   (when-not (instance? js/Error x)
+                                     (js->clj x :keywordize-keys true)))]
+      (update-fn (utils/restructure success? status-code parsed-response))
+      (update-fn {:success? false
+                  :status-code status-code
+                  :parsed-response {:message "Error parsing server response"}
+                  :original-request xhr}))))
 
-(defn render-with-ajax [ajax-response render-success & {:keys [loading-text rephrase-error error-handler]}]
+(defn render-with-ajax [{:keys [success? parsed-response] :as ajax-response} render-success
+                        & {:keys [loading-text rephrase-error error-handler]}]
   (assert (not (and rephrase-error error-handler)) "Provide EITHER error-handler OR rephrase-error")
-  (let [error (:error ajax-response)]
-    (cond
-      (nil? ajax-response)
-      [comps/Spinner {:text (or loading-text "Loading...")}]
-      error
-      (if error-handler
-        (error-handler error)
-        (style/create-server-error-message
-         (if rephrase-error
-           (rephrase-error error)
-           (:message error))))
-      :else (render-success))))
+  (cond
+    (nil? ajax-response)
+    [comps/Spinner {:text (or loading-text "Loading...")}]
+    (not success?)
+    (if error-handler
+      (error-handler parsed-response)
+      (style/create-server-error-message
+       (if rephrase-error
+         (rephrase-error parsed-response)
+         (:message parsed-response))))
+    :else (render-success)))
 
 (defn create-error-message-for-code
   ([message code error]
--- a/src/cljs/main/broadfcui/net.cljs
+++ b/src/cljs/main/broadfcui/net.cljs
@@ -5,21 +5,41 @@
   [broadfcui.common.style :as style]
   [broadfcui.common.components :as comps]))
 
-(defn handle-ajax-state [[success? xhr state error-keyword]]
+(defn handle-ajax-state [[success? xhr state error-keyword parsed-keyword]]
   (let [r (as-> xhr x
                 (aget x "responseText")
                 (js-invoke js/JSON "parse" x)
                 )]
     (if r
       (if success?
-        (swap! state assoc :parsed-response r)
+        (swap! state assoc (if parsed-keyword parsed-keyword :parsed-response) r)
         (swap! state assoc error-keyword (aget r "message")))
       (swap! state assoc error-keyword "Parsing Error"))
     r))
 
-(defn render-ajax-state [[state error-keyword passed-keyword loading-message function-to-render]]
+(defn render-ajax-state [[state error-keyword passed-keyword loading-message function-to-render & args]]
   (cond
     (error-keyword @state) (style/create-server-error-message (error-keyword @state))
-    (passed-keyword @state) (swap! state assoc passed-keyword (function-to-render))
+    (passed-keyword @state) (swap! state assoc passed-keyword (function-to-render args))
     :else [comps/Spinner {:text loading-message}])
   )
+
+(defn create-handle-ajax-response [state state-key]
+  (fn [{:keys [xhr success?]}]
+    (swap! state assoc state-key)
+    (let [parsed-response (as-> xhr x
+                  (aget x "responseText")
+                  (js-invoke js/JSON "parse" x))]
+     (cond
+       (complement parsed-response) (swap! state assoc-in [state-key :parse-error] "Error in parsing response");;("assgn parse error to error")
+       ((not success?)) (swap! state assoc-in [state-key :error] (aget parsed-response "message"));;("assign error to error")
+       :else (swap! state assoc-in [state-key :parsed-response] parsed-response)))
+
+    ) ;;writing to passed key map of "response" for
+
+(defn render-ajax [state state-key render-success]
+  (utils/log state-key)
+  (cond
+    (state-key state) [comps/Spinner {:text "Loading..."}]
+    (get-in state [state-key :error]) (style/create-server-error-message ((get-in state [state-key :error])))
+    :else (render-success)))) ;;reading the passed key map to properly render state
--- a/src/cljs/main/broadfcui/net.cljs
+++ b/src/cljs/main/broadfcui/net.cljs
@@ -1,52 +1,34 @@
 (ns broadfcui.net
   (:require
-  [dmohs.react :as react]
-  [broadfcui.utils :as utils]
-  [broadfcui.common.style :as style]
   [broadfcui.common.components :as comps]
+  [broadfcui.common.style :as style]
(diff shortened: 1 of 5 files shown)


zarsky-broad (approved):
👍 👍 👍 👍 👍

src/cljs/main/broadfcui/page/workspace/details.cljs
```diff
@@ -108,10 +108,13 @@
...
       {:endpoint (endpoints/get-workspace (:workspace-id props))
-       :on-done (fn [{:keys [success? get-parsed-response status-text]}]
+       :on-done (fn [{:keys [success? get-parsed-response xhr]}]
                   (if success?
                     (swap! state assoc :workspace (process-workspace (get-parsed-response)))
-                    (swap! state assoc :workspace-error status-text)))}))
+                    (swap! state assoc :workspace-error (as-> xhr x
+                                                            (aget x "responseText")
+                                                            (js-invoke js/JSON "parse" x)
+                                                            (aget x "message")))))}))
```
MattPutnam: Does `(:message (get-parsed-response))` not work?

UrsaStutsman: It does! Hmmm. Except I remember heading in that direction and @dmohs warning me off - I think maybe because get-parsed-response.. I don't remember why actually. So yeah - tagging him. If he's okay with that change - I am...

dmohs: I'm not crazy about `get-parsed-response` because it's tricky to see the function signature to understand how it works. In either case, there is one more step that needs to happen, which is handling the potential parse error if/when the error response is garbage. I'm also okay with simply ignoring that case since it is likely to be rare.

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,56 @@
...
+(defn create-handle-ajax-response [state state-key]
+  (fn [{:keys [xhr success?]}]
+    (swap! state assoc state-key nil)
+    (let [parsed-response (as-> xhr x
+                                (aget x "responseText")
+                                (try (js-invoke js/JSON "parse" x) (catch js/Error e e))
+                                (when-not (instance? js/Error x)
+                                  (js->clj x :keywordize-keys true)))]
+     (cond
+       (not parsed-response) (swap! state assoc-in [state-key :error] "Error parsing server response");;("assgn parse error to error")
```
zarsky-broad:
```cljs
(update-state :error "Error parsing server response")
```

zarsky-broad:
```cljs
(create-handle-ajax-response
  (fn [k v]
    (swap! state assoc-in [state-key k] v)))
```

src/cljs/main/broadfcui/page/groups/groups_management.cljs
```diff
@@ -13,84 +13,83 @@
...
+                   {:text (icons/icon {} :delete)
+                    :style {:float "right"}
+                    :onClick (fn []
+                               (swap! state assoc :deleting? true)
+                               (endpoints/call-ajax-orch
+                                {:endpoint (endpoints/delete-group groupName)
+                                 :on-done (fn [{:keys [success?]}]
+                                            (if success?
+                                             (this :-load-data))
+                                            )}))})))}]}
```
zarsky-broad: Oop, forgot to pull this up...

src/cljs/main/broadfcui/page/method_repo/methods_configs_acl.cljs
```diff
@@ -22,77 +23,79 @@
...
-                          public-status (or (:role public-user) no-access-level)]
-                      (swap! state assoc :acl-vec acl-vec
-                             :public-status (= public-status reader-level)
-                             :count-orig (count acl-vec)))
-                    (swap! state assoc :error status-text)))}))
-   :persist-acl
+       :on-done (net/handle-ajax-response
+                 (fn [k v]
+                   (swap! state assoc-in [:acl-response k] v)
+                   (let [acl-vec (filterv #(not= "public" (:user %)) v)
```
dmohs: Instead of saving each of these into state as well, consider moving this let into the render method.

zarsky-broad: Poked further, this stuff needs to be in state.

zarsky-broad:
@dmohs @MattPutnam @UrsaStutsman take a look, plz

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+                                     (js->clj x :keywordize-keys true)))]
+      (update-fn (utils/restructure success? status-code status-text parsed-response))
+      (update-fn (merge
+                  (utils/restructure success? status-code status-text)
+                  {:parsed-response {:message "Error parsing server response"}
+                   :original-request xhr})))))
+
+(defn render-with-ajax
+  ([ajax-response render-success] (render-with-ajax ajax-response render-success {}))
+  ([{:keys [success? parsed-response] :as ajax-response} render-success
```
MattPutnam: Bump `render-success` down a line, it's kinda lost on the end there.

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+        (if rephrase-error
+          (rephrase-error parsed-response)
+          (:message parsed-response))))
+     :else (render-success))))
+
+(defn create-error-message-for-code
+  ([message code error]
+   (if (= code (:statusCode error))
+     message
+     (:message error))))
```
MattPutnam: I feel like this should take a map of code->message that it can match against.  See `components/ErrorViewer`'s `:expect` property.

UrsaStutsman: Match against the code? (or the code and message?)

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+    (if-let [parsed-response (as-> xhr x
+                                   (aget x "responseText")
+                                   (try (js-invoke js/JSON "parse" x) (catch js/Error e e))
+                                   (when-not (instance? js/Error x)
+                                     (js->clj x :keywordize-keys true)))]
+      (update-fn (utils/restructure success? status-code status-text parsed-response))
+      (update-fn (merge
+                  (utils/restructure success? status-code status-text)
+                  {:parsed-response {:message "Error parsing server response"}
+                   :original-request xhr})))))
```
dmohs: If you want to rename this, I think you also need to rename it in `ajax`. Otherwise it's confusing. If you do, can you think of something more descriptive than "original?"

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+
+(defn render-with-ajax
+  ([ajax-response render-success] (render-with-ajax ajax-response render-success {}))
```
dmohs: Nit: I think passing `nil` would be more idiomatic.

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+  ([{:keys [success? parsed-response] :as ajax-response} render-success
+    {:keys [loading-text rephrase-error error-handler] :as extras}]
+   (assert (not (and rephrase-error error-handler)) "Provide EITHER error-handler OR rephrase-error")
```
dmohs: It's a bit difficult to understand how to use the error handlers. I suggest renaming `error-handler` to `handle-error` (start it with a verb) and passing `ajax-response`. Then, you don't really need `create-error-message-for-code`. See below for notes on usage.

src/cljs/main/broadfcui/page/method_repo/methods_configs_acl.cljs
```diff
@@ -22,77 +23,80 @@
...
-       :ok-button (when (:acl-vec @state) {:text "Save" :onClick #(react/call :persist-acl this)})}])
+        (net/render-with-ajax
+         (:acl-response @state)
+         #(this :-render-acl-form)
+         {:loading-text (str "Loading Permissions for " (:title props) "...")
+          :rephrase-error
+          #(net/create-error-message-for-code
+            (str "You are unauthorized to edit this "
+                 (clojure.string/lower-case (:entityType props)) ".")
+            403 %)}))
```
dmohs: This would become
```cljs
#(if (= 403 (:status-code %))
    (str "You are not...")
    (get-in % [:parsed-response :message]))
```
One less level of indirection, plus the status code is read from the XHR object instead of the parsed response, which is code that we wrote so may be less reliable.

src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -398,31 +399,28 @@
...
+        (:workspaces-response server-response)
+        (fn []
+          [:div {:style {:padding "0 1rem"}}
+           [WorkspaceTable
+            (assoc props
+              :workspaces workspaces
+              :billing-projects billing-projects
+              :disabled-reason disabled-reason)]])
+        {:loading-text "Loading workspaces..."
+         :rephrase-error #(get-in % [:workspaces :error-message])})))
```
dmohs: Not much harder:
```cljs
#(get-in % [:parsed-response :workspaces :error-message])
```

src/cljs/main/broadfcui/net.cljs
dmohs: The other advantage of passing `ajax-response` is that users of these functions get some predictability—a similarly-structure map is available whenever they need to read values.

dmohs (commented):
👍  Great pattern. Solves the problem of ensuring that the main cases are always handled. Couple of suggestions around error handling.

zarsky-broad:
@dmohs @MattPutnam mind glancing again post-updates?

src/cljs/main/broadfcui/net.cljs
```diff
@@ -0,0 +1,43 @@
...
+                  (utils/restructure success? status-code status-text xhr)
+                  {:parsed-response {:message "Error parsing server response"}})))))
+
+(defn render-with-ajax
+  ([ajax-response render-success] (render-with-ajax ajax-response render-success nil))
+  ([{:keys [success? parsed-response] :as ajax-response}
+    render-success
+    {:keys [loading-text rephrase-error handle-error] :as extras}]
+   (assert (not (and rephrase-error handle-error)) "Provide EITHER handle-error OR rephrase-error")
+   extras
```
MattPutnam: dead code

src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -435,7 +433,7 @@
 
 (react/defc Page
   {:render
-   (fn [{:keys [props]}]
+   (fn []
      [:div {:style {:marginTop "1.5rem"}}
       [WorkspaceList]])})
```
MattPutnam, line 438: This is now a completely dead React component.  Can we just make this a `def`?

MattPutnam (approved):
👍 Looks good, just a couple of minor things.

src/cljs/main/broadfcui/page/workspaces_list.cljs
zarsky-broad: Not easily. All Page elements are called as react components when they're rendered. We'd have to build an exception.

dmohs (approved):

PR #805, opened 2017-05-24
URL-handling design doc

dmohs:
https://broadinstitute.atlassian.net/browse/GAWB-2081


MatthewBemis (approved):
Sorry for the slow approval! This looks like a totally sane solution to me. My expectations were that this would require some gross hacks but you've proven that to be wrong. :)

dmohs:
@dvoet 
> there is still duplicated code because there is not shared library that handles this file

There would have to be a Scala implementation (which could be shared) and a ClojureScript implementation (same). I don't see this as duplicated code.

> backwards compatibility

Good point. In the current version, we also define a set of redirects, which this should probably handle as well.

> I also don't really like thurloe talking to ui

Since Thurloe is constructing UI URLs, it is implicitly depending on the UI. This really just makes the communication channel explicit.

MatthewBemis:
:+1: seems easier

PR #807, opened 2017-05-26
Move build target directory so cleans don't remove all static files

dmohs:
https://broadinstitute.atlassian.net/browse/GAWB-2157


dmohs:
I have received approval from my team, so merging this.

PR #812, opened 2017-05-30
[GAWB-1623] Method Config Edit: Cancel Editing does not restore origi…

rtitle:
…nal Snapshot ID

See: https://broadinstitute.atlassian.net/browse/GAWB-1623

My solution was to remember the current `snapshot-id` when entering Edit mode, and restore the stored snapshot ID on edit Cancel events.

It works locally - let me know if there is a better way to do this.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
+++ b/src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
@@ -95,7 +95,7 @@
                      (swap! state assoc :error status-text)))}))})
 
 
-(defn- render-side-bar [this state refs config editing? props]
+(defn- render-side-bar [{:keys [state refs config editing? props restore-on-cancel]}]
   [:div {:style {:width 290 :float "left"}}
    [:div {:ref "sidebar"}]
    (style/create-unselectable :div {:style {:position (when-not (:sidebar-visible? @state) "fixed")
@@ -129,13 +129,15 @@
         (when editing?
           [comps/SidebarButton {:color :success-state
                                 :text "Save" :icon :done
-                                :onClick #(do (commit state refs config props)
-                                              (stop-editing state))}])
+                                :onClick (fn []
+                                           (commit state refs config props)
+                                           (stop-editing state))}])
         (when editing?
           [comps/SidebarButton {:color :exception-state :margin :top
                                 :text "Cancel Editing" :icon :cancel
-                                :onClick #(do (react/call :load-new-method-template this (:prev-snapshot-id @state))
-                                              (stop-editing state))}])]))])
+                                :onClick (fn []
+                                           (restore-on-cancel (:prev-snapshot-id @state))
+                                           (stop-editing state))}])]))])
 
 
 (defn- input-output-list [values ref-prefix invalid-values editing? all-values]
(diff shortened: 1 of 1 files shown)


rtitle:
@zarsky-broad mind taking a look? Let me know if anything looks wonky.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -133,7 +134,8 @@
         (when editing?
           [comps/SidebarButton {:color :exception-state :margin :top
                                 :text "Cancel Editing" :icon :cancel
-                                :onClick #(stop-editing state)}])]))])
+                                :onClick #(do (react/call :load-new-method-template this (:prev-snapshot-id @state))
```
zarsky-broad: 1. If you stick this all in a `(fn [])`, you won't need the `do`.
2. Neater way to do this:
```cljs
(this :load-new-method-template (:prev-snapshot-id @state))
```

zarsky-broad: 3. It's entirely possible I'm misreading this (not running it right now), but couldn't you just use the `snapshot-id` you `let` previously and skip this `:prev-snapshot-id` thing entirely?

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -216,7 +218,7 @@
     [:div {}
      [comps/Blocker {:banner (:blocker @state)}]
      [:div {:style {:padding "1em 2em"}}
-      (render-side-bar state refs config editing? props)
+      (render-side-bar this state refs config editing? props)
```
zarsky-broad: If the only thing you're using from `this` is the one method, why not just pass that method directly here?

Also, not your fault, but this is an awfully large number of positional arguments. You might pass them with `utils/restructure` here and pull them out with `:keys` on the other end, but it's not a huge deal, as these are all private methods...

zarsky-broad (commented):
Totally reasonable. A few comments.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
rtitle: Thanks, good tips. I'll probably just pass the method directly instead of `this`, as you suggested below.

Re (3), I initially tried that and it didn't work: when I click Cancel it calls `:load-new-method-template` with the _current_ snapshot-id, not the original snapshot-id before edit was clicked. I could be missing something, but I think the `render-side-bar` function gets called while in edit mode, which overwrites the `onClick` callback on the Cancel button. Saving the previous snapshot-id in `@state` works around that.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
rtitle: 👍 to passing the method directly.

I'll give `utils/restructure` a shot.

rtitle:
@zarsky-broad made edits based on feedback ^

zarsky-broad (approved):

PR #816, opened 2017-06-02
[GAWB-2056] The status of the last run is showing green when there ar…

rtitle:
…e errors in them

https://broadinstitute.atlassian.net/browse/GAWB-2056

Not sure why string keys were being used here - the map lookups were returning `nil`. Using keywords fixes the issue. Also put keywords before the map, as that seems to be the convention.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/workspace/monitor/common.cljs
+++ b/src/cljs/main/broadfcui/page/workspace/monitor/common.cljs
@@ -29,15 +29,15 @@
 (def google-cloud-context "https://console.cloud.google.com/storage/browser/")
 
 (defn all-success? [submission]
-  (and (every? #(contains? wf-success-statuses (% "status")) (submission "workflows"))
-       (zero? (count (submission "notstarted")))))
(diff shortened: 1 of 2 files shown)


rtitle:
@dmohs want to check this out, make sure it makes sense?

dmohs (approved):
A while back we converted some of our AJAX calling code to convert from strings to keywords, so that could explain these string keys. Thanks for fixing!

rtitle:
Np, thanks for reviewing.

For next beer hour: I submit that this bug would not be possible in a strongly-typed language, since getting a map with the wrong key type would be a compile error :)

dmohs:
I submit that this isn't a very interesting discussion :) Of course that is true. The interesting bit is around, "is it worth it?" Now that I'm happy to dig into!

rtitle:
Fair enough :)

PR #849, opened 2017-07-08
converting existing regression tests into scala

andy7i:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/MethodConfigTabSpec.scala
+++ b/automation/src/test/scala/MethodConfigTabSpec.scala
@@ -4,9 +4,10 @@ import org.broadinstitute.dsde.firecloud.auth.{AuthToken, AuthTokens, Credential
 import org.broadinstitute.dsde.firecloud.data.TestData
 import org.broadinstitute.dsde.firecloud.pages.{WebBrowserSpec, WorkspaceListPage, WorkspaceMethodConfigPage, SubmissionDetailsPage}
 import org.broadinstitute.dsde.firecloud.{CleanUp, Config}
+import org.broadinstitute.dsde.firecloud.workspaces.WorkspaceFixtures
 import org.scalatest._
 
-class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp {
+class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp with WorkspaceFixtures[MethodConfigTabSpec] {
 
   val billingProject: String = Config.Projects.default
   val methodName: String = TestData.SimpleMethod.name + "_" + UUID.randomUUID().toString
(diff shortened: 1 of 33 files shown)


automation/src/test/scala/MethodConfigTabSpec.scala
```diff
@@ -142,18 +142,112 @@ class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp {
...
 
+  // This testcase requires pulling in new code from develop branch
   "import a method config from a workspace" in withWebDriver { implicit driver =>
 
   }
 
   "import a method config into a workspace from the method repo" in withWebDriver { implicit driver =>
+    val wsName = "TestSpec_FireCloud_import_method_config_from_workspace" + UUID.randomUUID.toString
+    api.workspaces.create(billingProject, wsName)
+    register cleanUp api.workspaces.delete(billingProject, wsName)
```
breilly2: These 3 lines can be replaced with:
```
  withWorkspace(billingProject, Option("TestSpec_FireCloud_import_method_config_from_workspace") { wsName =>
    ...
  }
```
I'm also strongly considering removing the need to wrap the workspace name prefix in `Option(...)`. Originally, it felt like the correct thing to do, but I don't think it takes anything away from the `withWorkspace` API to default to `""` instead of `None`.

automation/src/test/scala/MethodConfigTabSpec.scala
```diff
@@ -142,18 +142,112 @@ class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp {
...
+  "abort a workflow" in withWebDriver { implicit driver =>
+    val wsName = "TestSpec_FireCloud_abort_workflow" + UUID.randomUUID.toString
+    val shouldUseCallCaching = false
+    api.workspaces.create(billingProject, wsName)
+    register cleanUp api.workspaces.delete(billingProject, wsName)
+    api.importMetaData(billingProject, wsName, "entities", TestData.SingleParticipant.participantEntity)
+
+    signIn(uiUser)
+    val workspaceMethodConfigPage = new WorkspaceMethodConfigPage(billingProject, wsName).open
+    val methodConfigDetailsPage = workspaceMethodConfigPage.importMethod(TestData.SimpleMethod.namespace,
```
breilly2: I think it would be fine for this test case to use the web service API to import the method because importing a method through the UI is exercised in other test cases above and it is not really related to aborting a workflow.

automation/src/test/scala/MethodConfigTabSpec.scala
```diff
@@ -142,18 +142,112 @@ class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp {
...
+      TestData.SimpleMethod.name, TestData.SimpleMethod.snapshotId, methodConfigName, TestData.SimpleMethod.rootEntityType)
+    methodConfigDetailsPage.editMethodConfig(inputs = Some(TestData.SimpleMethod.inputs))
+    val submissionDetailsPage = methodConfigDetailsPage.launchAnalysis(TestData.SimpleMethod.rootEntityType, TestData.SingleParticipant.entityId, "", shouldUseCallCaching)
```
breilly2: The analysis could be launched using the web service API. I'm a little more on the fence about this one. It is closely related to aborting a submission. However, it would be unfortunate to not test aborting a submission in the presence of a bug with launching an analysis (which would be caught by a different test case).

automation/src/test/scala/MethodConfigTabSpec.scala
```diff
@@ -142,18 +142,112 @@ class MethodConfigTabSpec extends FreeSpec with WebBrowserSpec with CleanUp {
...
+
+    submissionDetailsPage.abortSubmission()
+    submissionDetailsPage.waitUntilSubmissionCompletes()
+    assert(submissionDetailsPage.verifyWorkflowAborted())
 
   }
 
   // negative tests
```
breilly2: Not related to your changes, but I don't think that "delete a method from a workspace" is a negative test.

automation/src/test/scala/WorkspaceSpec.scala
```diff
@@ -60,6 +60,22 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures[
         listPage.ui.hasWorkspace(billingProject, workspaceName) shouldBe false
       }
     }
+
+    "should be able to log out and log in as another user" in withWebDriver { implicit driver =>
```
breilly2: These test cases would be better in a different test suite, maybe something like `SignInSpec`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/BillingSpec.scala
```diff
@@ -19,18 +22,99 @@ class BillingSpec extends FreeSpec with WebBrowserSpec with CleanUp
         signIn(Config.Users.owner)
 
         val billingPage = new BillingManagementPage().open
-        val projectName = "billing-spec-create-" + makeRandomId()
-        logger.info(s"Creating billing project: $projectName")
+        val billingProjectName = "billing-spec-create-" + makeRandomId() // is this a unique ID?
```
breilly2: No, it's not guaranteed to be unique. Unfortunately, the max length of a Google project name is too short to allow us to use a UUID. `makeRandomId()` defaults to 7 characters/digits which will hopefully be unique enough for the foreseeable future. If it becomes a problem, I think we have room to increase the length to 8 or 9.

breilly2 (commented):
Just an initial set of comments.

automation/src/test/scala/SignInSpec.scala
```diff
@@ -0,0 +1,31 @@
...
+  implicit val authToken: AuthToken = AuthTokens.harry
+
+  "A user" - {
+    "with a registered account" - {
+
+      "should be able to log out and log in as another user" in withWebDriver { implicit driver =>
+        val listPageAsUser1 = signIn(Config.Users.harry)
+        listPageAsUser1.signOut()
+        val listPageAsUser2 = signIn(Config.Users.ron)
+        assert(listPageAsUser2.getUser().startsWith(Config.Users.ron.email))
```
breilly2, line 17: If you include this assertion in the test below, that test will be more comprehensive. Also, this test would then be a subset of the test below and could be removed.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -142,6 +142,16 @@ trait Orchestration extends FireCloudClient with LazyLogging {
     postRequestWithMultipart(Config.FireCloud.orchApiUrl + "api/workspaces/" + ns + "/" + wsName + "/" + "importEntities", fileName, fileContent)
   }
 
+  def importMethodConfigFromRepo(ns: String, wsName: String, configurationNamespace: String, configurationName: String, configurationSnapshotId: Int, destinationNamespace: String, destinationName: String)(implicit token: AuthToken) = {
```
breilly2, line 148: This should go in a new `methodConfigs` object to provide a structure similar to what we expose through swagger. I know we still have a couple of outliers that need to be relocated... sorry about that.

Also, I think it's helpful for these API methods to log what they're doing to help with debugging. Again, I'm sorry that we haven't yet done this universally. I've added both of these clean-up tasks to my TODO list.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -142,6 +142,16 @@ trait Orchestration extends FireCloudClient with LazyLogging {
...
+    postRequest(Config.FireCloud.orchApiUrl + "api/workspaces/" + ns + "/" + wsName + "/" + "method_configs/copyFromMethodRepo",
+      Map("configurationNamespace" -> configurationNamespace, "configurationName" -> configurationName, "configurationSnapshotId" -> configurationSnapshotId, "destinationNamespace" -> destinationNamespace, "destinationName" -> destinationName))
+  }
+
+  def launchWorkflow(ns: String, wsName: String, methodConfigurationNamespace: String, methodConfigurationName: String, entityType: String, entityName: String, expression: String, useCallCache: Boolean, workflowFailureMode: String = "NoNewCalls")(implicit token: AuthToken) = {
```
breilly2, line 153: Similarly, this should go in a `submissions` object.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Also, if it's in a `methodConfigs` object context, the function name can be shortened to `importFromMethodRepo` (or `copyFromMethodRepo` to make it more closely match the API vocabulary).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/pages/AuthenticatedPage.scala
```diff
@@ -15,10 +15,20 @@ abstract class AuthenticatedPage(implicit webDriver: WebDriver) extends FireClou
     ui.clickSignOut()
   }
 
+  def filter(searchText: String): Unit = {
```
breilly2, line 18: I don't think this is general enough that it should be in `AuthenticatedPage`. It looks like you added it for a test of managing methods on workspaces so I think this should move into `WorkspaceMethodConfigPage`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/pages/AuthenticatedPage.scala
```diff
@@ -15,10 +15,20 @@ abstract class AuthenticatedPage(implicit webDriver: WebDriver) extends FireClou
...
+    ui.filter(searchText)
+  }
+
+  def getUser(): String = {
```
breilly2, line 22: Nit: IntelliJ likes to infer things about functions starting with `get`. I've been favoring `read` instead, i.e. `readUserEmail`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/pages/AuthenticatedPage.scala
```diff
@@ -31,6 +41,21 @@ abstract class AuthenticatedPage(implicit webDriver: WebDriver) extends FireClou
...
+    def filter(searchText: String) = {
+      await enabled filterInput
+      searchField(filterInput).value = searchText
+      pressKeys("\n")
+    }
+
+    def getUser() = {
+      await enabled accountDropdown
+      val userDropdownText = find(accountDropdown).get.text
+      emailRegEx findFirstIn userDropdownText match {
```
breilly2, line 54: It might be worthwhile to put a `span` with a `data-test-id` around the email address so we can read it out without the need for a regex.

automation/src/test/scala/SignInSpec.scala
andy7i: I split them up to respect the principle of "one assert / test one _thing_". Do you think it is okay to combine them?

breilly2: Yes. It's all one behavior (logging out and back in as a different user). You'd still be testing one thing even with more than one assert. (I feel the "one assert" rule is more of a thought provoking exercise for people learning to how to test only one thing, not a rule to strictly adhere to.)

I'd even go as far as to say that the test that logs in as Harry then Ron then Harry should absolutely have the extra assert. Without it, the test would pass even if Harry was somehow never signed out.

PR #852, opened 2017-07-10
Swatomation

ahaessly:
merging testcode into ui code
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


rtitle:
Is this ready for review? I'm interested and would like to go through it (focusing mostly on the Scala stuff), but only if its authors say it's ready.

breilly2:
@rtitle Thank you for offering to review. I'm hesitant to have you look at it just yet because there is some remaining code clean-up on my list that will improve consistency and readability. I expect to get through enough of my list by start-of-business tomorrow (Wednesday) to make it worth reviewing. I'll keep you updated.

src/cljs/main/broadfcui/header.cljs
```diff
@@ -42,7 +42,7 @@
                             :color "#000" :textDecoration "none"
                             :padding "0.5rem" :border style/standard-line}
                     :data-test-id "account-dropdown"}
-              (-> auth2 (.-currentUser) (.get) (.getBasicProfile) (.getEmail))
+              [:div {:style {:display "inline-block"} :data-test-id "account-dropdown-email"} (-> auth2 (.-currentUser) (.get) (.getBasicProfile) (.getEmail))]
```
breilly2: Any reason not to use a `span` instead of a `div`? The `span` should be inherently inline so you wouldn't need the `inline-block` style.

README.md
```diff
@@ -68,3 +68,48 @@ To compile and build the `broadinstitute/firecloud-ui` docker image, run
...
+
+#### Running tests
+
+To run all tests:
+```
+sbt test -Djsse.enableSNIExtension=false -Dheadless=false
+sbt clean
+```
+
+TODO - running specific tests/suites
```
rtitle: You can run single suites like:
```
sbt "testOnly *GoogleSpec"
```
Or single tests within a suite like:
```
# matches test via substring
sbt "testOnly *GoogleSpec -- -z \"have a search field\""
```
See: http://www.scala-sbt.org/0.13/docs/Testing.html#Test+Framework+Arguments

automation/src/test/scala/GoogleSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+  * Example test for running with Selenium Grid until we work out docker
+  * networking to be able to access FiaBitC.
+  */
+class GoogleSpec extends FlatSpec with WebBrowserSpec with ParallelTestExecution with WebBrowser {
+
+  behavior of "Google"
+
+  it should "have a search field" in withWebDriver { implicit driver =>
+    go to new GooglePage()
+    assert(find("q").isDefined)
```
rtitle, line 15: I think you mentioned this in email, but I generally prefer ScalaTest matchers over `assert(boolean)` because it gives nicer error messages (and usually reads better). If you mix in the `ShouldMatchers` trait this could be written as:
```
find("q") shouldBe 'defined
``` 

See: http://www.scalatest.org/user_guide/using_matchers

Not really a big deal and I won't comment about other occurences of this.

automation/src/test/scala/GoogleSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+//    Thread sleep 10000
+  }
+
+  it should "have a 'Google Search' button" in withWebDriver { implicit driver =>
+    go to new GooglePage()
+    val button = find(xpath("//input[@value='Google Search'][@type='submit']"))
+    assert(button.isDefined)
+    //    Thread sleep 10000
```
rtitle: Remove commented code

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/APIException.scala
```diff
@@ -0,0 +1,4 @@
+package org.broadinstitute.dsde.firecloud.api
+
+
+class APIException (message: String = null, cause: Throwable = null) extends Exception(message, cause)
```
rtitle: This could be a `case class`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -0,0 +1,113 @@
...
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import org.broadinstitute.dsde.firecloud.config.AuthToken
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
+
+class FireCloudClient {
+  implicit val system = ActorSystem()
+  implicit val materializer = ActorMaterializer()
+  implicit val ec: ExecutionContextExecutor = ExecutionContext.global
```
rtitle: I've been telling people to avoid using the global execution context and use the `ActorSystem.dispatcher` instead. It probably doesn't matter much for tests. But I would still probably prefer:
```
implicit val ec: ExecutionContext = system.dispatcher
```

automation/src/test/scala/GoogleSpec.scala
breilly2: Looks like `ShouldMatchers` is deprecated in favor of `Matchers` which we are using in some of the other tests.

Thanks for the tip on using `defined` with matchers! I wasn't sure how to do that and I'm pretty sure there are some places where I can put that to good use.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -0,0 +1,113 @@
...
+import akka.stream.scaladsl.{Sink, _}
+import akka.util.ByteString
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.scala.DefaultScalaModule
+import org.broadinstitute.dsde.firecloud.config.AuthToken
+
+import scala.concurrent.duration._
+import scala.concurrent.{Await, ExecutionContext, ExecutionContextExecutor, Future}
+
+class FireCloudClient {
```
rtitle: Looking below, `FireCloudClient` is extended by other traits. So it might be better to make `FireCloudClient` a trait itself. (I didn't even realize that a trait could extend a class: https://stackoverflow.com/questions/12854941/why-can-a-scala-trait-extend-a-class).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -0,0 +1,113 @@
...
+  implicit val system = ActorSystem()
+  implicit val materializer = ActorMaterializer()
+  implicit val ec: ExecutionContextExecutor = ExecutionContext.global
+
+  val mapper = new ObjectMapper()
+  mapper.registerModule(DefaultScalaModule)
+
+  // TODO: Is this okay here or should it be in a separate object and explicitly imported?
```
rtitle: I'm fine with it here: it will automatically be available to subclasses. Could mark it `protected` so _only_ subclasses can access it (in general I like narrowing down implicit scope as much as possible).

automation/src/test/scala/GoogleSpec.scala
rtitle: Ah right, it's just `Matchers` now.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -0,0 +1,113 @@
...
+        val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
+        Await.result(entityFuture, 100.millis).decodeString("UTF-8")
+      case _ =>
+        val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
+        val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
+        throw new APIException(Await.result(entityFuture, 100.millis).decodeString("UTF-8"))
+    }
+  }
+
+  def parseResponse(response: HttpResponse): String = {
```
rtitle: What's the difference between `handleResponse` and `parseResponse`? They look identical to me.

Also I feel like this could be achieved more simply by acting on futures instead of chaining multiple `Await.result` calls. My attempt (annotated):
```
 private def sendRequestAndWaitForResponse(httpRequest: HttpRequest): String = {
    val future: Future[String] = for {
      response <- Http().singleRequest(httpRequest)  // makes the request and returns a Future[HttpResponse]
      responseAsString <- response.entity.toStrict(5 seconds).map(_.data.decodeString("UTF-8"))  // Loads the response into memory and decodes it as a String, returning a Future[String]
    } yield {
      // Throw exception if the response is a failure, else return it
      if (response.status.isFailure()) {
        throw new APIException(s"Request failed with status ${response.status.intValue}: $responseAsString")
      } else {
        responseAsString
      }
    }
    // Either returns a string or throws an exception
    Await.result(future, 5 minutes)
  }
```

breilly2: That is specifically on my list of things to clean up. If I remember correctly, the differences are subtle and have to do with needing to add support for something (I don't remember what) and feeling like it would require minor changes throughout the tests that I wasn't prepared to do just yet. Thanks for the implementation suggestion!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  def postRequestWithMultipart(uri:String, name: String, content: String)(implicit token: AuthToken): String = {
+    // val cont = ContentType.parse("text/tab-separated-values")
+    //val ranges = coalesceRanges(iRanges).sortBy(_.start)
+    //val formD = Multipart.ByteRanges(HttpEntity(ByteString("...")))
+    val formData = Multipart.FormData {
+      Source {
+        List (
+          Multipart.FormData.BodyPart.apply(name, HttpEntity(ByteString(content)))
```
rtitle: Instead of `Source(List(Multipart.FormData.BodyPart(...)))` just `Source.single(Multipart.FormData.BodyPart(...))`. 

Can also omit the `.apply`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+
+import com.typesafe.scalalogging.LazyLogging
+import org.broadinstitute.dsde.firecloud.config.{AuthToken, Config}
+
+import scala.concurrent.duration._
+
+
+trait Orchestration extends FireCloudClient with LazyLogging {
+
+  def responseAsList(response: String): List[Map[String, Object]] = {
```
rtitle, line 13: I wonder if this type of thing could be baked into the `parseResponse` in the superclass.  So you could do something like
```
val response: List[String] = parseResponse[List[Map[String, Object]]](getRequest(Config.FireCloud.orchApiUrl + "api/profile/billing"))
```

rtitle: Something like:
```
  def parseResponse2[A: ClassTag](response: HttpResponse): A = {
    val f = response.entity.toStrict(5 seconds).map(entity => mapper.readValue(entity.data.toArray, classTag[A].runtimeClass.asInstanceOf[Class[A]]))
    Await.result(f, 100 millis)
  }
```
The `.asInstanceOf` is unfortunate but that's what you get with Jackson (I'm trying to avoid saying use a more idiomatic scala JSON parser :)).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+    mapper.readValue(response, classOf[List[Map[String, Object]]])
+  }
+
+  private def retry[T](remainingBackOffIntervals: Seq[FiniteDuration])(op: => () => Boolean): Boolean = {
```
rtitle: I'm not sure if this is duplicated elsewhere but this looks like some code that could be shared instead of a `private def`.

rtitle: Also `op: => () => Boolean` seems wrong. There are too many layers of indirection. This is a "by-name parameter which returns a function which takes no arguments and returns a Boolean". But all you need is a "by-name parameter which returns a Boolean". So:
```
private def retry[T](remainingBackOffIntervals: Seq[FiniteDuration])(op: => Boolean): Boolean = {
  op match {
    // snip
  }
}
```
This should give you the behavior you need. More info: https://tpolecat.github.io/2014/06/26/call-by-name.html

rtitle: Also the method is parameterized so it should be `op: => T`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+    op() match {
+      case true => true
+      case false if remainingBackOffIntervals.isEmpty => false
+      case false if remainingBackOffIntervals.nonEmpty =>
```
rtitle: Log messages might be nice here. E.g.:
```
log.error(s"Operation failed after $n retries")
log.warning(s"Operation failed, retrying attempt $n/$m")
etc
```

Rawls has a similar retry function: https://github.com/broadinstitute/rawls/blob/develop/util/src/main/scala/org/broadinstitute/dsde/rawls/util/Retry.scala#L47

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+        Thread sleep remainingBackOffIntervals.head.toMillis
+        retry(remainingBackOffIntervals.tail)(op)
+    }
+  }
+
+  object billing {
+
+    object BillingProjectRole extends Enumeration {
```
rtitle, line 23: Just noting the Scala example for `Enumeration`: http://www.scala-lang.org/api/2.12.0/scala/Enumeration.html

It includes some type aliasing which essentially lets you refer to enum values without typing `.Value` all the time. But this should work fine as well.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    def addUserToBillingProject(projectName: String, email: String, role: BillingProjectRole.Value)(implicit token: AuthToken): Unit = {
+      putRequest(Config.FireCloud.orchApiUrl + s"api/billing/$projectName/${role.toString}/$email")
+    }
+
+    def createBillingProject(projectName: String, billingAccount: String)(implicit token: AuthToken): Unit = {
+      postRequest(Config.FireCloud.orchApiUrl + "api/billing", Map("projectName" -> projectName, "billingAccount" -> billingAccount))
+
+      retry(Seq.fill(30)(10.seconds)) { () =>
```
rtitle: Also if you take my retry advice above you would write this as 
```
retry(Seq.fill(30)(10.seconds)) { 
  // code to retry
}
```

rtitle: Also if retrying N times with a fixed delay is a common pattern it might be good to define retry as taking a number of attempts (`Int`) and a delay (`FiniteDuration`) which might be simpler to understand.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+        val response: String = parseResponse(getRequest(Config.FireCloud.orchApiUrl + "api/profile/billing"))
+        val projects: List[Map[String, Object]] = responseAsList(response)
+        projects.exists((e) =>
+          e.exists(_ == ("creationStatus", "Ready")) && e.exists(_ == ("projectName", projectName)))
+      } match {
+        case false => throw new Exception("Billing project creation did not complete")
+        case true => println("Finished creating billing project")
```
rtitle: Would be nice to use a logger such as logback-over-slf4j instead of `println`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+
+  object groups {
+
+    object GroupRole extends Enumeration {
+      val Member = Value("member")
+      val Admin = Value("admin")
+    }
+
+    def addUserToGroup(groupName: String, email: String, role: GroupRole.Value)(implicit token: AuthToken): Unit = {
+      logger.info(s"Adding user to group: $groupName $email")
```
rtitle: Oh here's a logger :)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+    }
+  }
+
+  object workspaces {
+
+    def create(namespace: String, name: String, authDomain: String = "")
+              (implicit token: AuthToken): Unit = {
+      logger.info(s"Creating workspace: $namespace/$name with auth domain: $authDomain")
+
+      val authDomainMap = authDomain match {
```
rtitle: Maybe `authDomain` should be a `Option[String]`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+
+  /*
+   *  Library requests
+   */
+
+  object library {
+    def setLibraryAttributes(ns: String, name: String, attributes: Map[String, Any])(implicit token: AuthToken): String = {
+      putRequest(Config.FireCloud.orchApiUrl + "api/library/" + ns + "/" + name + "/metadata", attributes)
```
rtitle: Seems like a good candidate for string interpolation

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Rawls.scala
```diff
@@ -0,0 +1,35 @@
...
+    def deleteUser(subjectId: String)(implicit token: AuthToken): Unit = {
+      logger.info(s"Deleting user: $subjectId")
+      deleteRequest(url + s"api/admin/user/$subjectId")
+    }
+
+    def doesUserExist(subjectId: String)(implicit token: AuthToken): Option[Boolean] = {
+      getRequest(url + s"api/admin/user/$subjectId").status match {
+        case StatusCodes.OK => Option(true)
+        case StatusCodes.NotFound => Option(false)
+        case _ => None
```
rtitle, line 31: Ah the old "troolean".

The only usage I see of this is doing a `.getOrElse(false)` -- maybe just make the method return a `Boolean`?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/AuthTokens.scala
```diff
@@ -0,0 +1,58 @@
+package org.broadinstitute.dsde.firecloud.config
+
+import com.google.api.client.googleapis.auth.oauth2.GoogleCredential
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport
+import com.google.api.client.json.jackson2.JacksonFactory
+
+import scala.collection.JavaConversions._
```
rtitle: I think the best practice is to import `scala.collection.JavaConverters._` instead of `JavaConversions._`. The difference is that `JavaConverters` enriches Java/Scala collection classes with `.asJava` and `.asScala` methods, while `JavaConversions` tries to do it for you behind the scenes. It tends to be better to be explicit about when you want a conversion to avoid surprises/unexpected errors.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/AuthTokens.scala
```diff
@@ -0,0 +1,58 @@
...
+    val email = Config.GCS.qaEmail
+
+    new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(email)
+      .setServiceAccountPrivateKeyFromPemFile(pemfile)
+      .setServiceAccountScopes(authScopes)
+      .setServiceAccountUser(userEmail)
+      .build();
```
rtitle: semicolon

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/AuthTokens.scala
```diff
@@ -0,0 +1,58 @@
...
+  def buildCredential(userEmail: String): GoogleCredential = {
+    val pemfile = new java.io.File(Config.GCS.pathToQAPem)
+    val email = Config.GCS.qaEmail
+
+    new GoogleCredential.Builder()
+      .setTransport(httpTransport)
+      .setJsonFactory(jsonFactory)
+      .setServiceAccountId(email)
+      .setServiceAccountPrivateKeyFromPemFile(pemfile)
+      .setServiceAccountScopes(authScopes)
```
rtitle: Probably would do `authScopes.asJava` if using `JavaConverters`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/TestData.scala
```diff
@@ -0,0 +1,62 @@
...
+      "sampleSet101"
+    private val samplePrefix = "sample"
+    val participantId = SingleParticipant.entityId
+    val samples = createSamples(samplePrefix, participantId, 101)
+    val sampleSetMembership = createSampleSet(entityId, samplePrefix, 101)
+  }
+
+  private def createSamples(samplePrefix: String, participantId: String, numberOfSamples: Int): String = {
+    "entity:sample_id\tparticipant_id\n" +
+      new Array[String](numberOfSamples).zipWithIndex.map { el => samplePrefix + el._2 + "\t" + participantId + "\n" }.mkString.stripSuffix("\n")
```
rtitle: Doesn't seem like you're using the index so `zipWithIndex` is unneeded?

rtitle: Another way: `(0 until numberOfSamples).map(i => ...)`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/AuthenticatedPage.scala
```diff
@@ -0,0 +1,54 @@
...
+    private val signOutLink = testId("sign-out")
+
+    def checkAccountDropdown: Boolean = {
+      find(accountDropdown).isDefined
+    }
+
+    def clickAccountDropdown(): Unit = {
+      click on accountDropdown
+    }
+we
```
rtitle: rogue `we`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/AuthenticatedPage.scala
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    def readUserEmail(): String = {
+      await enabled accountDropdownEmail
+      find(accountDropdownEmail).get.text
+    }
+  }
+
+  /*
+   * This must be private so that subclasses can provide their own object
```
rtitle, line 47: Not sure this is true? This `ui`'s class name would be `org.broadinstitute.dsde.firecloud.page. AuthenticatedPage.ui` while any subclasses' would be `org.broadinstitute.dsde.firecloud.page.TheSubClass.ui` so it shouldn't conflict.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
```diff
@@ -0,0 +1,31 @@
...
+    * Sends the browser to the URL for this Page object. Returns the page
+    * object to facilitate call chaining.
+    *
+    * @param webDriver implicit WebDriver for the WebDriverWait
+    * @return the Page object
+    */
+  def open(implicit webDriver: WebDriver): P = {
+    go to this
+    awaitLoaded()
+    this
```
rtitle: Could `this` be removed (and just return the result of `awaitLoaded()`)?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceDataPage.scala
```diff
@@ -0,0 +1,113 @@
...
+    def clickCopyFromAnotherWorkspaceButton() = {
+      click on (await enabled copyFromAnotherWorkspaceButtonQuery)
+    }
+
+    def clickChooseFileButton() = {
+      click on (await enabled chooseFileButton)
+    }
+
+    def uploadData(filePath: String) = {
+      val strThatWorked = "/Users/ansingh/Desktop/ui-11/firecloud-ui/automation/src/test/scala/org/broadinstitute/dsde/firecloud/data/participants.txt"
```
rtitle: Remove

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspacePage.scala
```diff
@@ -0,0 +1,13 @@
...
+
+import org.broadinstitute.dsde.firecloud.page.AuthenticatedPage
+import org.openqa.selenium.WebDriver
+
+
+/*
+A Workspace Page is any page within the workspace (i.e. the Summary tab, Data tab)
+ */
+class WorkspacePage(implicit webDriver: WebDriver) extends AuthenticatedPage {
+
```
rtitle, line 11: Any shared code that could live here? I think all the urls start with `/workspaces/$namespace/$name`, right?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -0,0 +1,224 @@
...
+
+  override def awaitLoaded(): WorkspaceSummaryPage = {
+    await condition { enabled(testId("submission-status")) || enabled(testId("workspace-details-error")) }
+    this
+  }
+
+  /**
+    * Dictionary of access level labels displayed in the web UI.
+    */
+  object AccessLevel extends Enumeration {
```
rtitle, line 23: I feel like I already saw this enum somewhere else -- move to a shared location?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/WebBrowserSpec.scala
```diff
@@ -0,0 +1,129 @@
...
+  override def withScreenshot(f: => Unit)(implicit driver: WebDriver): Unit = {
+    try {
+      f
+    } catch {
+      case t: Throwable =>
+        val date = new SimpleDateFormat("yyyy-MM-dd_HH-mm-ss-SSS").format(new java.util.Date())
+        val fileName = s"failure_screenshots/${date}_$suiteName.png"
+        val htmlSourceFileName = s"failure_screenshots/${date}_$suiteName.html"
+        try {
+          val tmpFile = new Augmenter().augment(driver).asInstanceOf[TakesScreenshot].getScreenshotAs(OutputType.FILE)
```
rtitle, line 118: Crazy, scala code that takes a screenshot :)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
```diff
@@ -0,0 +1,124 @@
...
+          val workspaceMethodConfigPage = new WorkspaceMethodConfigPage(billingProjectName, workspaceName).open
+          val methodConfigDetailsPage = workspaceMethodConfigPage.importMethodConfigFromRepo(TestData.SimpleMethod.namespace,
+            TestData.SimpleMethod.name, TestData.SimpleMethod.snapshotId, methodConfigName)
+          methodConfigDetailsPage.editMethodConfig(inputs = Some(TestData.SimpleMethod.inputs))
+          val submissionDetailsPage = methodConfigDetailsPage.launchAnalysis(TestData.SimpleMethod.rootEntityType, TestData.SingleParticipant.entityId)
+
+          submissionDetailsPage.waitUntilSubmissionCompletes()
+          assert(submissionDetailsPage.verifyWorkflowSucceeded())
+        }
+
```
rtitle: Great tests!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/metadata/MethodConfigTabSpec.scala
```diff
@@ -0,0 +1,249 @@
...
+      methodConfigDetailsPage.editMethodConfig(inputs = Some(TestData.SimpleMethod.inputs))
+      methodConfigDetailsPage.deleteMethodConfig()
+
+      workspaceMethodConfigPage.filter(methodConfigName)
+      assert(find(methodConfigName).size == 0)
+    }
+  }
+
+
+  // negative tests
```
rtitle, line 237: who needs negative tests :)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: I played around with this a bit and didn't really like anything I came up with. The `asInstanceOf` bothered me, as did the verbosity of specifying the type at the call site for something that would likely be very common. I'd be willing to give it another try though.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -0,0 +1,199 @@
...
+import org.broadinstitute.dsde.firecloud.fixture.{GroupFixtures, WorkspaceFixtures}
+import org.broadinstitute.dsde.firecloud.page.workspaces.WorkspaceSummaryPage
+import org.broadinstitute.dsde.firecloud.test.{CleanUp, WebBrowserSpec}
+import org.scalatest._
+
+/*
+ * This test SHOULD be able to run with ParallelTestExecution. However, Rawls
+ * currently has database deadlock problems when creating and deleting managed
+ * groups because they involve 3 operations across 2 tables. My initial
+ * attempt to retry the deadlock failed because Rawls also creates the groups
```
rtitle, line 14: Interesting that this causes Rawls deadlocks - I wonder if this is a reproduction of https://broadinstitute.atlassian.net/browse/GAWB-2256..

automation/src/test/scala/org/broadinstitute/dsde/firecloud/util/Util.scala
```diff
@@ -0,0 +1,36 @@
...
+
+  def makeRandomId(length: Int = 7): String = {
+    Random.alphanumeric.take(length).mkString
+  }
+
+  def makeUuid: String = {
+    UUID.randomUUID().toString
+  }
+
+  def retry[T](remainingBackOffIntervals: Seq[FiniteDuration])(op: => Option[T]): Option[T] = {
```
rtitle: Hey here's a common `retry`! See my comments above.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Good feedback. Yes, this does currently also live in `Util.scala`, though I've had it in mind to pull it into a `Retry.scala` instead. Both are copied from/inspired by Rawls code. In fact, the version in `Util` has the changes that you suggested. It's possible that one was added before the scala class and the other was added after. :)

build.sbt
```diff
@@ -0,0 +1,10 @@
+import Settings._
+
+lazy val root = project.in(file("automation"))
+  .settings(rootSettings:_*)
+
+name := "root"
+
+version := "1.0"
+
+scalaVersion := "2.11.8"
```
rtitle, line 10: Ah... we should probably be using scala 2.12.2 for any new projects. Was there a specific reason to using 2.11?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Is that 2.12 only? I think we're on 2.11. I read up on enumeration patterns before I chose this one. Should be try to be on 2.12 if it would be easy to move?

project/Dependencies.scala
```diff
@@ -0,0 +1,53 @@
...
+    // proactively pull in latest versions of Jackson libs, instead of relying on the versions
+    // specified as transitive dependencies, due to OWASP DependencyCheck warnings for earlier versions.
+    "com.fasterxml.jackson.core" % "jackson-annotations" % jacksonV,
+    "com.fasterxml.jackson.core" % "jackson-databind" % jacksonV,
+    "com.fasterxml.jackson.core" % "jackson-core" % jacksonV,
+    "com.fasterxml.jackson.module" % "jackson-module-scala_2.11" % jacksonV,
+    "ch.qos.logback" % "logback-classic" % "1.1.3",
+    "com.google.apis" % "google-api-services-oauth2" % "v1-rev112-1.20.0",
+    "com.google.api-client" % "google-api-client" % "1.22.0" exclude("com.google.guava", "guava-jdk5"),
+//    "com.getsentry.raven" % "raven-logback" % "7.8.6",
```
rtitle, line 19: Remove commented dependencies, this is really hard to read

project/Dependencies.scala
```diff
@@ -0,0 +1,53 @@
...
+  //    "org.elasticsearch"    % "elasticsearch"  % "2.4.1",
+//    ("com.google.org.broadinstitute.dsde.firecloud.api-client" % "google-org.broadinstitute.dsde.firecloud.api-client" % "1.22.0").exclude("com.google.guava", "guava-jdk5"),
+//    "com.google.apis" % "google-org.broadinstitute.dsde.firecloud.api-services-storage" % "v1-rev58-1.21.0",
+//    "com.google.apis" % "google-org.broadinstitute.dsde.firecloud.api-services-compute" % "v1-rev120-1.22.0",
+//    "com.jason-goodwin"   %% "authentikat-jwt" % "0.4.1",
+//    "com.sun.mail"         % "javax.mail" % "1.5.6",
+//    "org.ocpsoft.prettytime" % "prettytime" % "4.0.1.Final",
+//    "org.everit.json"      %  "org.everit.json.schema" % "1.4.1",
+    "org.specs2"          %%  "specs2-core"   % "3.7"  % "test",
+    "org.scalatest"       %%  "scalatest"     % "2.2.6"   % "test",
```
rtitle, line 48: scalatest 3 would be nice too

project/Settings.scala
```diff
@@ -0,0 +1,55 @@
+import Dependencies._
+//import Merging._
```
rtitle, line 2: Please remove commented code in this file too

src/cljs/main/broadfcui/auth.cljs
```diff
@@ -62,7 +62,7 @@
      ;; Google's code complains if the sign-in button goes missing, so we hide this component rather
      ;; than removing it from the page.
      [:div {:style {:display (when (:hidden? props) "none") :marginTop "2rem"}}
-      [comps/Button {:text "Sign In" :onClick #(this :-handle-sign-in-click)}]
+      [comps/Button {:text "Sign In" :onClick #(this :-handle-sign-in-click) :data-test-id "sign-in-button"}]
```
rtitle: A more UI-literate person should probably review the cljs changes

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: I had that originally but didn't like how it looked when calling. Either the argument would be omitted or it would always be `Option["blah"]` which seemed redundant to me. The code is currently split between the 2 options but I definitely want to consistently use one or the other. Is there a strong argument for using an `Option`?

rtitle:
I made it! The tests look great. I made a bunch of comments, mostly in util/test code.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Rawls.scala
breilly2: The current use of this doesn't care if the result is undetermined (not `OK` or `NotFound`). However, I didn't want to assume that to be true of any future uses. I'm still working on getting the hang of when it's better to use `Option`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
rtitle: It's just the 2.12 docs, scala.Enumeration has existed forever

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/AuthenticatedPage.scala
breilly2: IntelliJ's static analysis doesn't complain, but compiling subclasses that have an `object ui ...` results in an error:
```
Error:(133, 10) overriding object ui in class AuthenticatedPage;
 object ui cannot override final member
  object ui extends UI
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
breilly2: This `AccessLevel` encodes the display values in the UI. The other one that you saw was `Orchestration.WorkspaceAccessLevel` which encodes the values that the API works with. Although the string values are the same in this case, they are conceptually different things. In fact, I wish that the display values were not YELLING.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
breilly2: I think `GoogleGroupSyncMonitorSupervisor` is about reading the groups as they're defined in Rawls and updating Google to match. Therefore, it should only be reading from the group tables.

However, I think the queue of synchronization actions is actually in the database at some point so maybe that's the cause of GAWB-2256. Either way, I think we're quickly approaching the need to audit how we handle database operations and transactions which will result in a non-trivial amount of tech debt cleanup.

build.sbt
breilly2: Just copied from other FireCloud projects. I guess this would be a good time to make a move toward 2.12.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Okay. I can add that. I think I may have backed away from the type aliasing because it wasn't intuitive to me how to find usages of the enum in IntelliJ. I see now that you have to find usages of the type rather than the object which I guess makes sense.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
breilly2: We lost some git history here, but I think code in `handleResponse` used to be inline in `requestWithJsonContent` or something like that.

@rtitle, the only issue with your proposed `sendRequestAndWaitForResponse` is that callers sometimes want to inspect the response rather than deal with one of a `String` response or an `APIException`. I like how it acts on futures, though, so I'm trying to work it into the API.

Dockerfile-tests
```diff
@@ -0,0 +1,12 @@
+FROM hseeberger/scala-sbt
```
zarsky-broad, line 1: Consider using an Alpine image. It'll take less space and less RAM, and it'll likely run faster.

zarsky-broad: (This goes for all Docker images, if there's an Alpine version or you can make one yourself, life will likely be better…)

src/cljs/main/broadfcui/common.cljs
```diff
@@ -356,7 +356,8 @@
                              :target (:target props)
                              :onMouseOver #(swap! state assoc :hovering? true)
                              :onMouseOut #(swap! state assoc :hovering? false)
-                             :onClick (:dismiss props)}
+                             :onClick (:dismiss props)
+                             :data-test-id (:data-test-id props)}
```
zarsky-broad: At this point, I'm going to recommend that you `:keys` out `props`.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -558,12 +565,13 @@
...
    (fn [{:keys [props]}]
-     (let [{:keys [header content ok-button show-cancel? cancel-text show-close?]} props
+     (let [{:keys [header content ok-button show-cancel? cancel-text show-close? data-test-id]} props
            cancel-text (or cancel-text "Cancel")]
        [:div {}
-        [:div {:style {:borderBottom style/standard-line
+        [:div (merge {:style {:borderBottom style/standard-line
                        :padding "20px 48px 18px"
                        :fontSize "137%" :fontWeight 400 :lineHeight 1}}
+                     (when (some? data-test-id) {:data-test-id data-test-id}))
```
zarsky-broad: Is this `when some?` necessary? If a prop is nil, react will just not render it.

zarsky-broad (changes requested):
I think I'm going to ask that all `:data-test-id` values be wrapped in a `(when config/debug?)`.
This only has to be done on `defc`-ed components themselves and on raw html elements, but it should prevent the prop getting rendered to production.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
breilly2: Actually, I'm not sure why the things below this (`Orchestration`, `Rawls`, `Thurloe`) are traits with companion objects instead of just straight-up objects. @rtitle can you think of any reason not to make `FireCloudClient` a trait and the rest objects?

Dockerfile-tests
breilly2: @zarsky-broad Is there a trusted source for these images? I didn't see an obvious one to trust (though I guess I'm equally skeptical of hseeberger and I don't know why we chose it here).

src/cljs/main/broadfcui/common.cljs
breilly2: @zarsky-broad can you check this to make sure I did it right? Thanks.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
rtitle: It's a scala best practice to use `Option` for optional values. IMO there's nothing wrong with the below code. I would say the `Some("domain")` parameter is actually useful because it informs the caller (via the type!) that the parameter is optional. Otherwise if it were a `String` the caller would be left guessing whether to put `""`, or `null`, etc..

```
def create(namespace: String, name: String, authDomain: Option[String] = None)
          (implicit token: AuthToken): Unit = {
  val authDomainMap = authDomain match {
    case None => Map()
    case Some(a) => Map("authorizationDomain" -> Map("membersGroupName" -> a))
  }
  ...
}

// Later:
create("namespace", "name", Some("domain"))
```

There's a lot of literature about why to use Option. Here's a random one I just found via Google: http://danielwestheide.com/blog/2012/12/19/the-neophytes-guide-to-scala-part-5-the-option-type.html

src/cljs/main/broadfcui/common/components.cljs
breilly2: Good call. Looks like the outer `merge` isn't needed either.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Gotcha. I still feel like using `Option` and a default parameter value is redundant -- if you don't have one, you don't have to pass in anything -- but I guess using `Some(...)` isn't too bad.

Dockerfile-tests
zarsky-broad: You might consider something like https://hub.docker.com/r/gafiatulin/alpine-sbt/, or you could roll your own. Since the Dockerfiles are public, there's no need for an "official" one as long as you're ok with how it's built.

src/cljs/main/broadfcui/page/workspace/monitor/common.cljs
```diff
@@ -41,25 +41,29 @@
 
 (def success-icon
   (icons/icon {:style {:color (:success-state style/colors) :marginRight 4
-                       :width table-style/table-icon-size :height table-style/table-icon-size}}
+                       :width table-style/table-icon-size :height table-style/table-icon-size}
+               :data-test-id "status-icon" :data-test-value "success"}
```
breilly2: I can not put `(config/when-debug ...)` around these values because, for some reason unknown to me, the config is consistently not yet loaded when these are rendered. Putnam suggested adding a watch to the config atom but I don't see how that's going to work in this context. I'm going to need some additional help fixing this. It would be great to fix this before merging but I'm not going to hold up the merge for this one case.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
rtitle: sounds reasonable to me

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/AuthenticatedPage.scala
rtitle: oh, ok

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
rtitle: +1 on auditing our database access logic, some of the code is a little scary

build.sbt
rtitle: I guess the Scala upgrade will come in a future iteration?

rtitle:
It looks like my comments have been addressed. 👍 !

rtitle (approved):

build.sbt
breilly2: Yes, unfortunately updating scala, scalatest, and project/dependency clean-up will have to come later. I definitely want to do that soon, though.

zarsky-broad (approved):

PR #854, opened 2017-07-19
Fix data loading

MattPutnam:
This broke during the "make data table not load a bunch of unnecessary AJAX" story.

### DIFF
--- a/src/cljs/main/broadfcui/common/components.cljs
+++ b/src/cljs/main/broadfcui/common/components.cljs
@@ -763,7 +763,7 @@
    (fn [{:keys [props state]}]
      (let [body
            [:div {:hidden (and (:collapsed? @state) (:label props))
-                  :marginLeft (if (:label props) "0.5rem" 0)}
+                  :style {:marginLeft (if (:label props) "0.5rem" 0)}}
             (map (fn [node]
                    [:ul {:style {:margin "0.2rem" :padding "0.5rem"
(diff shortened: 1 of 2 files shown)


src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -763,7 +763,7 @@
    (fn [{:keys [props state]}]
      (let [body
            [:div {:hidden (and (:collapsed? @state) (:label props))
-                  :marginLeft (if (:label props) "0.5rem" 0)}
+                  :style {:marginLeft (if (:label props) "0.5rem" 0)}}
```
MattPutnam, line 766: Also this was yelling at me so I shoehorned it in

davidangb (approved):
wow, I wish github was smarter about code blocks that moved.

👍 as TL, pending Isaac's review

zarsky-broad:
Sorry, where does `:rows` come from?

MattPutnam:
I'm not actually sure why `:rows` needs to be initialized to an empty vector, and maybe it doesn't. Either way, reinitializing doesn't need to clear it out.

zarsky-broad:
So do we not want to bother with trying stripping it out?
Your call. I'm 👍

zarsky-broad (approved):

MattPutnam:
In the interest of not rocking the boat while dev is broken, I'm leaving it.

PR #855, opened 2017-07-19
Make :toolbar property a function

dmohs:


src/cljs/main/broadfcui/common/table/table.cljs
```diff
@@ -145,7 +145,7 @@
                                  {:initial-text (:filter-text query-params)
                                   :on-filter #(swap! state update :query-params assoc :filter-text % :page-number 1)}
                                  (:inner filter-bar-props))]]))
-         (list* (:items toolbar))]
+         (list* ((:items toolbar) {:columns columns}))]
```
MattPutnam: I don't think this is right--`columns` is the static column definitions. `column-display`?

dmohs: Crap I think you're absolutely right.

PR #863, opened 2017-07-21
Old table deprecation notice

jmthibault79:
This is probably nowhere close to the actual text you would want here.  But I think some kind of in-code reminder about the Old Table would be helpful.

### DIFF
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -10,6 +10,13 @@
    [broadfcui.utils :as utils]
    ))
 
+;;;
+;;;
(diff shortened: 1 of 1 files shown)


dmohs (approved):
👍

jmthibault79:
Also: what do you think about a component and reference rename?

e.g. in the `:require` section of `submission_details.cljs`
`[broadfcui.common.table :refer [Table]]`
becomes
`[broadfcui.common.deprecated_table :refer [DeprecatedTable]]`

jmthibault79:
Meh, maybe not worth it since only 4 instances remain:
1. attribute_editor
2. entity_selector
3. entity_viewer
4. submission_details but see #870

PR #890, opened 2017-08-04
Fix existing auth domain spec test

MatthewBemis:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
@@ -175,7 +175,7 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
     private val authDomainSelect = testId("workspace-auth-domain-select")
     private val billingProjectSelect = testId("billing-project-select")
     private val cloneButtonQuery: Query = testId("create-workspace-button")
-    private val lockedAuthDomainGroupsQuery: Query = testId("locked-auth-domain-group")
+    private val authDomainGroupsQuery: Query = testId("selected-auth-domain-group")
     private val workspaceNameInput: Query = testId("workspace-name-input")
 
     def clickCloneButton(): Unit = {
@@ -187,9 +187,9 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
     }
 
     def readAuthDomainGroups(): List[(String, Boolean)] = {
-      await visible lockedAuthDomainGroupsQuery
+      await visible authDomainGroupsQuery
 
-      findAll(lockedAuthDomainGroupsQuery).map { element =>
+      findAll(authDomainGroupsQuery).map { element =>
         (element.attribute("value").get, element.isEnabled)
       }.toList
     }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
@@ -175,7 +175,7 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
     private val authDomainSelect = testId("workspace-auth-domain-select")
     private val billingProjectSelect = testId("billing-project-select")
     private val cloneButtonQuery: Query = testId("create-workspace-button")
-    private def lockedAuthDomainGroup(groupName: String): Query = testId("locked-auth-domain-group-%s".format(groupName))
+    private val lockedAuthDomainGroupsQuery: Query = testId(s"locked-auth-domain-group")
     private val workspaceNameInput: Query = testId("workspace-name-input")
 
     def clickCloneButton(): Unit = {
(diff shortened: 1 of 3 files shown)


automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -181,7 +181,7 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
 
             val cloneWorkspaceName = workspaceName + "_clone"
             val cloneModal = summaryPage.ui.clickCloneButton()
-            cloneModal.ui.readPresetAuthDomain() should be(Some(authDomainName))
+            cloneModal.ui.readPresetAuthDomainGroup(authDomainName) should be(Some(authDomainName))
```
breilly2: This test should also somehow reflect that the selected group can't be changed. One way would be to "look" at the element in question; see `WorkspaceListPage.UI.looksDisabled`. Another option, which would also help with support for multi-group auth domains, would be to have `readPresetAuthDomainGroup` take no arguments and return a List[String]. It could be implemented by putting the same `data-test-id` on all selects, using `findAll`/`filter` to select the disabled ones, and extracting the selected values.

Minor nit: I think a name like `readLockedAuthDomainGroups` would be slightly more descriptive. I didn't spend a lot of time naming `readPresetAuthDomain`. Now the redesigned UI with the lock icon seems to suggest "lock" is more informative than "preset".

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -175,7 +175,7 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
     private val authDomainSelect = testId("workspace-auth-domain-select")
     private val billingProjectSelect = testId("billing-project-select")
     private val cloneButtonQuery: Query = testId("create-workspace-button")
-    private val presetAuthDomain: Query = testId("required-auth-domain")
+    private def lockedAuthDomainGroup(groupName: String): Query = testId("locked-auth-domain-group-%s".format(groupName))
```
breilly2: Nit: I think `s"locked-auth-domain-group-$groupName` is a little more fluent and readable than using `String.format`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -186,8 +186,9 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
       textField(workspaceNameInput).value = name
     }
 
-    def readPresetAuthDomain(): Option[String] = {
-      find(presetAuthDomain).map(_.text)
+    def readPresetAuthDomainGroup(groupName: String): Option[String] = {
+      await visible lockedAuthDomainGroup(groupName)
```
breilly2: I was going to suggest that this could be inlined:

`singleSel(await visible lockedAuthDomainGroup(groupName)).selection`

Unfortunately, unlike `await.enabled`, `await.visible` does not return the found element. 😞  However, that can be fixed 😃  by changing the implementation of `visible` to be like `enabled` but filtering on `isDisplayed`.

That said, I'm pretty much 50/50 on which way reads better so it's up to you whether or not you want to make this change.

breilly2 (changes requested):
The only change that is necessary is somehow verifying that the inherited auth domain group cannot be changed. The rest of the suggestions are optional.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -175,7 +175,7 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
     private val authDomainSelect = testId("workspace-auth-domain-select")
     private val billingProjectSelect = testId("billing-project-select")
     private val cloneButtonQuery: Query = testId("create-workspace-button")
-    private val presetAuthDomain: Query = testId("required-auth-domain")
+    private val lockedAuthDomainGroupsQuery: Query = testId(s"locked-auth-domain-group")
```
breilly2: This doesn't need to be an interpolated string anymore. Sorry, I gave some conflicting feedback.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -186,8 +186,12 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
       textField(workspaceNameInput).value = name
     }
 
-    def readPresetAuthDomain(): Option[String] = {
-      find(presetAuthDomain).map(_.text)
+    def readLockedAuthDomainGroups(): List[(String, Boolean)] = {
```
breilly2: `locked-auth-domain-group` is only ever rendered on locked select lists and the name of this function implies that only locked (not enabled) ones are returned, so I think it's redundant to return `List[(String, Boolean)]`. `List[String]` should suffice.

I do like the idea of checking that the elements are actually disabled, though, so some kind of filter is worth keeping. One option would be to put some `data-` attribute on all select lists, renamed this function to `readAuthDomainGroups`, and implement `readLockedAuthDomainGroups` in terms of that.

I tried to pull together a suggested implementation and didn't like any of the results. There's probably an obvious solution that I'm not thinking of, but I can share what I have with you if you're interested.

breilly2 (changes requested):
Sorry... I'm going to be picky about getting the page class API right.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
MatthewBemis: oh crap i thought i removed the `s`. good eye

breilly2 (approved):
👍  Just checking with Jackie about why the test is still failing in jenkins.

PR #896, opened 2017-08-09
GAWB-2424: Automated tests for multi-group auth domains

MatthewBemis:
This diff also includes the diff for https://github.com/broadinstitute/firecloud-ui/pull/890

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/GroupFixtures.scala
```diff
@@ -12,6 +12,8 @@ import org.scalatest.Suite
   */
 trait GroupFixtures extends CleanUp with LazyLogging { self: WebBrowserSpec with Suite =>
 
+  def groupNameToEmail(groupName: String): String = s"GROUP_$groupName@quality.firecloud.org" //probably want to pull the apps domain from config if it's there
```
MatthewBemis: just realized i hadn't resolved this comment yet, i'll need to add this to `application.conf`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -80,17 +85,15 @@ trait Orchestration extends FireCloudClient with LazyLogging {
...
-    def create(namespace: String, name: String, authDomain: Option[String] = None)
+    def create(namespace: String, name: String, authDomain: Set[String] = Set.empty)
               (implicit token: AuthToken): Unit = {
       logger.info(s"Creating workspace: $namespace/$name authDomain: $authDomain")
 
-      val authDomainList = authDomain match {
-        case Some(a) => List(Map("membersGroupName" -> a))
-        case None => List()
-      }
+      val authDomainGroupsList = authDomain.map(a => Map("membersGroupName" -> a))
```
breilly2: Nit: it's actually not a `List` anymore... and I guess it's okay that it's a `Set` so it should probably be called just `authDomainGroups`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
-          withWorkspace(projectName, "AuthDomainSpec_share", Some(authDomainName)) { workspaceName =>
-            val listPage = signIn(Config.Users.ron)
-            listPage.filter(workspaceName)
-            listPage.ui.hasWorkspace(projectName, workspaceName) shouldEqual false
+      "when the user is not inside of the group" - {
+        "when the workspace is shared with them" - {
+          "can be seen but is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec") { authDomainName =>
+              withWorkspace(projectName, "AuthDomainSpec_reject", Set(authDomainName)) { workspaceName =>
+                api.workspaces.updateAcl(projectName, workspaceName, Config.Users.ron.email, WorkspaceAccessLevel.Reader)
```
breilly2: This should be able to be done by passing an `AclEntry` to `withWorkspace`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+
+                workspaceListPage.ui.clickWorkspaceInList(projectName, workspaceName)
+                // micro-sleep just long enough to let the app navigate elsewhere if it's going to, which it shouldn't in this case
+                Thread sleep 500
+                workspaceListPage.validateLocation()
+                // TODO: add assertions for the new "request access" modal
+              }
+            }
+          }
+          "when the user is a billing project owner" - {
```
breilly2: I think this should be outdented 1 level to be a sibling of "when the workspace is shared with them". Sharing is irrelevant when considering the billing project owner.

That said, I think we could also have 1 test that does share with the billing project owner to make sure that sharing doesn't break anything. I don't think there's a need to go crazy with all of the permutations involving multi-group auth domains in that case, though... maybe just a single test with a multi-group auth domain.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+                listPage.filter(workspaceName)
+                listPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
+
+                val summaryPage = listPage.openWorkspaceDetails(projectName, workspaceName).awaitLoaded()
+                summaryPage.ui.readAuthDomainGroups should include(authDomainName)
+              }
+            }
+          }
+          "when the user is a billing project owner" - {
+            "can be seen but is not accessible" in withWebDriver { implicit driver =>
```
breilly2: If the billing project owner is in the auth domain group, shouldn't they be able to access the workspace? I think this test should be including Hermione in the call to `withGroup`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+                val listPage = signIn(Config.Users.george)
+                listPage.filter(workspaceName)
+                listPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
+
+                val summaryPage = listPage.openWorkspaceDetails(projectName, workspaceName).awaitLoaded()
+                summaryPage.ui.readAuthDomainGroups should include(authDomainName)
+              }
+            }
+          }
+          "when the user is a billing project owner" - {
```
breilly2: As above, I think this should be outdented 1 level.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+                    listPage.filter(workspaceName)
+                    listPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
+
+                    val summaryPage = listPage.openWorkspaceDetails(projectName, workspaceName).awaitLoaded()
+                    summaryPage.ui.readAuthDomainGroups should include(groupOneName)
+                    summaryPage.ui.readAuthDomainGroups should include(groupTwoName)
+                  }
+                }
+              }
+            }
```
breilly2, line 477: A couple more cases to add:

1. Share with one of the groups of a multi-group auth domain. 2 users in both groups. User in shared group should be able to see and access the workspace. User in the other group. User in the other group should not be able to see or access it.

2. One test that shares with writer access and verify correct permissions. I guess make it a multi-group auth domain and, along with some non-auth domain sharing tests (to be written), we should have enough confidence that the remaining cases work.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/GroupFixtures.scala
breilly2: Yes, I can help with this if you'd like. We already need this in config to support running the tests against dev/alpha.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -186,8 +186,16 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
       textField(workspaceNameInput).value = name
     }
```
andy7i: In this file, there are 2x `cloneWorkspace` methods that take in `authDomain: Option[String]`. For consistency / future usability, could you make these use `Set`s?

(Lines 41 and 159; not sure how to add comments to those lines..)

src/cljs/main/broadfcui/page/workspace/create.cljs
```diff
@@ -118,6 +118,7 @@
                 [:div {:style {:float "left" :width "90%"}}
                  (style/create-identity-select-name
                   {:value opt
+                   :data-test-id (config/when-debug "selected-auth-domain-group")
                    :disabled (utils/seq-contains? locked-groups opt)
```
andy7i: When cloning a workspace with an auth domain, the UI displays these "locked groups". These have the same testID as new groups selected for the auth domain. It might be helpful to have a different testID for the locked groups (in the disabled dropdowns).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
```diff
@@ -186,8 +186,16 @@ class CloneWorkspaceModal(implicit webDriver: WebDriver) extends FireCloudView {
...
+    def readAuthDomainGroups(): List[(String, Boolean)] = {
+      await visible authDomainGroupsQuery
+
+      findAll(authDomainGroupsQuery).map { element =>
+        (element.attribute("value").get, element.isEnabled)
+      }.toList
+    }
+
+    def readLockedAuthDomainGroups(): List[String] = {
+      readAuthDomainGroups().filterNot{ case (_, isEnabled) => isEnabled }.map { case (name, _) => name }
```
andy7i: If a separate data-test-id is created for the locked groups (mentioned below), I think this test would be slightly more robust by using those.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
-      withGroup("AuthDomainSpec") { authDomainName =>
-        withCleanUp {
-          val workspaceListPage = signIn(Config.Users.fred)
-
-          val workspaceName = "AuthDomainSpec_create_" + randomUuid
-          register cleanUp api.workspaces.delete(projectName, workspaceName)
-          val workspaceDetailPage = workspaceListPage.createWorkspace(projectName, workspaceName, Option(authDomainName)).awaitLoaded()
+            val workspaceName = "AuthDomainSpec_create_" + randomUuid
+            register cleanUp api.workspaces.delete(projectName, workspaceName)
+            val workspaceDetailPage = workspaceListPage.createWorkspace(projectName, workspaceName, Set(authDomainName)).awaitLoaded()
```
andy7i: Is `workspaceSummaryPage` a better name?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
 
-          workspaceDetailPage.ui.readAuthDomainGroups should include (authDomainName)
+            workspaceDetailPage.ui.readAuthDomainGroups should include(authDomainName)
 
-          workspaceListPage.open.filter(workspaceName)
-          workspaceListPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
+            workspaceListPage.open.filter(workspaceName)
+            workspaceListPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
```
andy7i: This is the second assertion in this testcase -- we might want that as a separate test.

breilly2 (changes requested):
I thought I had submitted this on Friday but maybe I didn't. I guess the most important changes are the test indentation, the billing project owner test that didn't look right, and filing in a couple additional cases.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
 
 
   "A workspace with an authorization domain" - {
+    "with one group inside of it" - {
+      "can be created" in withWebDriver { implicit driver =>
+        withGroup("AuthDomainSpec") { authDomainName =>
```
andy7i: Does `withGroup` need the user's email to add the user to the group?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,468 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+              withWorkspace(projectName, "AuthDomainSpec_reject", Set(authDomainName)) { workspaceName =>
+                api.workspaces.updateAcl(projectName, workspaceName, Config.Users.ron.email, WorkspaceAccessLevel.Reader)
+
+                val workspaceListPage = signIn(Config.Users.ron)
+                workspaceListPage.filter(workspaceName)
+                workspaceListPage.ui.looksRestricted(projectName, workspaceName) shouldEqual true
+
+                workspaceListPage.ui.clickWorkspaceInList(projectName, workspaceName)
+                // micro-sleep just long enough to let the app navigate elsewhere if it's going to, which it shouldn't in this case
+                Thread sleep 500
```
andy7i: I realize that there was an existing testcase with `sleep`. Just expressing my sadness about the sleep statements.

andy7i: Could we validate against whether the "request access" modal appears? This might be preferable for the testcases where we expect to stay on the workspace list page. On the flip side, I'm not sure if it's reasonable to validate against the modal NOT appearing.

andy7i: Actually ^ would just be validating that the workspace summary page has loaded.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
andy7i: Partially answering my own question, I see that if a user creates a group, that user is part of the group. It looks like the API uses an implicit auth token. The implicit authToken is declared on line 29.  Just to be sure, is this what we want to do?

andy7i:
If a workspace is made public, is it essentially the same as that workspace being shared with a group that includes all users?

MatthewBemis:
@andy7i yes- when we make a workspace public what we actually do under the hood is make the "All Users" group a reader on that workspace

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
MatthewBemis: yes this was intended, it also signs in as that same user when creating the workspace

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
MatthewBemis: definitely. that name came from pre-existing code so i'll see if there are other areas to change along with it

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/WorkspaceSummaryPage.scala
breilly2: I disagree somewhat... not necessarily with using a separate data-test-id but that it would be more robust. Regardless of the data-test-id, tests need to verify that the groups are locked and cannot be changed; there's no guarantee that they are actually locked just be cause the data-test-id implies it. At that point it doesn't really matter what the data-test-id is as long as it selects the right element(s).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
MatthewBemis: i put these two cases in the same test for the sake of speed. since they're both cosmetic assertions, it seemed wasteful to create and tear down a workspace to test something cosmetic that was already visible to the test. but maybe that's not best-practice. 

i'm happy to break them apart if we want this level of granularity and don't care about the impact on speed

breilly2: I definitely understand why. I do worry about the potential explosion of test cases that this split up could cause. `looksRestricted` is used in 14 test cases in `AuthDomainSpec`.

Maybe not all of those need to be there. Maybe we could have one test that reads the auth domain groups on the summary page, one that checks that it looks restricted on the list page, then the remaining cases that assert based on access/visibility and don't actually "read" the view like the first 2.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
breilly2: This was written before the "request access" modal existed. I agree that we should update it to be a better test.

I agree that we should only use `sleep` as an absolute last resort and we should always question its usage when we see it.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
MatthewBemis: i'm replacing all of the billing project owner tests with these three tests:

when there is a multi-group auth domain and it is shared with a billing project owner who is not in all of those groups, it should be seen but not accessible

when there is a multi-group auth domain and it is shared with a billing project owner who is in all of those groups, it should be seen and accessible

when there is a multi-group auth domain and it is not shared with a billing project owner, it should be seen but not accessible

i think that covers all of the real cases. the single-group auth domain tests were just redundant

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -28,169 +30,409 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
         }
       }
-    }
 
-    "when not shared with a user who is not in the authorization domain" - {
+      "can be cloned and retain the auth domain" in withWebDriver { implicit driver =>
+        withGroup("AuthDomainSpec", List(Config.Users.george.email)) { authDomainName =>
+          withWorkspace(projectName, "AuthDomainSpec_share", Set(authDomainName)) { workspaceName =>
+            withCleanUp {
+              api.workspaces.updateAcl(projectName, workspaceName,
```
breilly2: Not sure if I commented on this somewhere else, but this should be able to be done by passing an `AclEntry` to `withWorkspace`.

MatthewBemis: you did- i had updated that case but failed to update the 5 other places where this was done. i've taken care of all of them now

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+        withGroup("AuthDomainSpec", List(Config.Users.george.email)) { groupOneName =>
+          withGroup("AuthDomainSpec", List(Config.Users.george.email)) { groupTwoName =>
+            withGroup("AuthDomainSpec", List(Config.Users.george.email)) { groupThreeName =>
+              withWorkspace(projectName, "AuthDomainSpec_share", Set(groupOneName, groupTwoName), List(AclEntry(Config.Users.george.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+                withCleanUp {
+                  val listPage = signIn(Config.Users.george)
+                  val summaryPage = listPage.openWorkspaceDetails(projectName, workspaceName)
+
+                  val cloneWorkspaceName = workspaceName + "_clone"
+                  val cloneModal = summaryPage.ui.clickCloneButton()
```
breilly2: Since another test already checked that the groups cannot be removed when cloning, this test could simply use `WorkspaceSummaryPage.cloneWorkspace` instead of driving the modal itself. Not a big deal either way but you could simplify this if you want to. Just make sure to only pass the new group in since the existing groups will be preselected.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+          "can be seen but is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec") { groupOneName =>
+              withGroup("AuthDomainSpec") { groupTwoName =>
+                withWorkspace(projectName, "AuthDomainSpec_reject", Set(groupOneName, groupTwoName), List(AclEntry(Config.Users.ron.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+                  val workspaceListPage = signIn(Config.Users.ron)
+                  workspaceListPage.filter(workspaceName)
+
+                  workspaceListPage.ui.clickWorkspaceInList(projectName, workspaceName)
+                  // micro-sleep just long enough to let the app navigate elsewhere if it's going to, which it shouldn't in this case
+                  Thread sleep 500
```
breilly2: Should be able to replace this sleep with `workspaceListPage.ui.showsRequestAccessModal shouldEqual true`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+        "when the workspace is shared with them" - {
+          "can be seen but is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec") { authDomainName =>
+              withWorkspace(projectName, "AuthDomainSpec_reject", Set(authDomainName), List(AclEntry(Config.Users.ron.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+                val workspaceListPage = signIn(Config.Users.ron)
+                workspaceListPage.filter(workspaceName)
+
+                workspaceListPage.ui.clickWorkspaceInList(projectName, workspaceName)
+                // micro-sleep just long enough to let the app navigate elsewhere if it's going to, which it shouldn't in this case
+                Thread sleep 500
```
breilly2: Should be able to replace this sleep with `workspaceListPage.ui.showsRequestAccessModal shouldEqual true`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
 
-            register cleanUp { api.workspaces.delete(projectName, cloneWorkspaceName)(AuthTokens.george) }
-            cloneModal.cloneWorkspace(projectName, cloneWorkspaceName)
-            cloneModal.cloneWorkspaceWait()
-            summaryPage.cloneWorkspaceWait()
+      "when the user is in one of the groups" - {
+        "when shared with them" - {
+          "can be seen but is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec") { groupOneName =>
+              withGroup("AuthDomainSpec", List(Config.Users.george.email)) { groupTwoName =>
```
breilly2: I think it should be Ron in this group (or share with and sign in as George).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+                  workspaceListPage.validateLocation()
+                }
+              }
+            }
+          }
+        }
+        "when not shared with them" - {
+          "cannot be seen and is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec") { groupOneName =>
+              withGroup("AuthDomainSpec", List(Config.Users.george.email)) { groupTwoName =>
```
breilly2: I think it should be Ron in this group as well.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+                    }
+                  }
+                }
+              }
+            }
+          }
+          "when the user is a billing project owner" - {
+            "can be seen and is accessible" in withWebDriver { implicit driver =>
+              withGroup("AuthDomainSpec", List(Config.Users.hermione.email)) { authDomainName =>
+                withWorkspace(projectName, "AuthDomainSpec_share", Set(authDomainName)) { workspaceName =>
```
breilly2: Should this have a multi-group auth domain?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -26,171 +28,460 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
+              }
+            }
+          }
+        }
+        "when not shared with them" - {
+          "cannot be seen and is not accessible" in withWebDriver { implicit driver =>
+            withGroup("AuthDomainSpec", List(Config.Users.harry.email)) { groupOneName =>
+              withGroup("AuthDomainSpec", List(Config.Users.harry.email)) { groupTwoName =>
+                withWorkspace(projectName, "AuthDomainSpec_reject", Set(groupOneName, groupTwoName)) { workspaceName =>
+                  val workspaceListPage = signIn(Config.Users.george)
```
breilly2: Should sign in as Harry here.

breilly2 (changes requested):
Sorry... a few more changes to make.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
MatthewBemis: i'm so mad that i missed these somehow

MatthewBemis:
@breilly2 back to you

breilly2 (approved):
👍 Thanks for sticking with it! I know this was a bit of work to get right. There is a lot of value in having these tests.

PR #903, opened 2017-08-14
fix automated method config tests

ahaessly:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methods/MethodConfigDetailsPage.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methods/MethodConfigDetailsPage.scala
@@ -60,7 +60,6 @@ class MethodConfigDetailsPage(namespace: String, name: String, methodConfigNames
 
     def openLaunchAnalysisModal(): LaunchAnalysisModal = {
       await enabled methodConfigNameTextQuery
-      //await thenClick(openLaunchAnalysisModalButtonQuery)
       click on (await enabled openLaunchAnalysisModalButtonQuery)
       new LaunchAnalysisModal
     }
@@ -159,10 +158,6 @@ class LaunchAnalysisModal(implicit webDriver: WebDriver) extends FireCloudView {
     ui.clickLaunchButton()
   }
 
-  def verifyNoDefaultEntityMessage(): Boolean = {
-    ui.isNoDefaultEntitiesMessagePresent()
-  }
-
   def verifyWorkflowsWarning(): Boolean = {
     ui.isNumberOfWorkflowWarningPresent()
   }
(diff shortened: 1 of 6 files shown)


breilly2:
@ahaessly Do you know why so many tests failed in jenkins? https://fc-jenkins.dsp-techops.broadinstitute.org/job/fiab-test/146/consoleFull Ideally this PR would result in a successful test build so I'm especially interested in seeing that happen. There's definitely some strangeness in the output, such as `javax.net.ssl.SSLProtocolException: handshake alert:  unrecognized_name` errors. Maybe this branch doesn't have all of the latest config fixes?

src/cljs/main/broadfcui/common/entity_table.cljs
```diff
@@ -153,7 +153,9 @@
          (if (empty? entity-types)
            (on-done {:total-count 0 :filtered-count 0 :results []})
            (let [{:keys [page-number rows-per-page filter-text sort-column sort-order]} query-params
-                 entity-type (nth entity-types selected-filter-index)]
+                 entity-type (if (nil? selected-filter-index)
```
breilly2, line 156: Is this the change that avoids the javascript error that was popping up? Looks like it now lists entities of the "first" type if nothing is selected.

This is fine since this is the only place in scope that uses `selected-filter-index`. Another option that might be slightly cleaner would be to set `selected-filter-index` to 0 if it's `nil`. That would encode the "nothing selected means first" logic in a single place that all uses of `selected-filter-index` could benefit from. Since it's only used here, this is definitely a nit than something that needs to be changed.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methods/MethodConfigDetailsPage.scala
```diff
@@ -220,9 +216,9 @@ class LaunchAnalysisModal(implicit webDriver: WebDriver) extends FireCloudView {
       click on (await enabled launchAnalysisButtonQuery)
     }
 
-    def isNoDefaultEntitiesMessagePresent(): Boolean = {
-      await enabled emptyDefaultEntitiesMessageQuery
-      find(emptyDefaultEntitiesMessageQuery).size == 1
+    def getSelectedEntityText(): String = {
+      await enabled selectedEntityQuery
+      find(selectedEntityQuery).fold("")(_.text)
```
breilly2, line 221: I go back and forth between using `fold` and `map ... getOrElse` for `Option`. I should probably just get used to `fold`.

breilly2 (approved):
👍 Looks good. I do want to get to the bottom of the test failures in jenkins, but if it's a transient problem then it should be fine to just see the successful build on merge to develop.

ansingh7115 (approved):
Looks good to me, thanks for fixing this

src/cljs/main/broadfcui/common/entity_table.cljs
ahaessly: i did initially try that, but then when the entites render, the first entity is selected even though it is not of the type expected. I think the behavior was not to have any selected.

PR #906, opened 2017-08-15
GAWB-2104: Split "Public" and "My" Workspaces

MattPutnam:
Also did some work to use the new tabs everywhere, and make the 3 main pages look the same.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


rushtong:
When first navigating to the Method Repository, none of the sub-tabs were selected. The "All (XX)" should have been, but wasn't. Selecting any tab seems to reset the behavior correctly so that it will navigate to the right one on subsequent reloads. I'm guessing this has to do with local storage, but even after deleting my local storage, I could not replicate the original, unselected sub-tab condition.

rushtong:
This needs a rebase, especially to pull in the shield-icon fix.

rushtong:
For the jira ticket, please list each of the modified pages that need to be QA-ed in the QA section since the tab-style changes are more in-depth than the story describes.

rushtong:
I'm confused by the counts that show in the workspaces tab when a filter is applied. In my example, I have 22 "My" workspaces and 23 "Public" workspaces. Applying a filter that reduces the number displayed to a single one looks strange. The "My" is empty, but shows a count of 22. The "Public" has 1, but shows a count of 23. Logically, I know that the counts are not reflected in the filter, but the visual effect is off. For example, if I leave that filter in place, leave, come back, and see the page again, it takes a split-second to realize what's happening.

Suggestion: I think the counts should update to reflect the filtered effect. Something like "My Workspaces (0 visible of 22 total)". I think this might also require PO feedback.

screen shot 2017-08-17 at 8 03 07 am

Edit: The behavior with filter vs facet is slightly different. When a facet is applied, the counts *do* update. Additionally, in combination, the count differences are even more jarring. Here is a screen with filter and facet selections with counts that only apply to the facet selection, but not the filter selection.

screen shot 2017-08-17 at 8 07 19 am

rushtong:
In the condition where there are no rows in a table, the alignment of the background color block with the tabs is now more apparent compared to when the tabs didn't have a background. It now looks a bit jagged. Is it possible to align these? Not a big deal, but I think it would be cleaner if the right edge of the empty table aligned with the right edge of the tabs block above it. Screenshot shows old over the new.

screen shot 2017-08-17 at 8 17 09 am

src/cljs/main/broadfcui/page/method_repo/method_config_importer.cljs
```diff
@@ -312,28 +312,33 @@
...
                         :name name
                         :snapshot-id snapshotId}
                     type (if (= entityType "Configuration") :method-config :method)]
                 (links/create-internal
-                 {:data-test-id (config/when-debug (str name "_" snapshotId))
-                  :href (if workspace-id "javascript:;" (nav/get-link type id))
-                  :onClick (when workspace-id
-                             #(swap! state assoc :type type :id id))}
-                 (style/render-name-id name snapshotId))))
+                  {:data-test-id (config/when-debug (str name "_" snapshotId))
```
rushtong, line 327: I think this is used for test automation - is that right?

rushtong (changes requested):
Overall, looks really good! 
* I have one request about counts that I think should be addressed in some way.
* My request for table alignment is minor - don't let that hold things up.

src/cljs/main/broadfcui/page/method_repo/method_config_importer.cljs
ahaessly: yes, it is

ahaessly (approved):
👍

ahaessly:
saw a minor issue - if the browser is resized to something narrow the table on the launch analysis screen extends past the boundary
screen shot 2017-08-18 at 2 59 17 pm

rushtong (approved):

PR #914, opened 2017-08-21
GAWB-2031

MattPutnam:
Implemented the new method config I/O tables, and used them in the method config edit page and the config detail view inside the method repo. This also touched tooltip generation, so I checked every column of every table in the application, and fixed a few problems.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off


src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -685,16 +685,18 @@
...
       :allow-new? true
+      :allow-clear? false
+      :multiple true
       :minimum-input-length 3})
    :render
    (fn [{:keys [props]}]
      (style/create-identity-select {:ref "input-element"
                                     :defaultValue (:tags props)
-                                    :multiple true}
+                                    :multiple (:multiple props)}
```
zarsky-broad: should we revert this line now?

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -709,17 +711,40 @@
...
+   ;; React can't re-render since select2 clobbers the DOM...
+   :should-component-update
    (constantly false)
+   ;; ...but we do need to patch in new client-side options
+   :component-will-receive-props
+   (fn [{:keys [refs props next-props this]}]
+     (let [new-data (:data next-props)]
+       (when (not= (:data props) new-data)
+         (let [selected (this :get-tags)
+               selected-list (if (sequential? selected) selected [selected])
```
zarsky-broad: `selected` will always be sequential

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -0,0 +1,139 @@
...
+   :save
+   (fn [{:keys [state locals]}]
+     (swap! state dissoc :editing?)
+     @locals)
+   :render
+   (fn [{:keys [props state this]}]
+     (let [id (gensym "io-table-")]
+       [:div {:id id :style (:style props)}
+        (when (:editing? @state)
+          [:style {} (str ".select2-results__option" ;; TODO - just make this global?
```
rushtong, line 39: remove todo?

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -196,86 +190,47 @@
...
+                                workspace-attributes (get-in props [:workspace :workspace :workspace-attributes])
+                                selected-entity-type (get-in loaded-config [:methodConfiguration :rootEntityType])]
+                            (swap! locals assoc :entities-loaded? true)
+                            (swap! state assoc
+                                   :blocker nil
+                                   :entity-types entity-types
+                                   :autocomplete-list (build-autocomplete-list
+                                                       (utils/restructure workspace-attributes entity-types selected-entity-type)))
+                            (this :-begin-editing))
+                          ;; FIXME: :data-attribute-load-error is unused
```
rushtong, line 252: We should probably fix this at some point. Do we need a new story for it?

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -196,86 +190,47 @@
...
-                              :margin "0 0.5rem 0.5rem 0" :padding "0.5rem"
-                              :backgroundColor (:background-light style/colors)
-                              :border style/standard-line :borderRadius 2}}
-                (str name ": (" (when optional "optional ") type ")")]
-               (when (and error (not editing?) (not optional))
-                 (icons/icon {:style {:marginRight "0.5rem" :alignSelf "center"
-                                      :color (:exception-state style/colors)}}
-                             :error))
-               (when editing?
-                 [comps/Typeahead {:ref (str ref-prefix "_" name)
```
rushtong, line 266: The only other place we use `Typeahead` is in the library wizard. Wondering if we should add a story for migrating that to `TagAutocomplete` and removing  `Typeahead`.

rushtong (approved):
👍  Looks good, works good. Minor questions inline.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: I would be in favor of that.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: Yeah, I'll file it.

src/cljs/main/broadfcui/common/method/config_io.cljs
MattPutnam: Make global, as in add a .css file to make all `select2`s styled like that.

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -0,0 +1,139 @@
...
+                                          [:div {:style (clip table-style/table-cell-plank-left)} task])}
+                               {:header "Variable" :initial-width 180
+                                :as-text (fn [{:keys [variable optional?]}]
+                                           (str variable (when optional? " (optional)")))
+                                :sort-by :text
+                                :render (fn [{:keys [variable optional?]}]
+                                          [:div {:style (clip (if optional?
+                                                                (merge table-style/table-cell-plank-middle
+                                                                       table-style/table-cell-optional)
+                                                                table-style/table-cell-plank-middle))} variable])}
```
davidangb, line 93: visually, I think this column could use a left margin/padding. Update: this must have gotten fixed in the last 1-2 merge commits, I no longer see it.

src/cljs/main/broadfcui/common/components.cljs
```diff
@@ -757,7 +764,8 @@
        (if (.-loading res)
          "Loading..."
          (let [show-counts? (:show-counts? props)
-               tag-text (.createTextNode js/document (or (.-tag res) (.-text res)))
+               ;; Insert zero-width space after underscore to allow line wrapping
+               tag-text (.createTextNode js/document (string/replace-all (or (.-tag res) (.-text res)) #"\_" "_\u200b"))
```
davidangb: we're allowing wrapping on underscores? That feels wrong.

src/cljs/main/broadfcui/common/method/config_io.cljs
davidangb: +1 to global. I know we add/remove dom elements all over the place, but it feels weird to add/remove a <style> tag depending on whether or not the user is editing.

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -0,0 +1,139 @@
...
+                                          (utils/restructure task variable error-message value)
+                                          {:type (process-type (or inputType outputType))
+                                           :optional? optional})))))
+               :body {:empty-message (str "No " (string/capitalize (name io-key)))
+                      :style (merge
+                              table-style/table-light
+                              {:body-cell {:padding 0}
+                               :header-row {:borderBottom style/standard-line}
+                               :body-row (constantly {:margin "4px 0" :alignItems "baseline"})})
+                      :behavior {:filterable? false :reorderable-columns? false :allow-no-sort? true
```
davidangb, line 76: until we order inputs/outputs by the order they exist in the WDL, could we give an initial sort by task, then variable?

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -0,0 +1,139 @@
...
+                                           :minimum-input-length 0
+                                           :tags [value] :maximum-selection-length 1
+                                           :language {:maximumSelected (fn [] "")}
+                                           ;; `value` ensures that custom selections are initially selected when going to edit
+                                           ;; `distinct` because having multiple copies of the same screws things up
+                                           :data (distinct (if value (conj data value) data))
+                                           :placeholder (if optional? "Optional" "Select or enter")
+                                           :on-change (fn [value]
+                                                        (swap! locals update io-key assoc (keyword name)
+                                                               (if (empty? value) "" (first value))))}]
```
davidangb, line 127: 1. enter a value for an attribute, like "workspace.foo", then hit enter. "workspace.foo" is now my value.
2. click after "foo". Note that you get a cursor and are allowed to keep typing.
3. add "bar" so that you now have "workspace.foobar". Try to save this value - it doesn't work, and you revert back to "workspace.foo".

I think we should either allow users to edit the attribute, or require them to delete/re-add - but in the latter case we shouldn't give them a live cursor making them think they can edit it.

src/cljs/main/broadfcui/endpoints.cljs
```diff
@@ -500,8 +500,9 @@
        :entityType "Task"})
     (range (rand-int 50)))})
 
-(defn get-configuration [namespace name snapshot-id]
-  {:path (str "/configurations/" namespace "/" name "/" snapshot-id)
+(defn get-configuration [namespace name snapshot-id & [payload-as-object?]]
+  {:path (str "/configurations/" namespace "/" name "/" snapshot-id
+              (when payload-as-object? "?payloadAsObject=true"))
```
davidangb, line 512: not for this PR; for some future: we should probably be using some kind of Uri library that allows adding/removing query params, instead of concatenating strings ourselves.

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
davidangb: +1 please write the story!

davidangb (commented):
a few comments inline

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: Filed as GAWB-2544

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: Filed as GAWB-2544

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: Created as GAWB-2553

src/cljs/main/broadfcui/common/components.cljs
MattPutnam: Otherwise some of the attribute names are monstrously long and force a horizontal scroll

src/cljs/main/broadfcui/common/method/config_io.cljs
zarsky-broad: Honestly, the best solution to this is to roll our own typeahead for situations like this. We're already massively hacking the tag feature of this one. I don't think we have a solution to this problem in particular.

zarsky-broad: (it thinks you're trying to add a second tag, rather than edit the one in place)

src/cljs/main/broadfcui/common/components.cljs
davidangb: I think we'll still have the same problem if someone uses periods in a super long attribute name. Instead of special handling for underscores, how about a `word-break: break-all` CSS style?

ninja edit: https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/ for reference

src/cljs/main/broadfcui/common/method/config_io.cljs
MattPutnam: Actually, turns out this is already overriding a global. Then a lot of this is for making it look like a typical typeahead, and not a select2.

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -0,0 +1,139 @@
...
+                              {:body-cell {:padding 0}
+                               :header-row {:borderBottom style/standard-line}
+                               :body-row (constantly {:margin "4px 0" :alignItems "baseline"})})
+                      :behavior {:filterable? false :reorderable-columns? false :allow-no-sort? true
+                                 :sortable-columns? (not editing?)}
+                      :columns
+                      (concat [{:header "Task" :initial-width 180
+                                :column-data :task :sort-initial :asc
+                                :render (fn [task]
+                                          [:div {:style (clip table-style/table-cell-plank-left)} task])}
```
davidangb: what do you think of:
```
{:header "Task" :initial-width 180
                                :sort-initial :asc
                                :sort-by (fn [{:keys [task variable]}] (str task "/" variable))
                                :render (fn [{:keys [task]}] [:div {:style (clip table-style/table-cell-plank-left)} task])}
```
this gets the user a nice initial sort by task then variable ... the tradeoff is that clicking the Task column to sort asc/desc looks a little weird b/c it's also sorting on variable.

Might also consider making both task and variable sort case-insensitive.

davidangb (approved):
👍 assuming this passes PO review

PR #949, opened 2017-09-20
integrating the test checklist into the PR checklist

andy7i:
Feedback is happily accepted.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,7 +24,7 @@
   * Submitter updates documentation as needed
   * Submitter **reassigns to LR** for further feedback
 - [ ] **TL** sign off
-- [ ] **LR** sign off\
+- [ ] **LR** sign off
 - [ ] **Product Owner** sign off
 - [ ] **Submitter**: Verify all tests go green, including CI tests and automated UI tests.
 - [ ] **Submitter**: Squash commits and merge to develop. If adding test code, merge application code and test code at the same time.
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -15,18 +15,6 @@
   * Audit trails
 - [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
 - [ ] **Submitter**: If you're adding new automated UI tests, review the test plan with QA
-* Guidelines for new automation tests:
-  * Tests should only verify one behavior. Tests should be simple and as short as possible, for the purpose of maintainability, readability, and speed.
-  * Every test should be able to run independently (no dependency to other tests) and in parallel (does not interfere with other tests).
-  * Time.sleep() is deeply frowned upon.
-  * Elements are selected by unique ID-locators (the attribute “data-test-id”).
-  * The test should use the Page Object Model, including:
-    * Clean separation between test code (which is high level) and page specific code (such as locators / element).
-    * Use proper scoping for functions and variables.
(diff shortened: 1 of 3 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,6 +14,19 @@
   * Encryption
   * Audit trails
 - [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
+- [ ] **Submitter**: If you're adding new automated UI tests, review the test plan with QA
+* Guidelines for new automation tests:
```
breilly2: Are these guidelines in a separate doc somewhere that can be linked to? I feel like this is an unprecedented amount of content for the checklist.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,6 +14,19 @@
...
+  * Tests should only verify one behavior. Tests should be simple and as short as possible, for the purpose of maintainability, readability, and speed.
+  * Every test should be able to run independently (no dependency to other tests) and in parallel (does not interfere with other tests).
+  * Time.sleep() is deeply frowned upon.
+  * Elements are selected by unique ID-locators (the attribute “data-test-id”).
+  * The test should use the Page Object Model, including:
+    * Clean separation between test code (which is high level) and page specific code (such as locators / element).
+    * Use proper scoping for functions and variables.
+    * Declare element ID-locators only once, and as a constant.
+  * The test should cleanup after itself.
+  * Keep in mind that this test will be required to pass for every subsequent build. If the test is flaky, create a dev ticket to investigate and fix the flakiness.
```
aednichols: I think tests should meet a very high standard of reliability before we check them in.

Just a few unreliable tests can screw over the entire suite... e.g. if we add 5 tests that are only 95% reliable, that gets us a ~1/4 false negative rate for the suite.

.github/PULL_REQUEST_TEMPLATE.md
rushtong: 👍  to Brian's comment. These should all be in a README in the automation directory since we now open up that as a distinct and separate intellij project.

.github/PULL_REQUEST_TEMPLATE.md
andy7i: How do you see that being implemented? Would the user run it multiple times? And/or would it run multiple times on PR creation?

.github/PULL_REQUEST_TEMPLATE.md
andy7i: I had a similar concern. I'll move it. Thanks!

.github/PULL_REQUEST_TEMPLATE.md
andy7i: Or are you saying that this text should say something like:
  * Keep in mind that this test will be required to pass for every subsequent build. If the test is flaky, investigate and fix the flakiness before merging.

breilly2: We should change the wording here to avoid implying that you're allowed to write a flaky test as long as you create a ticket for fixing it. Rather we should always do our best to write tests that aren't flaky and they should not be merged if observed to be flaky. This is difficult and flakiness may not surface until after something makes it to dev.

Tickets should be created any time we discover flaky tests as long as they're not so prohibitively flaky that they must be fixed right away. This is more of an ongoing practice than a guideline to follow while writing tests. If I'm understanding @aednichols's concern, moving that sentence to some other documentation should address it.

breilly2 (approved):
👍  Excellent. Thanks for doing this!

PR #970, opened 2017-09-29
GAWB-2552: delete the right attributes

MattPutnam:
### Tedium corner

The problem was not that the wrong rows were getting deleted. The problem was that the wrong rows were being displayed in edit mode. A subtle but important distinction. The workspace attributes table is (perhaps foolishly) designed to work in both normal mode (with two columns, key and value, which are both sortable) and edit mode (with four columns; delete, key, value, and type; and the table is not sortable).

*As you know*, on re-render, React tries to detect what has actually changed in order to avoid re-rendering things that are still the same. However, it stops searching and concludes that there's no change if it can't find any changes before some unknown depth. The table component has a lot of depth. So we have to force a refresh through the use of a key.

I assumed that if the key changed, then the component would fully re-render, and that was that. However, it seems that React actually remembers past keys and will give you back the old component if the key returns to a former value.  We were keying on [editing boolean]+[row count], which would indeed re-render if you switch modes or add/delete rows. But when you add a row, save, then delete a row, you end up at a previously-used key. Combine with the fact that the editing version of the table shows attributes in the arbitrary order returned by the server, and you have a mess of confusion.

Ultimately, the *n*th row delete button was correctly deleting the *n*th row from the in-memory data, and subsequent "Save"ing correctly updated the server data, but the table wasn't displaying the right rows in the right place.

### The fix

Generate a new unique table key each time

### Automated testing - Infrastructure

Further fleshed out the idea of having components that mirror the cljs components. Now there's a `Component` to root them as we go forward. New components for TabBar and Collapse.

Also cleaned up how the various Workspace pages work. Hierarchy should be cleaner now.

### Automated testing - actual tests

* Test that the user can create workspace attributes
* Test that the user can delete a single attribute from a workspace already containing attributes, and that this works from any position
* Test that the user can create a workspace, add and delete attributes in a way that makes the table state look the same as it did once in the past, and that the right thing gets deleted from there.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Verify all tests go green, including CI tests and automated UI tests.
- [x] **Submitter**: Squash commits and merge to develop. If adding test code, merge application code and test code at the same time.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
@@ -137,13 +137,13 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures 
           val detailPage = listPage.openWorkspaceDetails(billingProject, workspaceName).awaitLoaded()
 
           detailPage.ui.beginEditing()
-          detailPage.ui.addWorkspaceAttribute("a", "a")
-          detailPage.ui.addWorkspaceAttribute("b", "b")
-          detailPage.ui.addWorkspaceAttribute("c", "c")
+          detailPage.ui.addWorkspaceAttribute("a", "X")
+          detailPage.ui.addWorkspaceAttribute("b", "Y")
+          detailPage.ui.addWorkspaceAttribute("c", "Z")
           detailPage.ui.save()
 
           // TODO: ensure sort, for now it's default sorted by key, ascending
-          detailPage.ui.readWorkspaceTable shouldBe List(List("a", "a"), List("b", "b"), List("c", "c"))
+          detailPage.ui.readWorkspaceTable shouldBe List(List("a", "X"), List("b", "Y"), List("c", "Z"))
         }
       }
 
@@ -194,16 +194,16 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures 
             val detailPage = listPage.openWorkspaceDetails(billingProject, workspaceName).awaitLoaded()
 
             detailPage.ui.beginEditing()
-            detailPage.ui.addWorkspaceAttribute("a", "a")
-            detailPage.ui.addWorkspaceAttribute("b", "b")
-            detailPage.ui.addWorkspaceAttribute("c", "c")
+            detailPage.ui.addWorkspaceAttribute("a", "W")
+            detailPage.ui.addWorkspaceAttribute("b", "X")
+            detailPage.ui.addWorkspaceAttribute("c", "Y")
             detailPage.ui.save()
 
             // TODO: better way to tell the page is ready
             Thread sleep 1000
 
             detailPage.ui.beginEditing()
-            detailPage.ui.addWorkspaceAttribute("d", "d")
+            detailPage.ui.addWorkspaceAttribute("d", "Z")
             detailPage.ui.save()
 
             Thread sleep 1000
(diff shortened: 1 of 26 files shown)


breilly2:
Ugh. Nice work tracking this down! I had tried to reproduce it and sometimes could but not reliably.

rushtong (approved):
👍 Works for me locally.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
```diff
@@ -128,6 +130,100 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures
...
+
+      "should be able to enter workspace attributes" in withWebDriver { implicit driver =>
+        withWorkspace(billingProject, "WorkspaceSpec_add_ws_attrs") { workspaceName =>
+          val listPage = signIn(Config.Users.harry)
+          val detailPage = listPage.openWorkspaceDetails(billingProject, workspaceName).awaitLoaded()
+
+          detailPage.ui.beginEditing()
+          detailPage.ui.addWorkspaceAttribute("a", "a")
+          detailPage.ui.addWorkspaceAttribute("b", "b")
+          detailPage.ui.addWorkspaceAttribute("c", "c")
```
davidangb: here and in other tests: I think you should use different strings for key and value. Otherwise, your test can't test that you haven't swapped key and value somewhere.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
```diff
@@ -128,6 +130,100 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures
...
+            val detailPage = listPage.openWorkspaceDetails(billingProject, workspaceName).awaitLoaded()
+
+            detailPage.ui.beginEditing()
+            detailPage.ui.addWorkspaceAttribute("a", "a")
+            detailPage.ui.addWorkspaceAttribute("b", "b")
+            detailPage.ui.addWorkspaceAttribute("c", "c")
+            detailPage.ui.save()
+
+            // TODO: better way to tell the page is ready
+            Thread sleep 1000
```
davidangb: agree with the TODO; this looks fragile, especially when we consider running this in different environments

src/cljs/main/broadfcui/page/workspace/summary/attribute_editor.cljs
```diff
@@ -102,11 +102,15 @@
              with-spaces {:error (str "Keys cannot have spaces: " (string/join ", " with-spaces))}
              invalid-numbers {:error (str "Invalid number for key(s): " (string/join ", " invalid-numbers))}
              :else {:success typed})))
+   :get-initial-state
+   (fn []
+     {:table-key (gensym)})
```
davidangb, line 107: I dunno if you need this in initial state, since you set it for any data changes ... but I *like* it here, for readability and understandability.

davidangb (commented):
a couple smallish comments inline. Love all the new capabilities in the tests.

I verified that the UI works for me manually; can't run the tests so can't comment there.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
breilly2: Yes, we should fix this. How does a person recognize that the page is ready?

davidangb (approved):
lgtm!

PR #980, opened 2017-10-03
GAWB-2548: refactored method fixtures, updated method config spec and publish spec

ahaessly:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
@@ -41,11 +41,11 @@ trait FireCloudClient {
       case true =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        Await.result(entityFuture, 100.millis).decodeString("UTF-8")
+        Await.result(entityFuture, 1.second).decodeString("UTF-8")
       case _ =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        throw APIException(Await.result(entityFuture, 100.millis).decodeString("UTF-8"))
+        throw APIException(Await.result(entityFuture, 1.second).decodeString("UTF-8"))
     }
   }
 
@@ -54,11 +54,11 @@ trait FireCloudClient {
       case true =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        Await.result(entityFuture, 100.millis).decodeString("UTF-8")
+        Await.result(entityFuture, 1.second).decodeString("UTF-8")
       case _ =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        throw APIException(Await.result(entityFuture, 100.millis).decodeString("UTF-8"))
+        throw APIException(Await.result(entityFuture, 1.second).decodeString("UTF-8"))
     }
   }
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
@@ -202,7 +202,7 @@ trait Orchestration extends FireCloudClient with LazyLogging {
     def createMethod(testname:String, method:Method, numSnapshots: Int = 1)
                     (implicit token: AuthToken): String = {
       val methodName = appendUnderscore(testname) + Util.makeUuid
-      for (i <- 1 to numSnapshots)
+      for (_ <- 1 to numSnapshots)
         createMethod(SimpleMethod.creationAttributes + ("name"->methodName))
       methodName
     }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
@@ -3,7 +3,11 @@ package org.broadinstitute.dsde.firecloud.api
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.api.WorkspaceAccessLevel.WorkspaceAccessLevel
 import org.broadinstitute.dsde.firecloud.config.{AuthToken, Config}
+import org.broadinstitute.dsde.firecloud.fixture.Method
+import org.broadinstitute.dsde.firecloud.fixture.MethodData.SimpleMethod
 import org.broadinstitute.dsde.firecloud.util.Retry.retry
+import org.broadinstitute.dsde.firecloud.util.Util
+import org.broadinstitute.dsde.firecloud.util.Util.appendUnderscore
 
 import scala.concurrent.duration._
 
@@ -158,18 +162,18 @@ trait Orchestration extends FireCloudClient with LazyLogging {
         Map("configurationNamespace" -> configurationNamespace, "configurationName" -> configurationName, "configurationSnapshotId" -> configurationSnapshotId, "destinationNamespace" -> destinationNamespace, "destinationName" -> destinationName))
     }
 
-    def createMethodConfigInWorkspace(wsNs: String, wsName: String, methodConfigVersion: Int,
-                                      methodNamespace: String, methodName: String, methodVersion: Int,
-                                      destinationNamespace: String, destinationName: String, inputs: Map[String, String], outputs: Map[String, String],
-                                      rootEntityType: String)(implicit token: AuthToken): String = {
-      logger.info(s"Creating method config: $wsNs/$wsName $methodConfigVersion method: $methodNamespace/$methodName destination: $destinationNamespace/$destinationName")
+    def createMethodConfigInWorkspace(wsNs: String, wsName: String, method:Method,
+                                      configNamespace: String, configName: String, methodConfigVersion: Int,
+                                      inputs: Map[String, String], outputs: Map[String, String],
+                                      rootEntityType: String)(implicit token: AuthToken): Unit = {
+      logger.info(s"Creating method config: $wsNs/$wsName $methodConfigVersion method: ${method.methodNamespace}/${method.methodName} config: $configNamespace/$configName")
       postRequest(apiUrl(s"api/workspaces/$wsNs/$wsName/methodconfigs"),
         Map("deleted" -> false,
           "inputs" -> inputs,
           "methodConfigVersion" -> methodConfigVersion,
-          "methodRepoMethod" -> Map("methodNamespace" -> methodNamespace, "methodName" -> methodName, "methodVersion" -> methodVersion),
-          "namespace" -> destinationNamespace,
-          "name" -> destinationName,
+          "methodRepoMethod" -> method.methodRepoInfo,
+          "namespace" -> configNamespace,
+          "name" -> configName,
           "outputs" -> outputs,
           "prerequisites" -> Map(),
           "rootEntityType" -> rootEntityType)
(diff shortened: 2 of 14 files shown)


automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -195,11 +199,23 @@ trait Orchestration extends FireCloudClient with LazyLogging {
   }
 
   object methods {
+    def createMethod(testname:String, method:Method, numSnapshots: Int = 1)
+                    (implicit token: AuthToken): String = {
+      val methodName = appendUnderscore(testname) + Util.makeUuid
+      for (i <- 1 to numSnapshots)
+        createMethod(SimpleMethod.creationAttributes + ("name"->methodName))
```
breilly2: Nit/style/for-future-reference: this can also be written as:
```
      for (_ <- 1 to numSnapshots)
        createMethod(SimpleMethod.creationAttributes + ("name"->methodName))
```
or
```
      (1 to numSnapshots).foreach { _ =>
        createMethod(SimpleMethod.creationAttributes + ("name" -> methodName))
      }
```
I don't know which of the 3 is better but I feel that `_` instead of `i` (and to some extent `foreach` instead of the implicit `map`) helps express that the expression only exists for its side effects.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/MethodData.scala
```diff
@@ -1,49 +1,65 @@
...
+  def methodRepoInfo = Map(
+    "methodNamespace" -> methodNamespace,
+    "methodName" -> methodName,
+    "methodVersion" -> snapshotId
+  )
+}
+
+/* METHODS */
 object MethodData {
+  val SimpleMethod = Method(
```
breilly2, line 29: This is great! I appreciate the readability gained by the use of named arguments instead of positional.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/MethodFixtures.scala
```diff
@@ -1,49 +1,31 @@
...
 import org.scalatest.Suite
 
 trait MethodFixtures extends CleanUp { self: WebBrowserSpec with Suite =>
 
 
-  def withConfigForRedactedMethodInWorkspace(testname:String, wsnamespace: String, wsname: String, withUnredactedSnapshots: Boolean)
-                                            (testCode: (String) => Any)
-                                            (implicit token: AuthToken): Unit = {
+  def withMethod(testName:String, method:Method, numSnapshots: Int = 1, cleanUp: Boolean = true)
+                (testCode: (String) => Any)
```
breilly2, line 12: Looking at usages, it might be helpful for `testCode` to be `(Method) => Any)`. This could eliminate the need for repeated occurrences of `MethodData.SimpleMethod.copy(methodName = methodName)`. I haven't actually tried this, though, so I don't know if there are negative consequences of the change.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -57,7 +57,7 @@ class WorkspaceMethodConfigDetailsPage(namespace: String, name: String, methodCo
     private val openLaunchAnalysisModalButtonQuery: Query = testId("open-launch-analysis-modal-button")
     private val openEditModeQuery: Query = testId("edit-method-config-button")
     private val editMethodConfigNameInputQuery: Query = testId("edit-method-config-name-input")
-    private val saveEditedMethodConfigButtonQuery: Query = testId("save-edited-method-config-button")
+    private val saveEdittedMethodConfigButtonQuery: Query = testId("save-editted-method-config-button")
```
breilly2: Nit: please re-fix this typo.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/WebBrowserUtil.scala
```diff
@@ -245,4 +245,15 @@ trait WebBrowserUtil extends WebBrowser {
...
   }
+
+  /**
+    * Creates a Query for an element with a class attribute.
+    *
+    * @param cssClass the expected class
+    * @param webDriver implicit WebDriver for the WebDriverWait
+    * @return a Query for the class
+    */
+  def cssClass(cssClass: String)(implicit webDriver: WebDriver): Query = {
```
breilly2: If we can, we should use Scalatests's `WebBrowser.className` instead of writing our own.

On further inspection, I guess `WebBrowser.className` might not work for matching multiple class names. Does this handle that better? If so, maybe this could be renamed `classNames` to better describe its similarities, and differences, to the built-in `className`.

Also, does the order of the class names in the argument matter? Does it require an exact match of all class names or will a subset work? Should this take a `List[String]` instead? Being in a shared utility class, I want to make sure the use case for this is clear.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -24,31 +24,31 @@ class MethodConfigSpec extends FreeSpec with WebBrowserSpec with CleanUp with Wo
...
       signIn(uiUser)
-      val methodConfigDetailsPage = new WorkspaceMethodConfigDetailsPage(billingProject, workspaceName, MethodData.SimpleMethodConfig.configNamespace, methodConfigName).open
+      val methodConfigDetailsPage = new WorkspaceMethodConfigDetailsPage(billingProject, workspaceName, SimpleMethodConfig.configNamespace, methodConfigName).open
 
-      val submissionDetailsPage = methodConfigDetailsPage.launchAnalysis(MethodData.SimpleMethodConfig.rootEntityType, TestData.SingleParticipant.entityId)
+      val submissionDetailsPage = methodConfigDetailsPage.launchAnalysis(SimpleMethodConfig.rootEntityType, TestData.SingleParticipant.entityId)
 
       submissionDetailsPage.waitUntilSubmissionCompletes() //This feels like the wrong way to do this?
-      assert(submissionDetailsPage.verifyWorkflowSucceeded())
+      submissionDetailsPage.verifyWorkflowSucceeded() shouldBe true
```
breilly2: If this is something like:
```
submissionDetailsPage.readWorkflowStatus() shouldBe SubmissionDetailsPage.SUCCESS_STATUS
```
then failure messages will be more meaningful and realize the full advantage of using should matchers instead of `assert`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -163,125 +163,150 @@ class MethodConfigSpec extends FreeSpec with WebBrowserSpec with CleanUp with Wo
...
+        SimpleMethodConfig.configName, SimpleMethodConfig.snapshotId, SimpleMethodConfig.configNamespace, methodConfigName)
 
       signIn(uiUser)
-      val methodConfigDetailsPage = new WorkspaceMethodConfigDetailsPage(billingProject, workspaceName, MethodData.SimpleMethodConfig.configNamespace, methodConfigName).open
+      val methodConfigDetailsPage = new WorkspaceMethodConfigDetailsPage(billingProject, workspaceName, SimpleMethodConfig.configNamespace, methodConfigName).open
       val workspaceMethodConfigPage = methodConfigDetailsPage.deleteMethodConfig()
 
       workspaceMethodConfigPage.filter(methodConfigName)
-      assert(find(methodConfigName).isEmpty)
+      find(methodConfigName).isEmpty shouldBe true
```
breilly2:
```
find(methodConfigName) shouldBe empty
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/FireCloudClient.scala
```diff
@@ -54,11 +54,11 @@ trait FireCloudClient {
       case true =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        Await.result(entityFuture, 100.millis).decodeString("UTF-8")
+        Await.result(entityFuture, 1.second).decodeString("UTF-8")
       case _ =>
         val byteStringSink: Sink[ByteString, Future[ByteString]] = Sink.fold(ByteString("")) { (z, i) => z.concat(i) }
         val entityFuture = response.entity.dataBytes.runWith(byteStringSink)
-        throw APIException(Await.result(entityFuture, 100.millis).decodeString("UTF-8"))
+        throw APIException(Await.result(entityFuture, 1.second).decodeString("UTF-8"))
```
davidangb, line 61: you could conceivably move these into a `val` somewhere to make it easy to change later ... but only 4 instances, possibly not worth it.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -163,125 +163,150 @@ class MethodConfigSpec extends FreeSpec with WebBrowserSpec with CleanUp with Wo
...
+            SimpleMethodConfig.configNamespace, configName, 1, SimpleMethodConfig.inputs, SimpleMethodConfig.outputs, "participant")
+
+          api.methods.redact(MethodData.SimpleMethod.copy(methodName = methodName))
+
           signIn(uiUser)
-          val methodConfigDetailsPage = new WorkspaceMethodConfigDetailsPage(billingProject, workspaceName, MethodData.SimpleMethodConfig.configNamespace, configName).open
-          assert(methodConfigDetailsPage.ui.isSnapshotRedacted())
+          val workspaceMethodConfigPage = new WorkspaceMethodConfigListPage(billingProject, workspaceName).open
+          workspaceMethodConfigPage.filter(configName)
+          find(cssClass(s"fa-exclamation-triangle fa ")).isDefined shouldBe true
```
breilly2: Specifics for locating elements (in this case by class names) should be in the page class. Also, it would be good for the page class return one of a possible set of states to check against instead of a boolean.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -195,11 +199,23 @@ trait Orchestration extends FireCloudClient with LazyLogging {
...
+      methodName
+    }
```
davidangb: I don't think this is used anywhere. Are you adding it for future use?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
davidangb: aaaaah the typo is killing me (but I realize we need it unless we change it everywhere)

breilly2 (changes requested):
Mostly small changes but ones worth making. I primarily reviewed overall structure and test infrastructure code. If there are other aspects of these changes that you want reviewed, such as changes to specific test cases, please let me know and I can take another look.

davidangb (approved):
👍 a few comments inline, nothing big. I am still getting set up on my new laptop and can't run these yet, so of course please make sure they're passing, etc!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
ahaessly: oh, so sorry. i will change it back to the right way: edited

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
ahaessly: it's used in "should be able to choose new method snapshot"

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/MethodFixtures.scala
ahaessly: it's actually easier to do the copy i think. that way the code generating the method doesn't have to do the copy and return a method.

breilly2: The code generating the method is here in one place so it would only have to be written once. There are currently multiple calls to `copy` that replace the method name, even multiple inside a single use of `withMethod`. Also, the fact that the test case's use of `SimpleMethod` (as opposed to some other method template) is repeated in the calls to `withMethod` and `copy`. It seems like a win to me.

PR #984, opened 2017-10-04
GAWB-2676: fix esc button issues with closing modal

ahaessly:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/table.cljs
+++ b/src/cljs/main/broadfcui/common/table.cljs
@@ -3,7 +3,7 @@
    [dmohs.react :as react]
    [clojure.set :as set]
    [broadfcui.common.components :as comps]
-   [broadfcui.components.text-filter :as text-filter]
+   [broadfcui.components.text-filter :refer [TextFilter]]
    [broadfcui.common.style :as style]
    [broadfcui.common.table.body :as body]
    [broadfcui.common.table.column-editor :refer [ColumnEditButton]]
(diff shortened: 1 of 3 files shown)


src/cljs/main/broadfcui/common/table.cljs
```diff
@@ -118,7 +119,7 @@
          (when (and (:filterable? behavior) (not (contains? external-query-params :filter-text)))
            (let [filter-bar-props (:filter-bar toolbar)]
              [:div {:style (:style filter-bar-props)}
-              [comps/TextFilter (merge
+              [text-filter/TextFilter (merge
```
MattPutnam: Since this is the only thing in this file, you can `:refer [TextFilter]` in the imports and just use `[TextFilter ...` here.

MattPutnam (approved):
👍

MattPutnam:
@davidangb @zarsky-broad The problem here is that [esc] does two things: It dismisses modals, and it clears (and submits) text fields. When focus was on a text field in a modal, both would fire, and the component unmount would win the race, leaving no ref for the filter to act on. The quick fix (which is what we did here) is to guard against empty refs and ignore. The "correct" fix is to make this behave right--text fields should .preventDefault() the [esc] event and clear/submit when there's text, and pass it through when there's not. But the entire rest of the application doesn't do that, and it might be a bunch of work.

davidangb (approved):
:+1: works for me!

PR #992, opened 2017-10-06
Automation overhaul (no JIRA)

MattPutnam:
When I first started writing automation tests, I found it annoying for a few reasons:

### Not enough abstraction of functional UI components

Pages were declaring testIds and operating on them at a low level. "Click on this test id, then wait for this test id to become visible, then type these characters into the text field at this id, then click on this id...". For some small pages this was okay, but large pages like the workspace summary page were too complicated.

I introduced the idea of `Component`s to encapsulate behavior of smallish UI widgets, which can be operated on via methods with sematic names. Some are tiny, but that's okay.

### Too easy to run into race conditions

The code, both the pages and the tests, were littered with clunky awaits, jammed in there whenever a problem had arisen.

There is now an architecture, which if followed, should prevent any unexpected races. `FireCloudView` now defines an abstract `awaitReady`, and `Component` provides a default implementation that simply waits for its underlying test id to become visible (but can be overridden). Pages must implement `awaitReady`, and should be able to delegate to any `Component` that does loading. `Page`'s `awaitReady` is called in `open`. If a page method creates a new page via `new`, it should also `await ready` that new page. If a page method causes the page to do some loading (e.g. a "save" button), it should do whatever it needs to do to await ready before returning. All of this means that if a test creates a page via `open`, it should never have to explicitly await, and all page methods should assume that they're ready when called.

**Takeaway:** train your brain to spot the following as an error:
```scala
def openThingy(): ThingyPage = {
  openThingyButton.doClick()
  new ThingyPage
}
```
Instead, these should look like:
```scala
def openThingy(): ThingyPage = {
  openThingyButton.doClick()
  await ready new ThingyPage
}
```
Basically, no naked `new`s.

Overall guidelines:

* Tests should instantiate their first view via `new WhateverPage(...).open`, and then strive to acquire subsequent pages through method return values. Tests should never need to explicitly await anything. Tests should interact with pages through semantically-named methods only.
* Pages should define their contents through Components, and strive to avoid raw test ids. Create new components as needed. See WorkspaceSummaryPage for some patterns to compose things. Pages should interact wtih Components through semantically-named methods. Any method that creates a new view/page should return that view/page and `await ready` it.
* Components are where the raw ids live. They should abstract away the guts of what's going on.

Remaining TODOs:

* Not everything strictly adheres to these guidelines. There are still a few places to fix.
* We need a better way for a Component to have sub-components. I'm having type issues when I try to use `Query`s from one context in another context.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off
- [x] **Submitter**: Verify all tests go green, including CI tests and automated UI tests.
- [x] **Submitter**: Squash commits and merge to develop. If adding test code, merge application code and test code at the same time.

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
@@ -5,14 +5,14 @@ import org.openqa.selenium.WebDriver
 import org.scalatest.selenium.WebBrowser
 
 /**
-  * Convenience parent class for FireCloud pages and other views such as modal
-  * dialogs.
+  * Parent class for all pages and components.
   */
 abstract class FireCloudView(implicit webDriver: WebDriver)
   extends WebBrowser with WebBrowserUtil {
+  def awaitReady(): Unit
 
   /** Query for the FireCloud Spinner component */
-  val spinner = testId("spinner")
+  val spinner: CssSelectorQuery = testId("spinner")
 
   def readText(q: Query): String = {
     find(q) map { _.text } getOrElse ""
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
@@ -1,4 +1,4 @@
-package org.broadinstitute.dsde.firecloud.page
+package org.broadinstitute.dsde.firecloud
 
 import org.broadinstitute.dsde.firecloud.test.WebBrowserUtil
 import org.openqa.selenium.WebDriver
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
@@ -11,9 +11,6 @@ abstract class FireCloudView(implicit webDriver: WebDriver)
   extends WebBrowser with WebBrowserUtil {
   def awaitReady(): Unit
 
-  /** Query for the FireCloud Spinner component */
-  val spinner: CssSelectorQuery = testId("spinner")
-
   def readText(q: Query): String = {
     find(q) map { _.text } getOrElse ""
   }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
@@ -1,8 +1,8 @@
-package org.broadinstitute.dsde.firecloud.page
+package org.broadinstitute.dsde.firecloud
 
 import org.openqa.selenium.WebDriver
 
-trait Stateful extends FireCloudView {
+trait Stateful { this: FireCloudView =>
   val element: Query
 
   def getState(implicit webDriver: WebDriver): String = {
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
@@ -12,4 +12,8 @@ trait Stateful { this: FireCloudView =>
   def stateOf(query: Query)(implicit webDriver: WebDriver): String = {
     query.element.attribute("data-test-state").get
   }
+
+  def awaitState(state: String)(implicit webDriver: WebDriver): Unit = {
+    await condition { getState == state }
+  }
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -1,7 +1,7 @@
 package org.broadinstitute.dsde.firecloud.component
 
 import org.broadinstitute.dsde.firecloud.FireCloudView
-import org.openqa.selenium.WebDriver
+import org.openqa.selenium.{JavascriptExecutor, WebDriver}
 
 abstract class Component(id: String, raw: Boolean = false)(implicit webDriver: WebDriver) extends FireCloudView {
   val element: CssSelectorQuery = if (raw) CssSelectorQuery(id) else testId(id)
@@ -15,4 +15,8 @@ abstract class Component(id: String, raw: Boolean = false)(implicit webDriver: W
   def isEnabled: Boolean = enabled(element)
 
   override def awaitReady(): Unit = awaitVisible()
+
+  def scrollToVisible(): Unit = {
+    webDriver.asInstanceOf[JavascriptExecutor].executeScript("arguments[0].scrollIntoView(true)", find(element).get.underlying)
+  }
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -8,7 +8,7 @@ abstract class Component(id: String)(implicit webDriver: WebDriver) extends Fire
 
   def findInner(id: String): CssSelectorQuery = testId(id) inside element
 
-  def awaitEnabled(): Unit = {
-    await enabled element
-  }
+  def awaitVisible(): Unit = await visible element
+
+  override def awaitReady(): Unit = awaitVisible()
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -12,6 +12,7 @@ abstract class Component(id: String, raw: Boolean = false)(implicit webDriver: W
   def awaitNotVisible(): Unit = await notVisible element
 
   def isVisible: Boolean = find(element).isDefined
+  def isEnabled: Boolean = enabled(element)
 
   override def awaitReady(): Unit = awaitVisible()
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -10,5 +10,7 @@ abstract class Component(id: String)(implicit webDriver: WebDriver) extends Fire
 
   def awaitVisible(): Unit = await visible element
 
+  def isVisible: Boolean = find(element).isDefined
+
   override def awaitReady(): Unit = awaitVisible()
 }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -3,12 +3,13 @@ package org.broadinstitute.dsde.firecloud.component
 import org.broadinstitute.dsde.firecloud.FireCloudView
 import org.openqa.selenium.WebDriver
 
-abstract class Component(id: String)(implicit webDriver: WebDriver) extends FireCloudView {
-  val element: CssSelectorQuery = testId(id)
+abstract class Component(id: String, raw: Boolean = false)(implicit webDriver: WebDriver) extends FireCloudView {
+  val element: CssSelectorQuery = if (raw) CssSelectorQuery(id) else testId(id)
 
   def findInner(id: String): CssSelectorQuery = testId(id) inside element
 
   def awaitVisible(): Unit = await visible element
+  def awaitNotVisible(): Unit = await notVisible element
 
   def isVisible: Boolean = find(element).isDefined
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
@@ -1,6 +1,6 @@
-package org.broadinstitute.dsde.firecloud.page.components
+package org.broadinstitute.dsde.firecloud.component
 
-import org.broadinstitute.dsde.firecloud.page.FireCloudView
+import org.broadinstitute.dsde.firecloud.FireCloudView
 import org.openqa.selenium.WebDriver
 
 abstract class Component(id: String)(implicit webDriver: WebDriver) extends FireCloudView {
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Table.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Table.scala
@@ -13,9 +13,9 @@ case class Table(private val id: String)(implicit webDriver: WebDriver)
 
   private def tab(name: String) = findInner(s"$name-tab")
 
-  private val prevPageButton = Button(findInner("prev-page").queryString)
-  private val nextPageButton = Button(findInner("next-page").queryString)
-  private def pageButton(page: Int) = Button(findInner(s"page-$page").queryString)
+  private val prevPageButton = Button("prev-page")
+  private val nextPageButton = Button("next-page")
+  private def pageButton(page: Int) = Button(s"page-$page")
   private val perPageSelector = findInner("per-page")
 
   override def awaitReady(): Unit = {
@@ -24,14 +24,14 @@ case class Table(private val id: String)(implicit webDriver: WebDriver)
   }
 
   def filter(text: String): Unit = {
-    awaitReady()
     searchField(filterField).value = text
     click on filterButton
+    awaitReady()
   }
 
   def goToTab(tabName: String): Unit = {
-    awaitReady()
     click on tab(tabName)
+    awaitReady()
   }
 
   def readDisplayedTabCount(tabName: String): Int = {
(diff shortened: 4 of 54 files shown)


MattPutnam:
jenkins retest

MattPutnam:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.dsde.firecloud.component
+
+import org.broadinstitute.dsde.firecloud.FireCloudView
+import org.openqa.selenium.{JavascriptExecutor, WebDriver}
+
+abstract class Component(id: String, raw: Boolean = false)(implicit webDriver: WebDriver) extends FireCloudView {
```
breilly2, line 6: The `raw` boolean feels special-casey to me. Could `Component` expect a `Query` (or `CssSelectorQuery` if needed) and let component implementations decide what to ask for and how to build the right query? I've played around with this a little and haven't yet found a syntax that works but would be willing to try harder.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
```diff
@@ -1,18 +1,18 @@
...
-  * dialogs.
+  * Parent class for all pages and components.
   */
 abstract class FireCloudView(implicit webDriver: WebDriver)
   extends WebBrowser with WebBrowserUtil {
+  def awaitReady(): Unit
 
   /** Query for the FireCloud Spinner component */
-  val spinner = testId("spinner")
+  val spinner: CssSelectorQuery = testId("spinner")
```
breilly2: This can probably be typed as a general `Query` instead of specifically a `CssSelectorQuery`.

Actually it looks like it's not even used anymore so it should probably just be removed.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -0,0 +1,198 @@
...
+import org.broadinstitute.dsde.firecloud.page.{PageUtil, _}
+import org.broadinstitute.dsde.firecloud.{FireCloudView, Stateful}
+import org.openqa.selenium.WebDriver
+import org.scalatest.selenium.Page
+
+/**
+  * Page class for the Workspace Detail page.
+  */
+class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver: WebDriver)
+  extends WorkspacePage(namespace, name) with Page with PageUtil[WorkspaceSummaryPage] with Stateful {
```
breilly2, line 17: Not sure how I feel about pages being `Stateful`. Seems to make sense for components. Pages can still contain components and do things like implement `awaitReady` in terms of its stateful components.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
```diff
@@ -1,8 +1,8 @@
-package org.broadinstitute.dsde.firecloud.page
+package org.broadinstitute.dsde.firecloud
 
 import org.openqa.selenium.WebDriver
 
-trait Stateful extends FireCloudView {
+trait Stateful { this: FireCloudView =>
```
breilly2, line 5: What would you think of making `Stateful`s be `Component`s? I think all current uses (except for `WorkspaceSummaryPage`, but see my comment below) are already `Component`s and that seems to me like the level that we would want to declare that a thing has state.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methodrepo/MethodRepoPage.scala
```diff
@@ -1,85 +1,56 @@
...
-    object MethodRepoTable extends Table("methods-table") {
-      private def methodLink(namespace: String, name: String) = findInner(s"method-link-$namespace-$name")
+  object MethodRepoTable extends Table("methods-table") {
+    private def methodLink(namespace: String, name: String) = Link(s"method-link-$namespace-$name")
 
-      def hasMethod(namespace: String, name: String): Boolean = {
-        find(methodLink(namespace, name)).isDefined
-      }
+    def hasMethod(namespace: String, name: String): Boolean = {
+      val link = methodLink(namespace, name)
```
breilly2: Looks like this line and the next can be removed.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
```diff
@@ -0,0 +1,22 @@
...
+  val element: CssSelectorQuery = if (raw) CssSelectorQuery(id) else testId(id)
+
+  def findInner(id: String): CssSelectorQuery = testId(id) inside element
+
+  def awaitVisible(): Unit = await visible element
```
breilly2, line 11: I'd be inclined to add this to `WebBrowserUtil`:
```
    def visible(component: Component, timeOutInSeconds: Long = defaultTimeOutInSeconds)
               (implicit webDriver: WebDriver): Unit = {
      withWaitForCondition(timeOutInSeconds) {
        find(component.element).exists(_.isDisplayed)
      }
    }
```
Same for `awaitNotVisible` below. Though I guess it depends on whether we want to increase or decrease use DSL syntax.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Table.scala
```diff
@@ -40,23 +40,23 @@ case class Table(private val id: String)(implicit webDriver: WebDriver)
   }
 
   def goToPreviousPage(): Unit = {
+    prevPageButton.doClick()
```
ahaessly: is there a reason to move away from the "click on x" to "x.doClick()"?
we still use "click on" in other places in this file

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
```diff
@@ -1,13 +1,13 @@
...
+import org.broadinstitute.dsde.firecloud.FireCloudView
 import org.openqa.selenium.WebDriver
 import org.scalatest.selenium.Page
-import org.scalatest.selenium.WebBrowser.go
 
 /**
   * Mix-in utilities for ScalaTest's Page.
   */
-trait PageUtil[P <: Page] { self: P =>
+trait PageUtil[P <: Page] extends FireCloudView { self: P =>
```
breilly2, line 10: `PageUtil` was intended to be not FireCloud-specific so I’d rather it not extend `FireCloudView`. I like the standard `awaitReady` but I’d like to make some tweaks to the code structure, maybe even just renaming things, to make a clean separation between FireCloud support and more general extensions to the Selenium API.

breilly2 (approved):
There's a lot here. While I looked at some of the tests themselves, I mostly looked at the infrastructure. Though there are some further changes I'd like to make, I like the direction that this is going. I don't see any problems that make it worth holding back these changes.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
```diff
@@ -29,10 +50,12 @@ case class ErrorModal(implicit webDriver: WebDriver) extends OKCancelModal {
     testId("error-modal").element != null
   }
 
-  def getErrorText(): String = {
+  def getErrorText: String = {
     readText(testId("message-modal-content"))
```
ahaessly, line 54: probably also a question for brian... do we want to remove the () when we have an empty parameter list? it seems we have left them in for most of the calls. but maybe it's more scala-y to remove them?

breilly2: Having the empty parens on a function that begins with "get" causes this warning in IntelliJ:

> Method with accessor-like name is empty-paren less... (⌘F1) 
>
> Methods that follow JavaBean naming contract for accessors are expected to have no side effects. The recommended convention is to use a parameterless method whenever there are no parameters and the method have no side effect. This convention supports the uniform access principle, which says that client code should not be affected by a decision to implement an attribute as a field or method. 
>
> * Refer to Programming in Scala, 10.3 Defining parameterless methods

I have left the empty parens for functions that wait for some state to change. While their only side effect is the passage of time, they observe a side effect (some UI state change). I could be persuaded otherwise but that's the convention I've been trying to follow.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/billing/BillingManagementPage.scala
```diff
@@ -54,145 +50,73 @@ class BillingManagementPage(implicit webDriver: WebDriver) extends Authenticated
...
+    val billingProjectLink = testId(projectName + "-link")
+    click on (await enabled billingProjectLink)
   }
 
 
-  def addUserToBillingProject(userEmail: String, role: String) = {
-    val modal = ui.openAddUserDialog()
+  def addUserToBillingProject(userEmail: String, role: String): Unit = {
+    addUserButton.doClick()
+    val modal = await ready new AddUserToBillingProjectModal
```
ahaessly, line 79: i'm a little confused as to why we don't use open here like we do for page objects

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methodrepo/MethodRepoPage.scala
```diff
@@ -1,85 +1,56 @@
...
+      link.isVisible
+      methodLink(namespace, name).isVisible
```
ahaessly: does isVisible really need to be called on both link and methodLink? if it's because of the side effects of the first.isVisible i think it should be named something else

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -1,274 +1,180 @@
...
 
-  override val url: String = s"${Config.FireCloud.baseUrl}#workspaces/$namespace/$name/method-configs/$methodConfigNamespace/$methodConfigName"
-
-  def launchAnalysis(rootEntityType: String, entityId: String, expression: String = "", enableCallCaching: Boolean = true) = {
-    val launchModal = ui.openLaunchAnalysisModal()
-    launchModal.launchAnalysis(rootEntityType, entityId, expression, enableCallCaching)
-    new SubmissionDetailsPage(namespace, name)
+  override def awaitReady(): Unit = {
+    await condition isLoaded
+    await spinner "Checking permissions..."
```
ahaessly, line 16: wondering why we call isLoaded instead of moving what it checks for to here...

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -1,274 +1,180 @@
...
+    await ready new ErrorModal
   }
 
-  def deleteMethodConfig(): WorkspaceMethodConfigListPage = {
-    ui.deleteMethodConfig()
-    new WorkspaceMethodConfigListPage(namespace, name)
+  def openEditMode(expectSuccess: Boolean = true): Unit = {
+    openEditModeButton.doClick()
+    if (expectSuccess)
+      saveEditedMethodConfigButton.awaitVisible()
```
ahaessly, line 52: i like this!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Table.scala
MattPutnam: I'm having issues getting inner components working. So for right now, we have to use raw `Query`s for nested components. I'll make this more consistent.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/Stateful.scala
MattPutnam: I think pages can have state too. I think it's the cleanest way to handle editable pages.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
MattPutnam: Yeah, I've been having lots of issues with passing `Query`s as parameters to other classes. I'll get something along the lines of: `expected: Button.this.Query actual: Component.this.Query`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methodrepo/MethodRepoPage.scala
MattPutnam: Yep, probably a bad merge.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
MattPutnam: But the summary page really is stateful. We could wrap the entire page content in a Stateful component, but that seems like a waste.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/billing/BillingManagementPage.scala
MattPutnam: Modals aren't Pages and don't have the `open` method. We could change that.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/methodrepo/MethodRepoPage.scala
MattPutnam: Nope, it was a bad merge

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -1,274 +1,180 @@
...
-      methodConfigNameElement.get.text == methodConfigName
+  def changeSnapshotId(newSnapshotId: Int): Unit = {
+    editMethodConfigSnapshotIdSelect.select(newSnapshotId.toString)
+  }
 
+  private def changeInputsOutputs(fields: Map[String, String]): Unit = {
+    for ((field, expression) <- fields) {
+      val fieldInputQuery: Query = xpath(s"//*[@data-test-id='$field-text-input']/..//input")
+      searchField(fieldInputQuery).value = expression
     }
```
ahaessly, line 89: i got confused as to how this works for output fields. then i realized that input here is refering to the text input field. not sure if this would benefit from some comments about that

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -1,274 +1,180 @@
...
+    isErrorTextPresent(errorText)
   }
 
   def verifyMissingInputsError(errorText: String): Boolean = {
-    ui.isErrorTextPresent(errorText)
-  }
-
-  def closeModal() = {
-    ui.closeModal()
+    isErrorTextPresent(errorText)
```
ahaessly: can these two be merged into one call verifyErrorText? we can do this later though

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
MattPutnam: I like having empty parens for functions with side effects, and I would suggest naming them in such a way as to avoid that warning. "Get" really does imply a lack of side effects.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
MattPutnam: isLoaded is used in a few places. I don't necessarily like that it is, but I didn't want to change unit tests for this (besides cleaning up how they wait for things to be ready).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -53,24 +59,21 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
           withWorkspace(projectName, "AuthDomainSpec_share", Set(authDomainName), List(AclEntry(Config.Users.george.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
             withCleanUp {
               withSignIn(Config.Users.george) { listPage =>
-                val summaryPage = listPage.openWorkspaceDetails(projectName, workspaceName)
+                val summaryPage = listPage.enterWorkspace(projectName, workspaceName)
```
ahaessly, line 62: i like this pattern, but it's not used everywhere (e.g. PublishSpec). we should remember to update later. and definitely add to the docs.

ahaessly (approved):
some questions and comments but looks good overall

PR #1008, opened 2017-10-17
Jr user pool

jacmrob:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/UserPool.scala
```diff
@@ -0,0 +1,58 @@
...
+    * Users in Owners, Curators, AuthDomainUsers, and Students
+    */
+  def chooseAnyUser(n: Int = 1): Seq[Credentials] = {
+    allUsers.getRandomCredentials(n)
+  }
+
+  /**
+    * Chooses an admin user.
+    */
+  def chooseAdmin(n: Int = 1): Seq[Credentials] = {
```
breilly2: To prevent callers needing just 1 user (which is most cases), I think it makes sense to have 2 forms for each of these functions, e.g.:
```
  /** Chooses an admin user. */
  def chooseAdmin: Credentials = chooseAdmins(1).head

  /** Chooses n admin users. */
  def chooseAdmins(n: Int): Seq[Credentials] = Admins.getRandomCredentials(n)
```
I pluralized the existing function (not actually not needed because function overloading works, but I think it reads better) and removed the default parameter value to encourage calling `chooseAdmin` instead of `chooseAdmins().head`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/AuthToken.scala
```diff
@@ -6,25 +6,8 @@ import com.google.api.client.json.jackson2.JacksonFactory
...
-  val fred = AuthToken(Config.Users.fred)
-  val elvin = fred
-  val george = AuthToken(Config.Users.george)
-  val bill = AuthToken(Config.Users.bill)
-  val lunaTemp = AuthToken(Config.Users.lunaTemp)
-  val nevilleTemp = AuthToken(Config.Users.nevilleTemp)
-  val draco = AuthToken(Config.Users.draco)
-}
+// todo: maybe make all auth tokens and map them to their credentials?  so when we grab random creds we grab pre-populated auth token?
+// is it expensive to make auth tokens?
```
breilly2: I'm not sure how expensive it is. Be aware, though, that we will be getting a new auth token each time we do `AuthToken(user)` whereas we would previously get the same auth token when referencing the ones in the `AuthTokens` object. I don't _think_ this will be a problem but wanted to raise awareness.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
```diff
@@ -25,38 +48,35 @@ object Config {
...
+    val Owners = UserSet(makeCredsMap(userDataJson("owners")))
+    val Curators = UserSet(makeCredsMap(userDataJson("curators")))
+    val Temps = UserSet(makeCredsMap(userDataJson("temps")))
+    val AuthDomainUsers = UserSet(makeCredsMap(userDataJson("authdomains")))
+    val Students = UserSet(makeCredsMap(userDataJson("students")))
 
-    val harry = Credentials(users.getString("harry"), notSoSecretPassword)
-    val ron = Credentials(users.getString("ron"), notSoSecretPassword)
-    val draco = Credentials(users.getString("draco"), notSoSecretPassword)
+    // defaults
```
breilly2, line 65: Do we ever want to reference static users? It looks like `owner` and `temp` are the only one ever referenced and could probably be replaced with things like `UserPool.chooseProjectOwner`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -29,7 +29,8 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
   val projectName: String = Config.Projects.common
 
   // Unless otherwise declared, this auth token will be used for API calls.
-  implicit val authToken: AuthToken = AuthTokens.fred
+  val defaultUser: Credentials = UserPool.chooseCurator().head
```
breilly2: Should add a comment that the choice of curator here is only to avoid the chance of choosing the same user as the tests (which use auth domain users), not because we specifically need to use a curator.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/SignInSpec.scala
```diff
@@ -13,14 +12,17 @@ class SignInSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Cle
     "with a registered account" - {
 
       "should be able to log in and out multiple times as multiple users" in withWebDriver { implicit driver =>
-        withSignIn(Config.Users.draco) { listPage =>
-          listPage.readUserEmail() shouldEqual Config.Users.draco.email
+        val users = UserPool.chooseStudent(2)
+        val user1 = users.head
+        val user2 = users(1)
```
breilly2: I think we should be able to make the assignment to `user1` and `user2` a little cleaner but I couldn't get it to work in the 10 minutes I spent on it. Just mentioning it here but let's save it for later clean-up.

breilly2 (approved):
👍  Any of my suggestions were minor clean-up. After that (or if there's good reason to not make changes), looks good!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
jacmrob: there were a few places where i opted to use hermione vs a user-pool chosen user, since she has some additional configuration around her regarding some of the methods/method configs we build-in during fiab populate.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/UserPool.scala
jacmrob: Nice

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
breilly2: If that's fixed now in the user setup, can we avoid choosing a specific user? I worry that making these available will tempt others to use them when they should be using the pool.

breilly2: And if we really need to use specific users for specific reasons, let's rename the val to be explicit about who we're using and add some comments here about why the users are special (and remove any that we don't need).

PR #1016, opened 2017-10-24
Iz make testing easier

zarsky-broad:
In short, when testing locally against a fiab, if it's in your `/etc/hosts`, you can just run 
```bash
./run-tests.sh fiab
```


zarsky-broad:
jenkins: retest

README.md
```diff
@@ -8,39 +8,48 @@ https://portal.firecloud.org
...
 
-This can take around 20 seconds to completely start. When ready, it will display the following message:
+This can take around 40 seconds to completely start. When ready, it will display the following message:
 
 ```
 Prompt will show when figwheel connects to your application
 ```
 
-To connect, reload the browser window. The prompt should appear less than ten seconds after you reload the page. If it is not connecting, make sure to check the JavaScript console for error messages.
+To connect, load [`http://local.broadinstitute.org`](http://local.broadinstiute.org) in your browser (Chrome is strongly recommended for development). The prompt should appear less than ten seconds after you reload the page. If it is not connecting, make sure to check the JavaScript console for error messages.
```
jmthibault79: typo: institute

zarsky-broad:
jenkins: retest

automation/docker/run-tests.sh
```diff
@@ -1,73 +1,79 @@
...
-TEST_CONTAINER=${4:-automation}
-VAULT_TOKEN=$5
-WORKING_DIR=${6:-$PWD}
-export WORKING_DIR=$WORKING_DIR
-
-if [ -z $VAULT_TOKEN ]; then
-    VAULT_TOKEN=$(cat ~/.vault-token)
+
+if [ "$FC_INSTANCE" = "local" ]; then
+  FC_INSTANCE="172.19.0.1"
```
ahaessly: where did this ip address come from?

automation/docker/run-tests.sh
```diff
@@ -1,73 +1,79 @@
...
 fi
 
-if [ "$DOCKERHOST" = "local" ]
-  then
-    docker pull chickenmaru/nettools
-    export DOCKERHOST=`docker run --net=docker_default -it --rm chickenmaru/nettools sh -c "ip route|grep default|cut -d' ' -f 3"`
-    echo "Docker Host from Docker Perspective: $DOCKERHOST"
+if [ "$FC_INSTANCE" = "fiab" ]; then
+  FC_INSTANCE="$(grep 'firecloud-fiab.dsde-dev.broadinstitute.org' /etc/hosts | awk '{print $1}')"
```
ahaessly: sometimes i have multiple entries in my /etc/hosts but comment out the one i'm not using. i'm guessing this won't skip the commented out lines

README.md
```diff
@@ -75,101 +84,81 @@ To compile and build the `broadinstitute/firecloud-ui` docker image, run
...
+
+* Working directory
+	* Defaults to `$PWD`.
+* Vault auth token
+	* Defaults to reading it from `~/.vault-token`.
+* `dev` or `qa`
+	* Environment of your FiaB. Defaults to `dev`.
+* Local UI
+	* Enter `local_ui` here to run against a local UI stack.
+
```
ahaessly: add a comment that: if you specify this param, you must specify all params

ahaessly (approved):
a few comments

automation/docker/run-tests.sh
zarsky-broad: It's reliably the address for the real localhost inside docker networks. See the firecloud-ui rsync script.

automation/docker/run-tests.sh
zarsky-broad: Hmm, yeah. In that case, you'd have to just put in the IP address as the argument.

README.md
zarsky-broad: Ah, I'd thought that since they were positional, that was assumed.

automation/docker/run-tests.sh
dmohs: I would document this something like as follows:
```
HOST_ADDRESS="172.19.0.1" # maybe link to doc here
...
if ...
  FC_INSTANCE="$HOST_ADDRESS"
```

automation/docker/run-tests.sh
dmohs: Option:
```
FC_INSTANCE="$(cat /etc/hosts | grep -v '#' | grep 'firecloud-fiab.dsde-dev.broadinstitute.org' | awk '{print $1}')"
```

dmohs (approved):
LGTM but these scripts are screaming for a change to [not bash].

zarsky-broad:
jenkins: retest

automation/Dockerfile-tests
```diff
@@ -1,4 +1,4 @@
-FROM hseeberger/scala-sbt
+FROM digitalgenius/alpine-scala-sbt
```
breilly2: I think I'd looked at this one and was uncomfortable with not being able to see the Dockerfile. I'd found `1science/sbt` which is more transparent... and at 71 MB is even smaller than digitalgenius/alpine-scala-sbt's 379 MB.

README.md
```diff
@@ -73,140 +70,11 @@ To compile and build the `broadinstitute/firecloud-ui` docker image, run
...
-```
+## Testing
+See [TESTING.md](TESTING.md)
 
 ## Troubleshooting
 
-If you have problems with IntelliJ, it may be due to artifacts left over from a previous import or
-build. If you have problems, first close the projects in IntelliJ and delete the `automation/.idea`
-and `automation/target` directories. Then repeat the IntelliJ project import instructions above.
+If you have problems with IntelliJ, it may be due to artifacts left over from a previous import or build. If you have problems, first close the projects in IntelliJ and delete the `automation/.idea` and `automation/target` directories. Then repeat the IntelliJ project import instructions above.
```
breilly2: Since this is related to the project in `automation`, this should probably be moved or copied into TESTING.md

TESTING.md
```diff
@@ -0,0 +1,132 @@
...
+```
+
+For more information see [SBT's documentation](http://www.scala-sbt.org/0.13/docs/Testing.html#Test+Framework+Arguments).
+
+##### Against a local UI
+
+Be sure you used the `local_ui` param when you rendered your configs (see above). When starting your UI, run:
+
+```bash
+FIAB=true [ENV=qa] ./config/docker-rsync-local-ui.sh
```
breilly2: We should probably not even publicize running with `ENV=qa` because only Jenkins should be doing that.

This applies throughout. I just didn't realize it until seeing it here.

If we want to keep the documentation here for the sake of completeness, we should add an emphatic statement about not using `qa` locally (or any network other than where the Jenkins nodes are).

automation/docker/run-tests.sh
```diff
@@ -1,73 +1,85 @@
...
-export DOCKERHOST=$DOCKERHOST
-TEST_CONTAINER=${4:-automation}
-VAULT_TOKEN=$5
-WORKING_DIR=${6:-$PWD}
-export WORKING_DIR=$WORKING_DIR
-
-if [ -z $VAULT_TOKEN ]; then
-    VAULT_TOKEN=$(cat ~/.vault-token)
+
+if [ "$FC_INSTANCE" = "local" ]; then
```
breilly2: Does this allow us to run headless against local.broadinstitue.org? I didn't know that would be possible so if that's the case that's awesome!

zarsky-broad:
jenkins: retest

zarsky-broad:
jenkins: retest

breilly2 (commented):
Just some comments... not a full review. If others are happy with it then so am I. Thanks for contributing the improvements!

ahaessly (approved):
👍

automation/docker/run-tests.sh
zarsky-broad: Yup! Runs headless against local UI, backed by a FiaB.

MattPutnam (approved):
👍

davidangb (approved):
you have a bunch of thumbs already. I looked at a pretty high level, did not try it myself. Looks good from here.

PR #1029, opened 2017-11-02
Background sign in (no JIRA)

breilly2:
Goodbye captchas!

Thank you Mohs for walking me though how to do this!

There's a new JS function on window that takes an auth token, verifies it with a quick web service call to Google, and creates a doppelganger auth2 object to stand in for the one usually created by Google's OAuth2 client library. FireCloud API calls, workflows, sign out, and subject ID based local storage keys all work.

To make sure everything was nicely normalized, I removed the `signIn` functions from `WebBrowserSpec` in favor of expanding out the `withSignIn` functionality of `UserFixtures`. Both real sign in (registered and non-registered users) and background sign in are supported.

The only caveat I found was that the signed-in state is lost if the browser is reloaded. Some tests caused this by creating a page object and calling `open` on it. Interestingly, as long as the post-# part of the URL changes, the entire page does not reload (I'm guessing it's our navigation handling code). However, if you open a page and the post-# part of the URL happens to be the same, the page is reloaded.

I added a class to the hierarchy under AuthorizedPage (and applied to all existing pages except RegistrationPage) that adds functions for the top navigation and changed calls to `open` to instead activate the top navigation links or inner tabs to avoid page reloads.

I also bumped parallelism up to 3 chrome nodes without issue. Successful test run https://fc-jenkins.dsp-techops.broadinstitute.org/view/Test/job/fiab-test/1183/ took 15 minutes. For another data point, the previous run with 1 failure used 4 nodes and took 13 minutes.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/RegistrationSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/RegistrationSpec.scala
@@ -3,21 +3,20 @@ package org.broadinstitute.dsde.firecloud.test.user
 import com.typesafe.scalalogging.LazyLogging
 import org.broadinstitute.dsde.firecloud.api.{Sam, Thurloe}
 import org.broadinstitute.dsde.firecloud.config.{AuthToken, Config, Credentials, UserPool}
+import org.broadinstitute.dsde.firecloud.fixture.UserFixtures
 import org.broadinstitute.dsde.firecloud.page.library.DataLibraryPage
 import org.broadinstitute.dsde.firecloud.page.workspaces.WorkspaceListPage
-import org.broadinstitute.dsde.firecloud.page.user.RegistrationPage
 import org.broadinstitute.dsde.firecloud.test.{CleanUp, WebBrowserSpec}
 import org.scalatest.{BeforeAndAfter, FreeSpec, Matchers}
-import org.broadinstitute.dsde.firecloud.test.Tags
 
 
 /**
   * Tests for new user registration scenarios.
   */
-class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with WebBrowserSpec with CleanUp with LazyLogging {
+class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with WebBrowserSpec
+  with UserFixtures with CleanUp with LazyLogging {
 
-  val email: String = Config.Users.temp.email
-  val password: String = Config.Users.temp.password
+  val testUser: Credentials = Config.Users.temp  // TODO: pull from user pool and fetch correct subject ID
   val subjectId: String = Config.Users.tempSubjectId
 
   val adminUser: Credentials = UserPool.chooseAdmin
@@ -41,9 +40,30 @@ class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with W
 
     "should allow a person to register" in withWebDriver { implicit driver =>
 
-      signIn(email, password)
-      val registrationPage = await ready new RegistrationPage
+      withSignInNewUserReal(testUser) { registrationPage =>
+        registerCleanUpForDeleteUser(subjectId)
 
+        registrationPage.register(
+          firstName = "Test",
+          lastName = "Dummy",
+          title = "Tester",
+          contactEmail = Some("test@firecloud.org"),
+          institute = "Broad",
+          institutionalProgram = "DSDE",
+          nonProfitStatus = true,
+          principalInvestigator = "Nobody",
+          city = "Cambridge",
+          state = "MA",
+          country = "USA")
+
+        new DataLibraryPage().validateLocation()
+      }
+    }
+  }
+
+  "should show billing account instructions for a newly registered user" in withWebDriver { implicit driver =>
+
+    withSignInNewUserReal(testUser) { registrationPage =>
       registerCleanUpForDeleteUser(subjectId)
 
       registrationPage.register(
@@ -59,36 +79,13 @@ class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with W
         state = "MA",
         country = "USA")
 
-      new DataLibraryPage().validateLocation()
+      await ready new DataLibraryPage()
+
+      val listPage = new WorkspaceListPage().open
+      listPage.clickCreateWorkspaceButton(true)
+
+      listPage.showsNoBillingProjectsModal() shouldBe true
     }
-  }
-
-  "should show billing account instructions for a newly registered user" in withWebDriver { implicit driver =>
-
-    signIn(email, password)
-    val registrationPage = await ready new RegistrationPage
-
-    registerCleanUpForDeleteUser(subjectId)
-
-    registrationPage.register(
-      firstName = "Test",
-      lastName = "Dummy",
-      title = "Tester",
-      contactEmail = Some("test@firecloud.org"),
-      institute = "Broad",
-      institutionalProgram = "DSDE",
-      nonProfitStatus = true,
-      principalInvestigator = "Nobody",
-      city = "Cambridge",
-      state = "MA",
-      country = "USA")
-
-    await ready new DataLibraryPage()
-
-    val listPage = new WorkspaceListPage().open
-    listPage.clickCreateWorkspaceButton(true)
-
-    listPage.showsNoBillingProjectsModal() shouldBe true
   }
 
 /*
--- a/src/cljs/main/broadfcui/auth.cljs
+++ b/src/cljs/main/broadfcui/auth.cljs
@@ -159,6 +159,34 @@
                                        :redirectUri (.. js/window -location -origin)})
                                :on-done #(swap! state assoc :hidden? true)})))))})
 
+(defn force-signed-in [{:keys [on-sign-in on-sign-out on-error]}]
+  (fn [auth-token]
+    (utils/ajax {:url (str "https://www.googleapis.com/oauth2/v3/tokeninfo?access_token=" auth-token)
+                 :on-done
+                 (fn [{:keys [status-code success? get-parsed-response raw-response]}]
+                   (if success?
+                     (let [{:keys [email sub]} (get-parsed-response)
+                           auth2 (clj->js
+                                  {:currentUser
+                                   {:get
+                                    (constantly
+                                     (clj->js
+                                      {:getAuthResponse
+                                       (constantly (clj->js {:access_token auth-token}))
+                                       :getBasicProfile
+                                       (constantly (clj->js {:getEmail (constantly email)
+                                                             :getId (constantly sub)}))}))
+                                    :listen (fn [])}
+                                   :signOut on-sign-out})]
+                       (utils/set-google-auth2-instance! auth2)
+                       (on-sign-in))
+                     (on-error {:status status-code :response raw-response})))})))
+
+(defn render-forced-sign-in-error [error]
+  [:div {}
+   [:div {} "Status: " (:status error)]
+   [:div {} "Response: " (:response error)]])
+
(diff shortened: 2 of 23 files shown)


src/cljs/main/broadfcui/main.cljs
```diff
@@ -175,9 +175,32 @@
    (fn []
      {:user-status #{}})
    :component-will-mount
-   (fn [{:keys [this]}]
+   (fn [{:keys [this state]}]
      (init-nav-paths)
-     (this :handle-hash-change))
+     (this :handle-hash-change)
+     (set! js/window.forceSignedIn
```
dmohs: `js/window.forceSignedIn` -> `(.-forceSignedIn js/window)` to be a bit more idiomatic.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -175,9 +175,32 @@
...
+           (fn [auth-token]
```
dmohs: Might be nice to push this into a separate file to keep concerns separated.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -175,9 +175,32 @@
...
+                                                            :getBasicProfile
+                                                            (fn [] (clj->js
+                                                                    {:getEmail (fn [] email)
+                                                                     :getId (fn [] sub)}))}))
+                                                :listen (fn [])}
+                                               :signOut
+                                               (fn [] (swap! state update :user-status disj :signed-in))})]
+                                (utils/set-google-auth2-instance! instance)
+                                (swap! state update :user-status conj :signed-in))
+                              (utils/log "Failed to validate token: " status-code)))}))))
```
dmohs: Do you check for log messages like this? I'm worried this will never be seen.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -175,9 +175,32 @@
...
+                                                (fn []
+                                                  (clj->js {:getAuthResponse
+                                                            (fn [] (clj->js {:access_token auth-token}))
+                                                            :getBasicProfile
+                                                            (fn [] (clj->js
+                                                                    {:getEmail (fn [] email)
+                                                                     :getId (fn [] sub)}))}))
+                                                :listen (fn [])}
+                                               :signOut
+                                               (fn [] (swap! state update :user-status disj :signed-in))})]
```
dmohs: `(fn [] (swap! state update :user-status disj :signed-in))` can be `#(swap! state update :user-status disj :signed-in)`

src/cljs/main/broadfcui/main.cljs
```diff
@@ -175,9 +175,32 @@
...
+                          :on-done
+                          (fn [{:keys [status-code success? get-parsed-response]}]
+                            (if success?
+                              (let [{:keys [email sub]} (get-parsed-response)
+                                    instance (clj->js
+                                              {:currentUser
+                                               {:get
+                                                (fn []
+                                                  (clj->js {:getAuthResponse
+                                                            (fn [] (clj->js {:access_token auth-token}))
```
dmohs: `(fn [] (clj->js {:access_token auth-token}))` -> `(constantly (clj->js {:access_token auth-token}))`

dmohs (approved):

src/cljs/main/broadfcui/main.cljs
breilly2: I will make this change if this is more idiomatic. However, while I guess I like the intent declared in words, I am confused by the preference for words over shorter symbolic idioms that seem more typical in our codebase. The one difference I can think of is that it might specify that I don't care what the function arguments are, however I could get that same effect with something like
```
(fn [& _] (clj->js {:access_token auth-token}))
```
or even simply
```
#(clj->js {:access_token auth-token})
```
Though if that's the only difference, I might even have a slight preference for how it's written because this function is mimicking an existing API where the function does not expect arguments. Thoughts?

src/cljs/main/broadfcui/main.cljs
breilly2: I was going to say that I don't really care much about raising visibility because I'd expect a developer to be in the console while debugging a problem like this anyway, but that's not the case if this happens sporadically during an automated test run. I could render an error modal that would be captured by a test failure screenshot if that sounds good to you.

src/cljs/main/broadfcui/main.cljs
MattPutnam: In general, the codebase itself is not a very good indicator of what's idiomatic or best practice. We're getting better, but only Mohs was an expert on this tech stack when we started, and even his opinions have evolved.

My (current) opinion is that if a function takes an argument but we don't care about it (for example, `onClick`s and other things that receive an event object), then acknowledge that via `(fn [_] ... )`. I also like `constantly` for no-arg functions, but that's kind of a new thing so it's not in much of the codebase yet. `constantly` is nice because it avoids a `#` macro, which is only bad because they can't be nested.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/RegistrationSpec.scala
```diff
@@ -59,36 +79,13 @@ class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with W
         state = "MA",
         country = "USA")
 
-      new DataLibraryPage().validateLocation()
-    }
-  }
+      await ready new DataLibraryPage()
```
ahaessly, line 82: it this correct that it is waiting for the data library page? or should it be workspace list page?

breilly2: It is correct. For a new user when they first register, they are taken to the data library page. On subsequent sign ins they are taken to the workspace list page. Isaac implemented this based on a PO request a number of months ago so ask him if you're curious and want to know more.

ahaessly (approved):
looks good - just one question making sure data library is the page loaded after registration.

src/cljs/main/broadfcui/main.cljs
dmohs: Exactly what Matt said. I like `constantly` lately because it communicates clearly that arguments are not used and not intended to be used to construct the return value.

src/cljs/main/broadfcui/main.cljs
dmohs: I use the console extensively for debugging, so keeping noise out is essential. That's my main concern. Use your own judgement.

src/cljs/main/broadfcui/main.cljs
breilly2: Thanks Putnam. That's very helpful for my learning! Sounds like `constantly` is the way to go here as Mohs originally suggested.

breilly2:
I moved the forced sign-in function into auth.cljs because it seemed to belong there. I had a challenge with rendering an error response in a modal, keeping the state of the modal display in main.cljs app while keeping knowledge about the shape of the error in auth.cljs. It might be a little sloppy so suggestions are welcome. This has been a great learning experience for me in react/clojurescript!

src/cljs/main/broadfcui/auth.cljs
```diff
@@ -159,6 +159,34 @@
                                        :redirectUri (.. js/window -location -origin)})
                                :on-done #(swap! state assoc :hidden? true)})))))})
 
+(defn force-signed-in [{:keys [on-sign-in on-sign-out on-error]}]
+  (fn [auth-token]
+    (utils/ajax {:url (str "https://www.googleapis.com/oauth2/v3/tokeninfo?access_token=" auth-token)
```
dmohs: Do you think there is any need for `(js/encodeURIComponent auth-token)` here?

src/cljs/main/broadfcui/auth.cljs
```diff
@@ -159,6 +159,34 @@
...
+                                  {:currentUser
+                                   {:get
+                                    (constantly
+                                     (clj->js
+                                      {:getAuthResponse
+                                       (constantly (clj->js {:access_token auth-token}))
+                                       :getBasicProfile
+                                       (constantly (clj->js {:getEmail (constantly email)
+                                                             :getId (constantly sub)}))}))
+                                    :listen (fn [])}
```
dmohs: Maybe, `(constantly nil)`, although perhaps you are trying to communicate that the return value is not used.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -178,29 +179,10 @@
...
-                                                :listen (fn [])}
-                                               :signOut
-                                               (fn [] (swap! state update :user-status disj :signed-in))})]
-                                (utils/set-google-auth2-instance! instance)
-                                (swap! state update :user-status conj :signed-in))
-                              (utils/log "Failed to validate token: " status-code)))}))))
+     (set! (.-forceSignedIn js/window)
+           (auth/force-signed-in {:on-sign-in #(swap! state update :user-status conj :signed-in)
+                                  :on-sign-out #(swap! state update :user-status disj :signed-in)
+                                  :on-error (fn [error] (swap! state assoc :force-sign-in-error error))})))
```
dmohs: `#(swap! state assoc :force-sign-in-error %)` would be more concise, but possibly less clear. Your call.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -209,6 +191,11 @@
                                public?
                                (contains? (:user-status @state) :signed-in))]
        [:div {}
+        (let [error (:force-sign-in-error @state)]
+          (when error
+            (modals/render-error {:header (str "Error validating access token")
```
dmohs: Could be combined as: `(when-let [error (:force-sign-in-error @state)] ...)`

dmohs (approved):
I love reading your code in any language. Your careful attention to detail and readability is exemplary.

src/cljs/main/broadfcui/auth.cljs
breilly2: Good suggestion. I assume it shouldn't hurt for valid input and anything we can do to sanitize input is a good thing. I'll try it to make sure it doesn't have any bad side effects.

src/cljs/main/broadfcui/auth.cljs
breilly2: Not intentional. I wan't sure what was the best way to express a no-op function but if you're happy with `(constantly nil)` then so am I.

src/cljs/main/broadfcui/main.cljs
breilly2: Maybe only less clear to learners like me. I've been bothered by the style being inconsistent with `:on-sign-in` and `:on-sign-out` so I like this change.

src/cljs/main/broadfcui/main.cljs
breilly2: More learning. Thanks!

src/cljs/main/broadfcui/auth.cljs
MattPutnam: Hmm... I feel like there should be a better way to show this intention. We don't have many (any?) other places where there's a no-op, because the best way to do it would be to design the library to gracefully handle not being given a function. So we don't have an established pattern.

To me, `(constantly nil)` suggests that the return value is being used, and it's always `nil`. `(fn [])` reads better as a no-op, but also kinda looks like a mistake. Maybe we should define `(def no-op (fn []))` in utils so this can be `:listen utils/no-op`.

Not a very strong opinion though.

dmohs: I like this suggestion. Perhaps `(defn no-op [])`.

PR #1074, opened 2017-11-21
GAWB-2431: upgrade logback version in automation

davidangb:
upgrade logback to latest version to avoid warnings from our security scanner.


ahaessly (approved):
👍

PR #1091, opened 2017-12-08
Adding Library test to check on TCGA message modal and complimentary framework

gdlex4015:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
@@ -245,7 +245,7 @@ trait Orchestration extends FireCloudClient with LazyLogging {
   */
    object NIH {
     def addUserToNIH(jwt: String)(implicit token: AuthToken): Unit = {
-      logger.info(s"Addidng user to NIH whitelist")
+      logger.info(s"Adding user to NIH whitelist: $jwt")
       postRequest(apiUrl(s"/api/nih/callback"), Map("jwt" -> jwt))
     }
   }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
@@ -240,7 +240,15 @@ trait Orchestration extends FireCloudClient with LazyLogging {
       parseResponse(getRequest(apiUrl(s"api/methods/$ns/$name/$snapshotId/permissions")))
     }
   }
-
+  /*
+   *  NIH requests
+  */
+   object NIH {
+    def addUserToNIH(jwt: String)(implicit token: AuthToken): Unit = {
+      logger.info(s"Addidng user to NIH whitelist")
+      postRequest(apiUrl(s"/api/nih/callback"), Map("jwt" -> jwt))
+    }
+  }
   /*
    *  Submissions requests
    */
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
@@ -50,7 +50,7 @@ object Config {
   object Users {
     private val notSoSecretPassword = users.getString("notSoSecretPassword")
     private val userDataJson = JSON.parseFull(scala.io.Source.fromFile(users.getString("userDataPath")).getLines.mkString).get.asInstanceOf[Map[String, Map[String,String]]]
-    val jwt = users.getString("jwt")
+    val tcgaJsonWebToken = users.getString("tcgaJsonWebTokenKey")
     def makeCredsMap(jsonMap: Map[String, String]): Map[String, Credentials] = {
       for((k,v) <- jsonMap) yield (k, Credentials(v, notSoSecretPassword))
     }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
@@ -50,7 +50,7 @@ object Config {
   object Users {
     private val notSoSecretPassword = users.getString("notSoSecretPassword")
     private val userDataJson = JSON.parseFull(scala.io.Source.fromFile(users.getString("userDataPath")).getLines.mkString).get.asInstanceOf[Map[String, Map[String,String]]]
-
+    val jwt = users.getString("jwt")
     def makeCredsMap(jsonMap: Map[String, String]): Map[String, Credentials] = {
       for((k,v) <- jsonMap) yield (k, Credentials(v, notSoSecretPassword))
     }
@@ -93,6 +93,7 @@ object Config {
     val rawlsApiUrl: String = fireCloud.getString("rawlsApiUrl")
     val samApiUrl: String = fireCloud.getString("samApiUrl")
     val thurloeApiUrl: String = fireCloud.getString("thurloeApiUrl")
+    val tcgaAuthDomain: String = fireCloud.getString("tcgaAuthDomain")
   }
 
   object ChromeSettings {
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
@@ -69,14 +69,5 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
    }
 }
 
-//case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
- // val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
-  //def validateLocation: Boolean = {
-   // testId("push-message").element != null
- // }
 
-  //def getRequestAccessText: String = {
-  //  readText (testId ("push-message"))
- // }
-//}
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
@@ -66,9 +66,11 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
 }
 
 case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
+  val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
   def validateLocation: Boolean = {
     testId("push-message").element != null
   }
+
   def getRequestAccessText: String = {
     readText (testId ("push-message"))
   }
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
@@ -65,3 +65,12 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
   }
 }
 
+case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
+  def validateLocation: Boolean = {
+    testId("push-message").element != null
+  }
+  def getRequestAccessText: String = {
+    readText (testId ("push-message"))
+  }
+}
+
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
@@ -63,16 +63,20 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
   def validateLocation: Boolean = {
     testId("message-modal-content").element != null
   }
+
+    def readMessageModalText: String = {
+      readText (testId ("message-modal-content"))
+   }
 }
 
-case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
-  val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
-  def validateLocation: Boolean = {
-    testId("push-message").element != null
-  }
+//case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
+ // val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
+  //def validateLocation: Boolean = {
+   // testId("push-message").element != null
+ // }
 
-  def getRequestAccessText: String = {
-    readText (testId ("push-message"))
-  }
-}
+  //def getRequestAccessText: String = {
+  //  readText (testId ("push-message"))
+ // }
+//}
 
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
@@ -121,13 +121,13 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
   "As a user with no TCGA permissions" - {
     "a published workspace" - {
       "with TCGA control access" - {
-        "should see a information message about getting TCGA access" ignore withWebDriver { implicit driver =>
+        "should see a information message about getting TCGA access" in withWebDriver { implicit driver =>
           //log in as a Curator and create/publish TCGA workspace
           val curatorUser = UserPool.chooseCurator
           implicit val curatorAuthToken: AuthToken = curatorUser.makeAuthToken()
 
           api.NIH.addUserToNIH(Config.Users.jwt)
-          withWorkspace(namespace, "PublishSpec_curator_TCGA_", Set(Config.FireCloud.tcgaAuthDomain)) { wsName =>
+          withWorkspace(namespace, "TCGA_", Set(Config.FireCloud.tcgaAuthDomain)) { wsName =>
             withCleanUp {
               val data = LibraryData.metadata + ("library:datasetName" -> wsName)
               api.library.setLibraryAttributes(namespace, wsName, data)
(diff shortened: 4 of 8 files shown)


automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+
+  "As a user with no TCGA permissions" - {
+    "a published workspace" - {
+      "with TCGA control access" - {
+        "should see a information message about getting TCGA access" in withWebDriver { implicit driver =>
+          //log in as a Curator and create/publish TCGA workspace
+          val curatorUser = UserPool.chooseCurator
+          implicit val curatorAuthToken: AuthToken = curatorUser.makeAuthToken()
+
+          api.NIH.addUserToNIH(Config.Users.jwt)
```
breilly2: Who is this "jwt" user? I worry that they're not in the automated test user pools and aren't going to be configured for FiaB instances.

Also, this line appears to cause a side effect that is never reverted. If it's important that this status change during the test, it will need to be reset for the next test run (and some thought put into the possibility of parallel test runs). If it just needs to be done once, it should be done while populating FiaB.

I had actually considered this problem a while back. This is, in fact, why I'd pushed for setting aside some auth domain users. We can chat more about it if you want.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
```diff
@@ -50,7 +50,7 @@ object Config {
   object Users {
     private val notSoSecretPassword = users.getString("notSoSecretPassword")
     private val userDataJson = JSON.parseFull(scala.io.Source.fromFile(users.getString("userDataPath")).getLines.mkString).get.asInstanceOf[Map[String, Map[String,String]]]
-
+    val jwt = users.getString("jwt")
```
breilly2: The test failures on this PR are because of the missing config. You'll need to make that change in the `firecloud-develop` repo. If you name it the same as this branch ("gd-library"), the next automated test run should pick it up.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -230,7 +230,15 @@ trait Orchestration extends FireCloudClient with LazyLogging {
       parseResponse(getRequest(apiUrl(s"api/methods/$ns/$name/$snapshotId/permissions")))
     }
   }
-
+  /*
+   *  NIH requests
+  */
+   object NIH {
+    def addUserToNIH(jwt: String)(implicit token: AuthToken): Unit = {
```
breilly2, line 247: It's not clear what `jwt` means. Is this supposed to be an email address or some other identifier? The parameter should be renamed to be something more descriptive of what it is, e.g. `email`, `name`, etc.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
```diff
@@ -230,7 +230,15 @@ trait Orchestration extends FireCloudClient with LazyLogging {
...
+      logger.info(s"Addidng user to NIH whitelist")
```
breilly2: It might be helpful (and shouldn't hurt) to log the user being added:
```
    logger.info(s"Addidng user to NIH whitelist: $jwt")
```
And just make sure that the `$jwt` part gets updated to match whatever name you choose for the parameter.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
```diff
@@ -93,6 +93,7 @@ object Config {
     val rawlsApiUrl: String = fireCloud.getString("rawlsApiUrl")
     val samApiUrl: String = fireCloud.getString("samApiUrl")
     val thurloeApiUrl: String = fireCloud.getString("thurloeApiUrl")
+    val tcgaAuthDomain: String = fireCloud.getString("tcgaAuthDomain")
```
breilly2, line 96: More config that may be missing from this build. This probably also needs to be populated in FiaB.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+          withWorkspace(namespace, "TCGA_", Set(Config.FireCloud.tcgaAuthDomain)) { wsName =>
+            withCleanUp {
+              val data = LibraryData.metadata + ("library:datasetName" -> wsName)
+              api.library.setLibraryAttributes(namespace, wsName, data)
+              register cleanUp api.library.unpublishWorkspace(namespace, wsName)
+              api.library.publishWorkspace(namespace, wsName)
+
+              //log in as a user with no TCGA access to make sure TCGA info message is displayed to you in Libraryval studentUser = UserPool.chooseStudent
```
breilly2: Nit: Stray "val studentUser = UserPool.chooseStudent" at the end of this comment.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+              val studentUser = UserPool.chooseStudent
```
breilly2: You test that a non-TCGA user must request access. However, you never check that a TCGA user can actually access the workspace. That's fine given the name of the test case, but then you probably don't need to be calling `addUserToNIH` above.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
```diff
@@ -357,7 +357,7 @@ class WorkspaceSpec extends FreeSpec with WebBrowserSpec with WorkspaceFixtures
         }
       }
       "and does have canCompute permission" - {
-        "should be able to launch analysis" ignore withWebDriver { implicit driver =>
+        "should be able to launch analysis" in withWebDriver { implicit driver =>
```
breilly2: I think this was disabled for being unreliable for some reason that I don't remember. I'm all for it being re-enabled but it will probably also need some changes to make it more stable.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+              withSignIn(studentUser) { _ =>
+                val page = new DataLibraryPage().open
+                page.hasDataset(wsName) shouldBe true
+                page.openDataset(wsName)
+                //verify that Request Access modal is shown
+                val requestAccessModal   = RequestAccessModal()
+                requestAccessModal.validateLocation shouldBe true
```
breilly2: Nit: A couple too many spaces before the `=`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
```diff
@@ -65,3 +65,14 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
   }
 }
 
+case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
+  val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
+  def validateLocation: Boolean = {
+    testId("push-message").element != null
+  }
+
+  def getRequestAccessText: String = {
```
breilly2: "get" is unnecessary. In fact, IntelliJ warns about functions beginning with "get" in certain circumstances. Instead, this could be something like `readRequestAccessText`... or even simply `messageText`; the class name `RequestAccessModal` gives context to what the text is for.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+                requestAccessModal.getRequestAccessText.contains(requestAccessModal.requestAccessText)
```
breilly2: This line evaluates to a boolean that is never checked. It should be:
```
requestAccessModal.getRequestAccessText should contain(requestAccessModal.requestAccessText)
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +115,47 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+                requestAccessModal.clickOk()
```
breilly2: Is this right? I thought it just had a "Done" button which, at least in this case, maps to `clickCancel()` since it's simply dismissing the modal without doing anything.

breilly2 (changes requested):

andy7i:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
andy7i: The jwt refers to the JSON web token to link a user to our mock NIH. This parameter is the JWT from NIH login.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
andy7i: Also a typo in "Addidng"

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
andy7i: looks like ok-button is right

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
andy7i: We may want to store the jwt in Vault.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
andy7i: This modal is specific to the Library page so we should handle this in DataLibraryPage

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
andy7i: Agreed.. on the other hand, I believe there are no other tests with tcga so another option is to allow all curators be linked to NIH.

andy7i (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
breilly2: Ahh, I think I get it. It's a single token, not per-user information. The user added is the one the auth token is associated with. So this is much like the X-FireCloud for Thurloe?

I'm going suggest then that this be pulled out of config here in `Orchestration` rather than it being a parameter to this function.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
breilly2: Oh, and the one I was working with recently was on the workspace list page. Wow... too bad these are different/inconsistent. That's why I was confused about ok/cancel below.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
breilly2: We could use curators, though then we can't prove that it's the NIH linking, not being a curator, that gives them access. The best test would be one where a user tries to access but is blocked, is added to the whitelist, and finally tries to access and succeeds.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
andy7i: Hmm.. maybe we should standardize modal verification? Either:
modals should use the same data-test-id for all messages and we verify the text
OR the data-test-id is unique for the message and we verify that the data-test-id is present.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/config/Config.scala
breilly2: This should probably move from `Users` to maybe `FireCloud`. I thought this represented some special user. Or maybe just renaming it to `jsonWebToken` if that's what it really is. (I think the rename should be done regardless of which config section it ends up in.)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: I think we are confusing what curator does in this test case: I need curator to create a TCGA workspace (through the endpoint), and that is the end of curator user involvement. As per reversing it according to Matt that would involve several admin endpoints and might be unreliable, if we want to have curators without NIH access we should consider adding a curator with it and using that specific user in this test 
case (and others that call for NIH access)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: That would drive me crazy too :-)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: I do, to make sure that workspace created by curator is TCGA protected

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: this particular modal only has OK button

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceSpec.scala
gdlex4015: Inadvertent change

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
breilly2: Yes, I was confused about exactly what was happening in this test. I understand now and made some suggestions in person, which I will write in appropriate places in this PR as well.

breilly2: Actually, the appropriate place is here. :)

We decided that a basic feature of curator accounts will be that they're on the NIH whitelist. This is fine as long as we don't need to test cases where curators are not on the whitelist. To make sure that they are always on the whitelist, they will be added during FiaB populate as well as refreshed here in the tests. This line should be extracted to a function called something like `refreshNIHWhitelistStatus`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/api/Orchestration.scala
MatthewBemis: i think it's still useful to have this function take a jwt. different jwts may be associated with different users on different whitelists.

i.e. we may want a test scenario where a user has a JWT that will get them into the TARGET group but not the TCGA group, or vice versa. or a valid JWT that resolves to an NIH username that isn't on any of our whitelists

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
```diff
@@ -63,16 +63,20 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
...
+      readText (testId ("message-modal-content"))
+   }
 }
 
-case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
-  val requestAccessText = "For access to TCGA controlled data please apply for access via dbGaP"
-  def validateLocation: Boolean = {
-    testId("push-message").element != null
-  }
+//case class RequestAccessModal(implicit webDriver: WebDriver) extends OKCancelModal {
```
breilly2: Nit: please delete commented-out code.

breilly2 (approved):
One nit then 👍

andy7i (approved):

gdlex4015:
jenkins retest

gdlex4015:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
```diff
@@ -63,5 +63,11 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
   def validateLocation: Boolean = {
     testId("message-modal-content").element != null
   }
+
```
andy7i, line 66: fyi @ahaessly is adding the same functionality
https://github.com/broadinstitute/firecloud-ui/pull/1109/files#diff-7a49eedc1ad01e518a6547aec80ab6f9R67

andy7i:
jenkins retest

andy7i:
jenkins retest

PR #1092, opened 2017-12-12
Make table filtering more resilient to wrong types [fix attempt for GAWB-2872]

aednichols:
Unsurprisingly, the `string/lower-case` function does not work on things that are not strings:
```
client:cljs.user!{:conn 2}=> (clojure.string/lower-case 0)
#object[TypeError TypeError: s.toLowerCase is not a function]
```
```
client:cljs.user!{:conn 2}=> (clojure.string/lower-case :a)
#object[TypeError TypeError: s.toLowerCase is not a function]
```

Let's make everything a string then!
```
client:cljs.user!{:conn 2}=> (clojure.string/lower-case (str :a))
":a"
```
```
client:cljs.user!{:conn 2}=> (clojure.string/lower-case (str 0))
"0"
```
`source` is supposed to be a string, so this won't affect any case that is not already undefined behavior. I'm looking at you, `a.toLowerCase`.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/common/table/utils.cljs
+++ b/src/cljs/main/broadfcui/common/table/utils.cljs
@@ -43,6 +43,7 @@
     data))
 
 (defn- matches-filter-text [filter-tokens source]
+  (assert (string? source) (str "Attempting to filter table row text that is not a string: " source)) ; GAWB-2872
   (let [lc-source (string/lower-case (str source))]
     (every? (fn [word] (utils/contains lc-source word)) filter-tokens)))
 
--- a/src/cljs/main/broadfcui/common/table/utils.cljs
+++ b/src/cljs/main/broadfcui/common/table/utils.cljs
@@ -43,7 +43,7 @@
     data))
 
 (defn- matches-filter-text [filter-tokens source]
-  (let [lc-source (string/lower-case source)]
+  (let [lc-source (string/lower-case (str source))]
     (every? (fn [word] (utils/contains lc-source word)) filter-tokens)))
 
 (defn- filter-rows [{:keys [filter-text]} columns data]


MattPutnam (approved):
👍

src/cljs/main/broadfcui/common/table/utils.cljs
```diff
@@ -43,7 +43,8 @@
     data))
 
 (defn- matches-filter-text [filter-tokens source]
-  (let [lc-source (string/lower-case source)]
+  (assert (string? source) (str "Attempting to filter table row text that is not a string: " source)) ; GAWB-2872
```
breilly2, line 46: Are you throwing an exception here to help find the problem? I'm curious what `source` is that isn't a string and I think it's worth finding out. However, if CLJS exceptions are like in other languages, this will prevent `string/lower-case` from being called on anything other than a string so the `(str ...)` change below won't change anything.

breilly2 (commented):
Thank you for looking into this!

src/cljs/main/broadfcui/common/table/utils.cljs
aednichols: > this will prevent string/lower-case from being called on anything other than a string

Correct, this is what we want - the tests found a bug, and knowing what the bad data is will help us fix it.

Asserts are compiled out in production, so the `str` protects users from (console) JS errors.

breilly2: Ahh, good point. I knew they were suppressed somehow but didn't know they were compiled out.

The other option would be to not match if `source` is not a string. Since we don't know what it is when it's not a string, we don't know if it's appropriate to match on its string representation.

davidangb (approved):
:+1: good find! (did not test)

aednichols:
Test failure is FreeTrialSpec, which is just awaiting updating by Isaac

PR #1099, opened 2017-12-14
WS list: gracefully handle workspaces with incomplete definitions [GAWB-2872]

aednichols:
@zarsky-broad and I discovered that the workspace response, or some unknown transformation thereof taking place on the frontend, may sometimes contain illegal workspace definitions, along the lines of
```
{
  "workspaceSubmissionStats": {
    "lastSuccessDate": "2017-05-15T15:03:11.269Z",
    "lastFailureDate": "2017-05-15T15:03:01.503Z",
    "runningSubmissionsCount": 0
  },
  "accessLevel": "PROJECT_OWNER",
  "owners": ["test.firec@gmail.com"],
  "public": false,
  "workspace": {}
}
```
Instead of throwing `a.toLowerCase`, we now display an inert dummy. 

This feels like the least surprising thing we could do; we thought about filtering them out, but that could be even more confusing (_"I have 5 workspaces, why do I only see 4?"_ vs. _"Huh, something is seriously wrong with that one workspace."_).

Looks like this:

screenshot-2017-12-14 firecloud broad institute

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/workspaces_list.cljs
+++ b/src/cljs/main/broadfcui/page/workspaces_list.cljs
@@ -182,8 +182,8 @@
         domain-groups (map :membersGroupName auth-domain)]
     ;; GAWB-2872
     (let [{:keys [namespace name]} workspace-id]
-      (assert (string? namespace) (str "Cannot list workspace because it has no namespace." (when name (str " Name is \"" name "\"."))))
-      (assert (string? name) (str "Cannot list workspace because it has no name." (when namespace (str " Namespace is \"" namespace "\".")))))
+      (assert namespace (str "Cannot list workspace because it has no namespace." (when name (str " Name is \"" name "\"."))))
+      (assert name (str "Cannot list workspace because it has no name." (when namespace (str " Namespace is \"" namespace "\".")))))
     (assoc ws
       :workspace-id workspace-id
       :workspace-column-sort-value (mapv string/lower-case (replace workspace-id [:namespace :name]))
--- a/src/cljs/main/broadfcui/page/workspaces_list.cljs
+++ b/src/cljs/main/broadfcui/page/workspaces_list.cljs
@@ -412,7 +412,15 @@
    (fn [{:keys [props state]}]
      (let [{:keys [server-response]} @state
            workspaces (map
-                       (fn [ws] (assoc ws :status (common/compute-status ws)))
+                       (fn [ws]
+                         ;; GAWB-2872
+                         (if (or (nil? (:name (:workspace ws))) (nil? (:namespace (:workspace ws))))
+                          {:workspace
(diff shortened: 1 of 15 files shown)


src/cljs/main/broadfcui/page/workspaces_list.cljs
```diff
@@ -412,7 +412,15 @@
    (fn [{:keys [props state]}]
      (let [{:keys [server-response]} @state
            workspaces (map
-                       (fn [ws] (assoc ws :status (common/compute-status ws)))
+                       (fn [ws]
+                         ;; GAWB-2872
+                         (if (or (nil? (:name (:workspace ws))) (nil? (:namespace (:workspace ws))))
```
zarsky-broad: you should try: 
```cljs
(let [{:keys [name namespace]} (:workspace ws)]
  (if (and name namespace)
    (assoc ws ...
```

zarsky-broad (approved):

zarsky-broad:
We should fix unit tests before we merge this.

aednichols:
jenkins retest

aednichols:
Having trouble telling whether they're my code or just noise

davidangb:
Some questions:

1. How could I reproduce this locally, so I can see it in action?
2. In lieu of the above ... what happens if I click on one of those workspaces? They say No Access, so I assume nothing ... but wanna check. 
3. In lieu of the above ... any impact to search/filters, pagination, etc?

breilly2:
Let's chat about this. This may eliminate the errors, but it doesn't address the underlying problem and, I fear, takes some of the heat off of actually fixing it. We need to add something here to gather information so that we can better understand what's happening.

aednichols:
@davidangb and @breilly2:

We do not have reliable repro steps nor a root cause. To simulate the bug and examine the behavior of this PR's code, you can replace
```
(get-in server-response [:workspaces-response :parsed-response])
```
with
```
(concat (get-in server-response [:workspaces-response :parsed-response]) (vec (repeat 10 {})))
```
in `workspace_list.cljs`. (You get 10 dummy workspaces at the end of your WS list; can also make the same change on top of `develop` and compare behaviors.)

I manually tested paging, search, and filtering - no unexpected behaviors. Filtering specifically works pretty sensibly, the "dead" workspaces fall under, in respective categories, "No Access", "Exception", and "Un-published".

davidangb (approved):
:+1: I'm okay with this; I think Brian has a point about finding the culprit but think we could push this in. If the problem goes away, we have a pretty good idea of what's happening.

PR #1108, opened 2018-01-05
GAWB-2634 SWAT OrchestrationApiSpec: grant/remove google role access

jmthibault79:
- update SamApiSpec to match proxy group endpoint changes

- [x] **Submitter**: Include the JIRA issue number in the PR description
- [x] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [x] **Submitter**: If you changed a URL that is used elsewhere (e.g. in an email), comment about where it is used and ensure the dependent code is updated.
- [x] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [ ] **Submitter**: If you're adding new automated UI tests, review the test plan with QA
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Product Owner** sign off
- [ ] **Submitter**: Verify all tests go green, including CI tests and automated UI tests.
- [ ] **Submitter**: Squash commits and merge to develop. If adding test code, merge application code and test code at the same time.
- [ ] **Submitter**: Delete branch after merge
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

PR #1110, opened 2018-01-08
Corrects some problems with a previous PR for GAWB-2634

jmthibault79:
I was making some bad assumptions for the GAWB-2634 Swat tests.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [x] **TL** sign off
- [x] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
@@ -1,5 +1,6 @@
 package org.broadinstitute.dsde.firecloud.test.api.orch
 
+import akka.http.scaladsl.model.StatusCodes
 import com.google.api.client.googleapis.json.GoogleJsonResponseException
 import com.google.api.services.bigquery.model.GetQueryResultsResponse
 import org.broadinstitute.dsde.firecloud.api.{APIException, Orchestration, Rawls}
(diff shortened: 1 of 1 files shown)


jmthibault79:
See the previous version at #1108

jmthibault79:
I requested review from a bunch of people who might want to weigh in.  I'm not going to assume you all will :)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
```diff
@@ -94,37 +95,21 @@ class OrchestrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
...
-        val removeEx = intercept[APIException] {
-          Orchestration.billing.removeGoogleRoleFromBillingProjectUser(projectName, user.email, role)(userToken)
-        }
-        removeEx.getMessage should include(errorMsg1)
-      }(ownerToken)
-
-      // owner is not an owner of this project
-      val errorMsg2 = "The caller does not have permission"
 
-      val otherProject = "broad-dsde-dev"
```
jmthibault79: I was running tests in dev with a user who actually *was* a broad-dsde-dev owner.  Oops.  These are not actually two different cases.

helgridly (approved):
this seems to make sense, i think

jmthibault79:
Tests pass!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
```diff
@@ -94,37 +95,21 @@ class OrchestrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
...
+      val errorMsg = "You must be a project owner"
+      val unownedProject = "broad-dsde-production"
```
breilly2: If something changed and the test failed, we could actually side-effect broad-dsde-procuction. I think we should use some other project to avoid the chance of doing that.

If there are side-effects due to a test failure, I guess you should probably consider having to undo those... though you might not have permissions to do so. Might be best to create a project in the test case.

The more I suggest that and talk about test runtime (as in this morning's meeting), the more I want to build something special for billing projects so we can reuse them for multiple test cases.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
```diff
@@ -94,37 +95,21 @@ class OrchestrationApiSpec extends FreeSpec with Matchers with ScalaFutures with
...
-        Orchestration.billing.addGoogleRoleToBillingProjectUser(otherProject, user.email, role)(ownerToken)
+        Orchestration.billing.addGoogleRoleToBillingProjectUser(unownedProject, user.email, role)(ownerToken)
       }
-      addEx.getMessage should include(errorMsg2)
+      addEx.getMessage should include(errorMsg)
+      addEx.getMessage should include(StatusCodes.Forbidden.intValue.toString)
 
       val removeEx = intercept[APIException] {
-        Orchestration.billing.removeGoogleRoleFromBillingProjectUser(otherProject, user.email, role)(ownerToken)
+        Orchestration.billing.removeGoogleRoleFromBillingProjectUser(unownedProject, user.email, role)(ownerToken)
```
breilly2: This is probably fine, however I can't not point out that this request doesn't make sense because `user.email` is (hopefully!) not in broad-dsde-production.

breilly2 (changes requested):
Sorry, I have to disagree with using broad-dsde-production in a test.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
jmthibault79: This is a very good point.

jmthibault79: Talked about this with Hussein - it does not actually matter for this functionality whether the user is a member of the google project.  Consider that we have many FC users who aren't members of *any* google project.  It does make sense to use this endpoint to give access to these users.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
jmthibault79: Updated to not use `broad-dsde-production` - instead uses `broad-dsde-dev` and two non-owners.

jmthibault79:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/api/orch/OrchestrationApiSpec.scala
breilly2: My comment here was specifically about "remove". I agree that it shouldn't matter; I'd expect the code to check permissions for the actor on the project first and never even get to checking for the target user on the project. However I like to make sure to think through these implementation assumptions before building them into test code so I have confidence that the test will remain valid in the future.

jmthibault79:
jenkins retest

breilly2 (approved):
👍  Joel explained to me in person that management of this Google role is specific to a project but completely separate from membership in the billing project. (Thought for the future... would it make sense for this to be a type of project membership that could be managed in the same place in the UI?)

gdlex4015 (approved):

PR #1115, opened 2018-01-10
Gd dsdeqa 164

gdlex4015:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/LibraryData.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/LibraryData.scala
@@ -1,27 +1,6 @@
 package org.broadinstitute.dsde.firecloud.fixture
 
 object LibraryData {
-  val metadata = Map(
-    "library:datasetName" -> "name",
-    "library:datasetVersion" -> "v1.0",
-    "library:datasetDescription" -> "desc",
-    "library:datasetCustodian" -> "cust",
-    "library:datasetDepositor" -> "depo",
-    "library:contactEmail" -> "name@example.com",
-    "library:datasetOwner" -> "owner",
-    "library:institute" -> Seq("inst", "it", "ute"),
-    "library:indication" -> "indic",
-    "library:numSubjects" -> 123,
-    "library:projectName" -> "proj",
-    "library:datatype" -> Seq("data", "type"),
-    "library:dataCategory" -> Seq("data", "category"),
-    "library:dataUseRestriction" -> "dur",
-    "library:useLimitationOption" -> "skip",
-    "library:studyDesign" -> "study",
-    "library:cellType" -> "cell",
-    "library:requiresExternalApproval" -> false,
-    "library:technology" -> Seq("is an optional", "array attribute"),
-    "_discoverableByGroups" -> Seq("all_broad_users"))
 
   val consentCodes = Map(
     "library:useLimitationOption" -> "questionnaire",
@@ -36,4 +15,31 @@ object LibraryData {
     "library:RS-G" -> "N/A",
     "library:NAGR" -> "No")
 
+  val metadataBasic = Map(
+     "library:datasetName"->"name",
+     "library:datasetVersion"->"v1.0",
+     "library:datasetDescription"->"desc",
+     "library:datasetCustodian"->"cust",
+     "library:datasetDepositor"->"depo",
+     "library:contactEmail"->"name@example.com",
+     "library:datasetOwner"->"owner",
+     "library:institute"->Seq("inst","it","ute"),
+     "library:indication"->"indic",
+     "library:numSubjects"->123,
+     "library:projectName"->"proj",
+     "library:datatype"->Seq("data","type"),
+     "library:dataCategory"->Seq("data","category"),
+     "library:dataUseRestriction"->"dur",
+     "library:studyDesign"->"study",
+     "library:cellType"->"cell",
+     "library:requiresExternalApproval"->false,
+     "library:useLimitationOption"-> "skip",
+     "library:technology"->Seq("is an optional","array attribute"),
+     "_discoverableByGroups"->Seq("all_broad_users"))
+
+   val metadataORSP = metadataBasic ++ Map(
+      "library:useLimitationOption"->"orsp",
+      "library:orsp"->"some orsp")
+
 }
+
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
@@ -54,6 +54,10 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
     testId("message-modal-content").element != null
   }
 
+  def readMessageModalText: String = {
+    readText (testId ("message-modal-content"))
+  }
+
   def getMessageText: String = {
     readText(testId("message-modal-content"))
   }
@@ -77,4 +81,6 @@ case class SynchronizeMethodAccessModal(implicit webDriver: WebDriver) extends O
   override def awaitReady(): Unit = await.visible(testId(grantButtonId), 1)
 
 }
+
+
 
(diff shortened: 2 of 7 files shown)


automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -97,13 +98,38 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+              api.library.publishWorkspace(namespace, wsName)
+              //clone workspace
+              withSignIn(curatorUser) { listPage =>
+                val workspaceNameCloned = "PublishSpec_curator_publish_cloned" + randomUuid
+                val workspaceSummaryPage = new WorkspaceSummaryPage(namespace, wsName).open
+                register cleanUp api.workspaces.delete(namespace, workspaceNameCloned)
+                workspaceSummaryPage.cloneWorkspace(namespace, workspaceNameCloned)
+                //Verify default group "All users"
+                val accessGroup = Orchestration.library.getDiscoverableGroups(namespace, workspaceNameCloned)
+                accessGroup.size shouldBe 0
```
breilly2: It's not at all clear to me that this checks what the comment says it's checking. Can you explain? Also, could this be checked through the UI instead of calling orch?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +142,51 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
...
+
+  "As a user with no TCGA permissions" - {
+    "a published workspace" - {
+      "with TCGA control access" - {
+        "should see a information message about getting TCGA access" in withWebDriver { implicit driver =>
+          //log in as a Curator and create/publish TCGA workspace
+          val curatorUser = UserPool.chooseCurator
+          implicit val curatorAuthToken: AuthToken = curatorUser.makeAuthToken()
+
+          api.NIH.addUserToNIH(Config.Users.tcgaJsonWebToken)
```
breilly2: I think I remember why this is here but it still looks very out of place to me. I definitely would like us to come up with a better way of managing this.

Can we at least extract a method named something like `makeSureCuratorIsInNihWhitelist`?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/LibraryData.scala
```diff
@@ -19,9 +19,10 @@ object LibraryData {
      "library:studyDesign"->"study",
      "library:cellType"->"cell",
      "library:requiresExternalApproval"->false,
-     "library:useLimitationOption"->"orsp",
+     //"library:useLimitationOption"->"orsp",
```
MatthewBemis: a couple of leftover comments here and below

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
MatthewBemis: after talking with @breilly2 we think that this would be a good opportunity to harden NIH related tests and make them a lot more immune to side effects.

because some of the NIH group data lives outside of our direct control and is shared across all FiaBs in an environment, there could be some unreliability in these types of tests.

our proposal: 

there is a special authdomain pool of users. this test should select one of those users to do a fresh NIH link and then do the tests. specifically:

1) at the beginning of the test, forcibly remove every trace that the user has of being NIH-linked. we have admin endpoints to remove them from the TCGA-dbGaP-Authorized workbench group and also to remove their NIH information from thurloe
2) do the link
3) do the test
4) do step 1 again

could implement this as a `withFreshNihLink(jwt: String)` function or however you see fit. it would be great to start this pattern now so when we build up the NIH related tests, there's an expectation that they're reliable

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
```diff
@@ -116,5 +142,51 @@ class PublishSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Wo
         }
       }
     }
+
+  }
+
+  "As a user with no TCGA permissions" - {
+    "a published workspace" - {
+      "with TCGA control access" - {
```
MatthewBemis, line 180: this text confused me at first. it might be more clear if it read "with TCGA-dbGaP-Authorized as part of the auth domain"

MatthewBemis (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
breilly2: Yes, I think it's definitely worth making some improvements here. We could put it off for now, but I'm sure we'd never get back to it so now is really the time.

breilly2 (changes requested):
Echoing Matt's comments and adding a couple additional ones.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: The endpoint getDiscoverableGroups checks what groups can see the dataset in the library, for open access the endpoint returns [] group that translates to All Users. It can be done through the UI but it is faster through API

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
gdlex4015: Not sure I understand, this endpoint adds user to NIH whitelist using JWT value of the admin user

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/library/PublishSpec.scala
breilly2: Please add this explanation as a code comment here.

andy7i (approved):
Brian and Matt's comments, then 👍

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/LibraryData.scala
```diff
@@ -19,9 +19,14 @@ object LibraryData {
...
+     "library:useLimitationOption"-> "skip",
      "library:technology"->Seq("is an optional","array attribute"),
-     "library:orsp"->"some orsp",
+     //"library:orsp"->"some orsp",
```
MatthewBemis: you can get rid of these commented out lines now that you have `metadataORSP`

gdlex4015:
jenkins retest

gdlex4015:
jenkins retest

gdlex4015:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/Modals.scala
```diff
@@ -54,6 +54,10 @@ case class MessageModal(implicit webDriver: WebDriver) extends OKCancelModal {
     testId("message-modal-content").element != null
   }
 
+  def readMessageModalText: String = {
```
breilly2: The implementations of `readMessageModalText` and `getMessageText` are identical. We should keep one and eliminate the other. I would vote to keep `readMessageModalText` (unless one of them is actually not used anywhere in which case that one should just be removed).

breilly2 (approved):
👍  One small change and then okay to go without re-review.

I only gave it a quick look, mostly assuming that it's the same changes as the last time I looked at it. Hard to tell for sure because of the rebase.

PR #1150, opened 2018-01-29
Downgrade selenium/node-chrome to 3.8.1-erbium

rtitle:
See if this fixes failing fiab-test runs...

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/docker/hub-compose-fiab.yml
+++ b/automation/docker/hub-compose-fiab.yml
@@ -18,7 +18,7 @@ services:
 
 
   chrome:
-    image: selenium/node-chrome
+    image: selenium/node-chrome:3.8.1-erbium
     depends_on:
       - hub
     links:
--- a/automation/docker/hub-compose.yml
+++ b/automation/docker/hub-compose.yml
@@ -15,7 +15,7 @@ services:
 
 
   chrome:
-    image: selenium/node-chrome
+    image: selenium/node-chrome:3.8.1-erbium
     depends_on:
       - hub
     links:


breilly2 (approved):
👍

andy7i (approved):
👍

aednichols (approved):
Even if we weren't trying to fix something, I like the idea of pinning so the platform is more predictable over time

breilly2:
@aednichols The problem is that this dependency locks in the version of Chrome. Since Chrome auto-updates in the wild, our test environment also has to stay up-to-date to make sure FC doesn't break when people's browsers get updated.

An alternative to always taking the latest version is to lock the version and manually respond to updates. The advantage is that incompatibilities don't immediately affect everyone. The disadvantage is that we have to be very proactive about watching for updates and manually bumping the version after making sure everything works. (This is probably similar to BITS deciding whether or not to vet software updates before letting them automatically roll out to Broadies' computers.)

aednichols:
I thought the Docker way of running Selenium locally controlled the Chrome version. No?

breilly2:
I'm not sure I follow. May need to discuss in person.

We can control the Chrome version used for our tests. We cannot control the Chrome version in the wild which we have to support. Therefore, one way or another, we have to keep our tests running on the latest versions.

aednichols:
It's not important - but I thought you were referring to dev workstations when you said in the wild

breilly2:
Ah, by "in the wild" I meant end-users' computers... but really applies to any workstation since Chrome is pretty aggressive about updating itself.

PR #1199, opened 2018-02-25
close open Modal before Sign Out to avoid click WebDriverException

aweng98:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
@@ -108,6 +108,7 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
                   workspaceListPage.validateLocation()
                   // TODO: add assertions for the new "request access" modal
                   // TODO: end test somewhere we can access the sign out button
+                  new RequestAccessModal().cancel()
                 }
               }
             }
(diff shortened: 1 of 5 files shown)


aweng98:
https://broadinstitute.atlassian.net/browse/GAWB-3235

andy7i:
I can see the usefulness of `closeModal`. I think it would be more foolproof and less code if we make `signOut()` handle any leftover open modals. One way we could do this is to reload the page or load a page such as workspaces list page, profile page, etc.

breilly2:
FYI, the way background sign-in works, a reload forgets the signed-in status... effectively "logging out". Also, the way background sign-in (used by almost all tests) works, there really isn't a sign-out from Google's perspective. I think another solution to this whole thing is to not bother trying to sign out from most tests.

ahaessly (approved):
👍

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -108,6 +108,7 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
                   workspaceListPage.validateLocation()
                   // TODO: add assertions for the new "request access" modal
                   // TODO: end test somewhere we can access the sign out button
+                  new RequestAccessModal().cancel()
```
MattPutnam: This seems brittle. If the modal doesn't appear, then this will fail. I think that could happen if we change the way the FiaB is populated, e.g. who's the owner on each billing project. It would probably be better to just do a `page.closeModal()` here (and comment that it's the request access modal).

andy7i:
@breilly2 Agreed on your point. We could eliminate the signOut from test cleanup. If we want to keep a form of signout, signing out via Google might be the most foolproof way`https://mail.google.com/mail/logout?hl=en`

aweng98:
I don't want to change test behavior. not sign out at the end of test is changing it. We should add real sign out via Google in a separate smoke test.

PR #1221, opened 2018-03-08
[BillingSpec] break out one test but consolidate the rest

MatthewBemis:
Consolidate all of billing spec into 1 test with a newly created billing project. Also retry creating the billing project up to 3 times.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
@@ -8,10 +8,10 @@ import org.broadinstitute.dsde.firecloud.fixture.{TestData, UserFixtures}
 import org.broadinstitute.dsde.firecloud.page.billing.BillingManagementPage
 import org.broadinstitute.dsde.firecloud.page.workspaces.methodconfigs.WorkspaceMethodConfigDetailsPage
 import org.broadinstitute.dsde.workbench.auth.AuthToken
-import org.broadinstitute.dsde.workbench.config.{Config, UserPool}
+import org.broadinstitute.dsde.workbench.config.{Config, Credentials, UserPool}
 import org.broadinstitute.dsde.workbench.fixture._
 import org.broadinstitute.dsde.workbench.model.{UserInfo, WorkbenchEmail, WorkbenchUserId}
-import org.broadinstitute.dsde.workbench.service.{Google, Rawls}
+import org.broadinstitute.dsde.workbench.service.{Google, Orchestration, Rawls}
 import org.broadinstitute.dsde.workbench.service.test.{CleanUp, WebBrowserSpec}
 import org.scalatest.{FreeSpec, Matchers}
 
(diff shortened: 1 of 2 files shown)


MatthewBemis:
jenkins retest

MatthewBemis:
jEnKiNs ReTeSt

MatthewBemis:
jenkins retest

dvoet:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
```diff
@@ -23,110 +23,115 @@ class BillingSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Cl
...
+    logger.info(s"Creating billing project: $billingProjectName")
+
+    billingPage.createBillingProject(billingProjectName, Config.Projects.billingAccount)
+    register cleanUp Rawls.admin.deleteBillingProject(billingProjectName, UserInfo(OAuth2BearerToken(user.makeAuthToken().value), WorkbenchUserId("0"), WorkbenchEmail(user.email), 3600))(UserPool.chooseAdmin.makeAuthToken())
+
+    val statusOption = billingPage.waitForCreateDone(billingProjectName)
+
+    statusOption match {
+      case None | Some("failure") if trials > 1 =>
+        logger.info(s"failure or timeout creating project $billingProjectName, retrying ${trials-1} more times")
```
gpolumbo-broad, line 126: Why do timeouts occur and how frequently do they occur?  For tests, simply retrying might be acceptable behavior, but in production, while not a failure, it's probably not ideal.  Should we do something more here to try to keep track of how frequently this timeout occurs and bubble that up as a metric that gets tracked?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
```diff
@@ -23,110 +23,115 @@ class BillingSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Cl
...
+          withWebDriver { implicit driver =>
+            withSignIn(user) { listPage =>
+              //BEGIN: Test creating billing project in UI
+              val billingPage = new BillingManagementPage().open
+              val billingProjectName = createNewBillingProject(user, billingPage)
+              //END: Test creating billing project in UI
+
+              //BEGIN: Test creating workspace
+              val workspaceName = "BillingSpec_makeWorkspace_" + randomUuid
+              register cleanUp api.workspaces.delete(billingProjectName, workspaceName)
```
gpolumbo-broad, line 71: All new to me, and I don't have the source for the `CleanUp` trait yet.  However, I'm going to assume that this trait knows how to respond to just about any failure anywhere in this test to ensure that any and all created assets are properly cleaned up after a successful run or any failure?  

Without knowing details, is there anything else we can or should be cleaning up out of this test?  Users?  Buckets?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
```diff
@@ -23,110 +23,115 @@ class BillingSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Cl
...
+         * Clearly a lot of functionality bleeds between these test cases. In order to save runtime during
+         * the tests, we're combining all five. The idea is that if any piece of basic functionality
+         * in these tests was broken (creating a workspace in a new project, creating a new project, etc), then
+         * most of these tests would fail anyway so grouping them together isn't much worse.
+         *
+         * NOTE: this single test case MUST create a brand new billing project each time it's run. If we use a
+         * project from GPAlloc, this entire test becomes worthless and our test coverage for billing is garbage.
+         *
+         * E comes last because it makes the project unusable.
+         */
```
breilly2, line 50: Excellent documentation! Thank you!

gpolumbo-broad (approved):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
```diff
@@ -23,110 +23,115 @@ class BillingSpec extends FreeSpec with WebBrowserSpec with UserFixtures with Cl
...
+              newBillingAccount shouldBe None
+              //END: should be able to change the billing account associated with the project
             }
           }
         }
       }
     }
   }
+
+  private def createNewBillingProject(user: Credentials, billingPage: BillingManagementPage, trials: Int = 3): String = {
```
breilly2, line 115: This bothers me for 2 reasons:
1. We will not notice these log messages and will therefore not be aware of changes in reliability. Getting logs into Kibana could help with this, especially if we can set up an alert if it happens too frequently.
2. With 3 retries, we are saying that we accept as low as ~33% reliability. A secondary check that gives us an alert while not failing the test could help with this.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
breilly2, line 126: Maybe `warn` rather than `info`.

aweng98:
can you break this test to two smaller sized tests? it helps with reducing overall execution time.

breilly2 (commented):
I will approve as long as we have some visibility/tracking of the retries. Doesn't have to be immediate as long as we have a plan but we _must ***not***_ to do this.

breilly2:
@aweng98 I disagree. Splitting it will not speed it up. Each one would need to create a new billing project which (other than running a workflow) is the time-consuming operation. Even running in parallel, splitting would not improve speed; also, It would decrease reliability.

dvoet:
I have another PR coming in rawls to log failures. My hope is that these will eventually go to kibana and we can monitor/diagnose. As it stands there is no way other than error to get any attention.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
dvoet: does it really matter?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
dvoet: look in the workbench-libs repo for the source.

cleaning up the workspace cleans up the bucket (and so does cleaning up the project which is below). Users are cleaned when the fiab is torn down (fiab-stop).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/billing/BillingSpec.scala
breilly2: Doubtful. I probably tend to use info when I should use debug or trace so this feels more important to me than those. I don't really care a lot. I should have said "nit".

dvoet:
I understand everyone's concerns over retrying this but I still think it is required for stability of the tests. We need to rebuild this functionality at some point soon anyway and maybe make it more reliable. In production there are 1200 projects and 59 failures. 5% failure rate, not great but ok. I don't want 5% failure rate in out tests.

breilly2 (approved):
Gah! A word disappeared from my previous review. I meant to say, "we must not _forget_ to do this." As long as we have a plan (and we don't forget to execute it), then I reluctantly agree that we're at the point where we need to do this for reliability reasons.

aweng98:
@breilly2  In terms of testing, is it valid to test creating new billing-projects concurrently? two test files may or may not run at same time. but when they do run at same time, even though it's just creating 2 billing-projects, I think that's a good testing. I think splitting worth consideration.
but that's outside scope of this PR.

breilly2:
@aweng98 Sounds like you're asking about a concurrency test for billing project creation. We don't currently have one of those. This PR doesn't change that fact or attempt to change it. We would have to write a new case which, as you point out, is outside the scope of this PR... and I don't have particularly strong feelings about the need for such a test.

This is also also separate from test run speed/throughput which was the reason you originally suggested splitting the test and the point that, in this case, I was disagreeing with.

PR #1224, opened 2018-03-09
Fix page timing issues

aweng98:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
```diff
@@ -1,6 +1,6 @@
 package org.broadinstitute.dsde.firecloud
 
-import org.broadinstitute.dsde.workbench.service.test.{Awaiter, WebBrowserUtil}
+import org.broadinstitute.dsde.workbench.service.test.{WebBrowserUtil, Awaiter}
```
breilly2: Nit: I'm not sure what's going on with your editor but there's no reason for these import changes, which I'd actually say are now in the wrong order.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Clickable.scala
```diff
@@ -1,20 +1,20 @@
...
   */
-trait Clickable { this: Component =>
+trait Clickable extends LazyLogging { this: Component =>
   /**
     * Click on the element modeled by this Component
     */
   def doClick()(implicit webDriver: WebDriver): Unit = {
-    awaitVisible()
-    scrollToVisible()
-    awaitEnabled()
```
breilly2, line 17: I think we want to continue doing these 3 things here in `Clickable` where we know what conditions must be met before clicking on something. We don't want to rely on every subtype's implementation of `awaitReady` to do the right thing.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
```diff
@@ -60,7 +61,13 @@ abstract class Component(queryString: QueryString)(implicit webDriver: WebDriver
   def isVisible: Boolean = find(query).exists(_.isDisplayed)
   def isEnabled: Boolean = enabled(query)
 
-  override def awaitReady(): Unit = awaitVisible()
+  override def awaitReady(): Unit = {
+    // combined awaitVisible and awaitEnabled
+    await condition {
+      isVisible && isEnabled
+    }
+    scrollToVisible()
```
breilly2: I don't think we should be scrolling the view in `awaitReady`. From what I've seen, the only reason we use `scrollToVisible` at all is for clicking on things that may be off screen... and that has only been necessary with more recent versions of selenium (and may even be a bug that has been fixed and this is therefore not needed anymore).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
```diff
@@ -17,7 +17,8 @@ trait PageUtil[P <: Page] extends FireCloudView { self: P =>
     * @return the Page object
     */
   def open(implicit webDriver: WebDriver): P = {
-    go to this
+    webDriver.navigate().to(this.url)
```
breilly2: As discussed in person, please include some comments here about the difference between the two ways to navigate and why we're trying this experiment.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/user/SignInPage.scala
```diff
@@ -32,6 +32,8 @@ class SignInPage(val baseUrl: String)(implicit webDriver: WebDriver) extends Fir
 
   private val signInButton = Button(CSSQuery("#sign-in-button"))
 
+  def isLoaded = signInButton.isVisible
```
breilly2: Seeing the `isOpen` function on `Modal`, I kinda like that name for pages too; it matches the `open` function.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -12,12 +11,18 @@ import org.scalatest.selenium.Page
...
   extends WorkspacePage(namespace, name) with Page with PageUtil[WorkspaceMethodConfigDetailsPage] {
 
+  private val tab = TabBar()
+  private val methodConfigListpage = new WorkspaceMethodConfigListPage(namespace, name)
+
   override def awaitReady(): Unit = {
-    await condition isLoaded
-    await spinner "Checking permissions..."
+    await condition {
+      tab.isTabOpen("Method Configurations")
```
breilly2: This probably isn't a problem right now, but this won't always hold if another tab is loaded. Just need to watch out for calling `awaitReady` at times other than when the page is first opened.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -92,7 +97,7 @@ class WorkspaceMethodConfigDetailsPage(namespace: String, name: String, methodCo
   }
 
   def isLoaded: Boolean = {
-    openLaunchAnalysisModalButton.isVisible
+    openLaunchAnalysisModalButton.isVisible && openLaunchAnalysisModalButton.isStateEnabled
```
breilly2: We should have a function on `Button` that encapsulates these two checks.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -17,6 +17,8 @@ import org.scalatest.selenium.Page
 class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver: WebDriver)
   extends WorkspacePage(namespace, name) with Page with PageUtil[WorkspaceSummaryPage] with Stateful {
 
+  wsSummaryPage: WorkspaceSummaryPage =>
```
breilly2: Style nit: this should be on the same line as the `{` above. Also, there might be another way to reference the containing page... I just can't think of it off the top of my head.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -56,18 +60,61 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
     val publishButton = Button("publish-button" inside this)
     val unpublishButton = Button("unpublish-button" inside this)
     val shareWorkspaceButton = Button("share-workspace-button" inside this)
+
```
breilly2, line 61: Some of the things below open modals. Wouldn't it be more correct to check for those modals than to just check for the parent page being ready?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -56,18 +60,61 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
...
+      shareWorkspaceButton.doClick()
+      await condition new AclEditor().isOpen()
+    }
   }
 
   private val workspaceAttributesArea = Collapse("attribute-editor", new FireCloudView {
     override def awaitReady(): Unit = {
       table.awaitReady()
-      awaitState("enabled")
+      wsSummaryPage.awaitReady()
```
breilly2, line 106: The more I see these, the more I think it's not right for sub components to delegate their ready state to the containing page. _Actions_ could result in the need to await for the page but sub components' state should be local to that component.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/FreeTrialSpec.scala
```diff
@@ -1,50 +1,55 @@
...
-    userAuthToken = testUser.makeAuthToken()
-    subjectId = Orchestration.profile.getUser()(userAuthToken)("userId").toString
-    Try(trialKVPKeys foreach { k => Thurloe.keyValuePairs.delete(subjectId, k)})
+    testUserAuthToken = testUser.makeAuthToken()
+    subjectId = Orchestration.profile.getUser()(testUserAuthToken)("userId").toString
+    Try(trialKVPKeys foreach { k => Thurloe.keyValuePairs.delete(subjectId, k)(adminAuthToken)})
   }
 
+   after {
+    Try(trialKVPKeys foreach { k => Thurloe.keyValuePairs.delete(subjectId, k)(adminAuthToken)})
```
breilly2: Good catch. This should definitely happen (proactively) after a test, not just (defensively) before a test.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/FreeTrialSpec.scala
```diff
@@ -1,50 +1,55 @@
...
+  }
+  
+
   private def setUpEnabledUserAndProject(user: Credentials): Unit = {
-    implicit val authToken: AuthToken = campaignManager.makeAuthToken()
-    api.trial.createTrialProjects(1)
+    implicit val managerAuthToken: AuthToken = campaignManager.makeAuthToken()
+    api.trial.createTrialProjects(1)(managerAuthToken)
```
breilly2: Either this doesn't need to be explicitly specified or `managerAuthToken` above doesn't need to be `implicit`. I'm not familiar enough with this test to know, but we should decide which token makes the most sense to be implicit and make that one, and no others, implicit. (I'm making a guess here that you're passing it explicitly because the compiler complained that there were too many implicit candidates.)

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
MattPutnam: I don't think scrollToVisible should be part of awaitReady

automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
```diff
@@ -9,7 +9,6 @@ import org.scalatest.selenium.WebBrowser
   */
 abstract class FireCloudView(implicit webDriver: WebDriver)
   extends Awaiter with WebBrowser with WebBrowserUtil {
-  def awaitReady(): Unit
```
MattPutnam, line 12: Why?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Modal.scala
```diff
@@ -21,6 +21,10 @@ abstract class Modal(id: String)(implicit webDriver: WebDriver) extends Componen
     clickXButton()
     awaitDismissed()
   }
+
+  def isOpen(): Boolean = {
+    find(CssSelectorQuery("body.broadinstitute-modal-open")).isDefined
+  }
```
MattPutnam: This isn't correct. This is checking if *any* modal is open. We can tell if a particular modal is open via `modal.isVisible`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/TabBar.scala
```diff
@@ -5,11 +5,20 @@ import org.openqa.selenium.WebDriver
...
     click on tab(name)
   }
 
+  /**
+    * If WebElement `tab` has some children div, then it's open.
+    * @param name
+    */
+  def isTabOpen(name: String): Unit = {
+    !findAll(s"${tab(name).queryString} > div").isEmpty
+  }
```
MattPutnam: I don't think this is correct. The tabs themselves don't have any children.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
```diff
@@ -17,7 +17,8 @@ trait PageUtil[P <: Page] extends FireCloudView { self: P =>
...
+    //go to this
```
MattPutnam: Why?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
MattPutnam: We don't need to be testing the state of the tab bar. This page class shouldn't know anything about the tab bar, that's WorkspacePage's responsibility.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
MattPutnam: I don't like this. We sometimes come to this page when we expect this button to stay disabled.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -12,12 +11,18 @@ import org.scalatest.selenium.Page
...
-    await spinner "Checking permissions..."
+    await condition {
+      tab.isTabOpen("Method Configurations")
+      isLoaded
+    }
+    await.spinner("Checking permissions...",15) // brittle. won't work if text changes
   }
 
-  override val url: String = s"${Config.FireCloud.baseUrl}#workspaces/$namespace/$name/method-configs/$methodConfigNamespace/$methodConfigName"
+  override val url: String = s"${methodConfigListpage.url}/$methodConfigNamespace/$methodConfigName"
```
MattPutnam: I don't like the idea of creating a separate page just to borrow its url.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
MattPutnam: Unnecessary

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -45,7 +47,9 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
   }
 
   private val sidebar = new Component(TestId("sidebar")) with Stateful {
-    override def awaitReady(): Unit = getState == "ready"
+    override def awaitReady(): Unit = {
+      await condition (getState == "ready")
+    }
```
MattPutnam: Can just use `awaitState("ready")`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/FreeTrialSpec.scala
```diff
@@ -103,20 +108,17 @@ class FreeTrialSpec extends FreeSpec with BeforeAndAfterEach with Matchers with
...
         assert(billingAccountUponEnrollment.nonEmpty, s"The user's project is not associated with a billing account.")
 
-        api.trial.terminateUser(testUser.email)
+        api.trial.terminateUser(testUser.email)(adminAuthToken)
 
         val billingAccountUponTermination = Google.billing.getBillingProjectAccount(billingProject.get)(trialAuthToken)
         val errMsg = "The trial user's billing project should have been removed from the billing account."
         assert(billingAccountUponTermination.isEmpty, errMsg)
-
-        registerCleanUpForDeleteTrialState()
```
MattPutnam: Why has this been deleted from all of these tests? Also, this one needs to get moved to the top.

MattPutnam (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
aweng98: isn't this already defined in Awaiter?
```
trait Awaiter {
  def awaitReady(): Unit
}
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/TabBar.scala
aweng98: inspecting DOM - when a tab is selected, two child divs rendered. only opened tab has child divs. this works.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Modal.scala
aweng98: good catch!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
aweng98: for this reason (a theory):
The driver.get() method is handled by the browser as a fragment identifier.
The navigate().to() method is handled by the browser as a address/location/URL bar input.

breilly2 (changes requested):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/FreeTrialSpec.scala
aweng98: oh this is not suppose be included. I will revert changes.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/FireCloudView.scala
aweng98: I blame this on Intellij. I don't know how to tell it NOT do it.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/PageUtil.scala
aweng98: I need to re-find the source. I will revert this change so PR is not blocked by it.

aweng98:
FreeTrialSpec change wasn't suppose to be included. It is reverted back.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
aweng98: to remove it, how do I then access outer `awaitReady` from inside of sidebar?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
aweng98: I think when parent component is "ready" or "enabled", all sub components are "ready" or "enabled". If one sub-component is ready but while other sub-components are not, it could makes UI to move or jitter. negative side-effect chain reaction.

breilly2: Interesting point. However it makes sense for a page's ready to be implemented by delegating down to the sub component's ready. If the sub component invokes the parent ready then it'll be an endless loop. The checks have to be one way. I think it's then up to the previous action to know how wide the await needs to be... either just a sub component or the whole page.

aweng98:
jenkins retest

aweng98:
jenkin slave out of space.
jenkins retest

aweng98:
jenkins retest

aweng98:
jenkins retest

aweng98:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/TabBar.scala
aweng98: new method is now removed. I need some time to come up with a better way handling check.

aweng98:
jenkins retest

aweng98:
jenkins retest

aweng98:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
```diff
@@ -12,12 +11,17 @@ import org.scalatest.selenium.Page
 class WorkspaceMethodConfigDetailsPage(namespace: String, name: String, methodConfigNamespace: String, val methodConfigName: String)(implicit webDriver: WebDriver)
   extends WorkspacePage(namespace, name) with Page with PageUtil[WorkspaceMethodConfigDetailsPage] {
 
+  private val tab = TabBar()
```
MattPutnam: Remove

MattPutnam (approved):

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
```diff
@@ -60,7 +61,11 @@ abstract class Component(queryString: QueryString)(implicit webDriver: WebDriver
   def isVisible: Boolean = find(query).exists(_.isDisplayed)
   def isEnabled: Boolean = enabled(query)
 
-  override def awaitReady(): Unit = awaitVisible()
+  override def awaitReady(): Unit = {
+    await condition {
+      isVisible && isEnabled
```
breilly2: `Clickable`s need to be enabled. I don't think that `Component`s in general should make this check.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
breilly2: Yeah, it's not ideal. You could extract a function into a companion object:
```
object WorkspaceMethodConfigListPage {
  def url(namespace: String, name: String): String = s"${Config.FireCloud.baseUrl}#workspaces/$namespace/$name/method-configs"
}
```
to enable reuse of that logic while avoiding extra object instantiation.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
breilly2: I still really dislike the parent references. Awaits should delegate down, not up.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -56,18 +58,61 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
...
+      wsSummaryPage.awaitReady()
     }
 
     val newButton = Button("add-new-button")
     val table = Table("workspace-attributes")
 
-    def clickNewButton(): Unit = newButton.doClick()
+    def clickNewButton(): Unit = {
+      newButton.doClick()
+      awaitReady()
```
breilly2, line 114: This is `awaitReady` on the `Collapse`? I wouldn't expect that clicking this button affects the `Collapse` component itself, rather the content inside the `Collapse`. Looks like we could wait on the key input with a `data-test-id` of `-key`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -132,7 +174,7 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
     */
 
   def share(email: String, accessLevel: String, share: Boolean = false, compute: Boolean = false, grantMethodPermission: Option[Boolean] = None): WorkspaceSummaryPage = {
-    sidebar.shareWorkspaceButton.doClick()
+    sidebar.clickShareWorkspaceButton()
```
breilly2:
```
val aclEditor = sidebar.clickShareWorkspaceButton()
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
```diff
@@ -148,14 +190,14 @@ class WorkspaceSummaryPage(namespace: String, name: String)(implicit webDriver:
   }
 
   def openShareDialog(email: String, accessLevel: String): AclEditor = {
-    sidebar.shareWorkspaceButton.doClick()
+    sidebar.clickShareWorkspaceButton()
```
breilly2:
```
val aclEditor = sidebar.clickShareWorkspaceButton()
```

breilly2 (changes requested):
Almost there. Mostly nits, however I still have a problem with the components awaiting their parents. I think this is the wrong direction to be delegating.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
aweng98: no, it's not `awaitReady` from `Collapse`. It's calling `workspaceAttributesArea.awaitReady` defined at line# 104.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
aweng98: changed

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/summary/WorkspaceSummaryPage.scala
aweng98: changed

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/methodconfigs/WorkspaceMethodConfigDetailsPage.scala
aweng98: done

aweng98:
reverted WorkspaceMethodConfigDetailsPage.scala

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Component.scala
aweng98: changed back

breilly2 (approved):
👍 Approving so we can thaw develop. However, I want to quickly (immediately) continue the discussion about how to structure awaits of pages and subcomponents.

aweng98:
I'll schedule a meeting to discuss your concern among ui automation
stakeholders. in meantime, we can also observe how well jenkin jobs go for
next couple days.

thank you
Alex

PR #1255, opened 2018-03-30
Enhanced error messages for /me [GAWB-3116]

aednichols:
- Adds some family-friendly details to the output of the "Error loading user information" screen
- Replaces the "Maintenance Mode" and "Server Unavailable" popups with equivalent banners
  - The popup with hard-coded generic information would routinely cover up a useful error message
  - The new banner coexists with the banners managed by Comms (see screenshots)

Definitely needs PO input, when PO is back in the office!

Requires https://github.com/broadinstitute/firecloud-orchestration/pull/615

---

Orch down, proxy throwing 503s:

screen shot 2018-03-30 at 4 48 13 pm

To simulate: run local UI against local Orch. Stop Orch with `re-stop` in the SBT console.

---

Missing auth:

screen shot 2018-03-30 at 4 49 20 pm

To simulate: modify [line 85 of `UserApiService.scala`](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala#L85) from
```
case None =>
```
to
```
case _ =>
```

---

Maintenance mode:

screen shot 2018-03-30 at 5 01 26 pm

To simulate: modify [line 85 of `UserApiService.scala`](https://github.com/broadinstitute/firecloud-orchestration/blob/develop/src/main/scala/org/broadinstitute/dsde/firecloud/webservice/UserApiService.scala#L85) from
```
case None =>
```
to
```
case _ =>
```
and line 86 from
```
respondWithErrorReport(Unauthorized, "No authorization header in request.", requestContext)
```
to
```
respondWithErrorReport(BadGateway, "Error of your choice", requestContext)
```

---

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/auth.cljs
+++ b/src/cljs/main/broadfcui/auth.cljs
@@ -214,20 +214,26 @@
         :not-active [:div {:style {:color (:exception-reds style/colors)}}
                      "Thank you for registering. Your account is currently inactive."
                      " You will be contacted via email when your account is activated."]
-        [:div {:style {:color (:state-exception style/colors) :lineHeight "2rem"}}
-         "Error loading user information. Please try again later." [:br]
-         (str "("(:message (:error @state)) ")")])])
+        [:div {:style {:lineHeight "2rem"}}
+         [:div {:style {:color (:state-exception style/colors)}}
+          "Error loading user information. Please try again later."]
+         [:div {:style {:color (:text-lighter style/colors)}} (str "What went wrong: " (:message (:error @state)))]])])
    :component-did-mount
    (fn [{:keys [props state]}]
      (ajax/call-orch "/me"
-                     {:on-done (fn [{:keys [success? status-code get-parsed-response]}]
+                     {:on-done (fn [{:keys [success? status-code get-parsed-response raw-response]}]
                                  (if success?
                                    ((:on-success props))
                                    (case status-code
                                      403 (swap! state assoc :error :not-active)
                                      ;; 404 means "not yet registered"
                                      404 ((:on-success props))
-                                     (swap! state assoc :error (get-parsed-response)))))}
+                                     ;; Borked servers often return HTML pages instead of JSON, so suppress JSON parsing
+                                     ;; exceptions because they are useless ("Unexpected token T in JSON...")
+                                     (let [[parsed-error parsing-error] (get-parsed-response false false)]
+                                       (swap! state assoc :error (if (some? parsing-error)
+                                                                   {:message (str "The FireCloud server returned a status code of " status-code ".")}
+                                                                   parsed-error))))))}
                      :service-prefix ""))})
 
 (react/defc RefreshCredentials
--- a/src/cljs/main/broadfcui/auth.cljs
+++ b/src/cljs/main/broadfcui/auth.cljs
@@ -214,10 +214,13 @@
         :not-active [:div {:style {:color (:exception-reds style/colors)}}
                      "Thank you for registering. Your account is currently inactive."
                      " You will be contacted via email when your account is activated."]
-        [:div {:style {:lineHeight "2rem"}}
-         [:div {:style {:color (:state-exception style/colors)}}
+        [:div {}
+         [:div {:style {:color (:state-exception style/colors) :paddingBottom "1rem"}}
           "Error loading user information. Please try again later."]
-         [:div {:style {:color (:text-lighter style/colors)}} (str "What went wrong: " (:message (:error @state)))]])])
+         [:table {:style {:color (:text-lighter style/colors)}}
+          [:tbody {:style {}}
+           [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "What went wrong:"] [:td {} (:message (:error @state))]]
+           [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "Status code:"] [:td {} (:statusCode (:error @state))]]]]])])
    :component-did-mount
    (fn [{:keys [props state]}]
      (ajax/call-orch "/me"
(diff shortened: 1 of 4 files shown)


src/cljs/main/broadfcui/auth.cljs
```diff
@@ -214,19 +214,30 @@
...
                      " You will be contacted via email when your account is activated."]
-        [:div {:style {:color (:state-exception style/colors)}}
-         "Error loading user information. Please try again later."])])
+        [:div {}
+         [:div {:style {:color (:state-exception style/colors) :paddingBottom "1rem"}}
+          "Error loading user information. Please try again later."]
+         [:table {:style {:color (:text-lighter style/colors)}}
+          [:tbody {:style {}}
+           [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "What went wrong:"] [:td {} (:message (:error @state))]]
+           [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "Status code:"] [:td {} (:statusCode (:error @state))]]]]])])
```
davidangb, line 223: consider using the `ErrorViewer` component here, to enable easy expanding/hiding of stack traces.

src/cljs/main/broadfcui/auth.cljs
```diff
@@ -214,19 +214,30 @@
...
                                      ;; 404 means "not yet registered"
                                      404 ((:on-success props))
-                                     (swap! state assoc :error true))))}
+                                     ;; Borked servers often return HTML pages instead of JSON, so suppress JSON parsing
+                                     ;; exceptions because they are useless ("Unexpected token T in JSON...")
+                                     (let [[error-json parsing-error] (get-parsed-response true false)]
+                                       (swap! state assoc :error (if parsing-error
+                                                                   {:message (str "Cannot reach the API server. The API server or one of its subsystems may be down.")
+                                                                    :statusCode status-code}
+                                                                   error-json))))))}
```
davidangb, line 240: nice!

src/cljs/main/broadfcui/main.cljs
```diff
@@ -249,9 +233,14 @@
...
-            (:maintenance-mode? @state)
-            #(swap! state dissoc :showing-system-down-dialog? :maintenance-mode?)))
+           (let [title (if (:maintenance-mode? @state) "Maintenance Mode"
+                                                       "Server Unavailable")
+                 msg (if (:maintenance-mode? @state) "FireCloud is currently undergoing planned maintenance. We should be back online shortly."
+                                                     "FireCloud service is temporarily unavailable.")]
+             (notifications/render-alert {:title title
+                                          :message msg
+                                          :link "http://status.firecloud.org/"
+                                          :cleared? false} nil)))
```
davidangb: in functional testing, this banner can get "stuck" visible. To reproduce using local UI + local orch:
1. start all services, healthy. Log in to the UI?. Navigate to workspace list. Notice all works.
2. shut down orch, navigate to Library using top nav. Notice banner.
3. start orch, navigate to Workspaces. Notice system is healthy but banner remains.

src/cljs/main/broadfcui/auth.cljs
aednichols: The `ErrorViewer` doesn't have an easy way to auto expand (possibly by environment, as described in the story), and I felt it was fairly important for status codes to be visible in screenshots (they are in the expandable section)

Will do some more tinkering and see if we can do better

src/cljs/main/broadfcui/main.cljs
aednichols: Yes, this is an excellent observation.

Right now, the AJAX "health monitor" that flags the system as down has no functionality to mark it as being back up.

With the popups, this works fine because the user just closes the popup. Unfortunately, the banner will persist until the user refreshes the page.

It's not immediately obvious to me what condition we should check to make the banner go away (a single successful AJAX seems too simplistic/naive), will put some thought into it

aednichols: After thinking about this a bit, I decided to implement the last paragraph, above.

When Orch gets into a state where it returns 5xx codes (what we care about for the calling FC "down"), it just uniformly returns that until it's back up. A single Orch AJAX success is the right thing to check, I believe.

aednichols:
Looks like this now:

screen shot 2018-04-04 at 11 18 26 am

davidangb (approved):
:+1: for the code.

On the UX side, I am definitely worried that in intermittent conditions, the new banner will be confusing to end users as it appears and disappears. However, I 100% realize this is no different than the current modal - it's just a change, so users can be disconcerted. I think we'll have to solve this through release notes/documentation/demo/etc.

aednichols:
Yeah, I will definitely write some nice release notes for this and run them by Moran.

src/cljs/main/broadfcui/auth.cljs
rushtong: 👍 I didn't know we could recover from a non-json server response. There used to be a devops task (that I cannot find in Jira any more) for dealing with the error html that comes back from orch's load balancers. That story, if it still exists, might be unnecessary with this.

rushtong (approved):
👍 Changes look good. Local config generation seems to be broken this morning ~so I can't test locally~ but was able to use an older config to test. Also looks good when there is no orch running at all: `Status code: 0`

src/cljs/main/broadfcui/auth.cljs
davidangb: I think we should keep the story in the backlog - it's also useful for server-to-server requests that expect json.

PR #1258, opened 2018-04-02
Convert AuthDomainSpec to GPAlloc

jmthibault79:
  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
@@ -7,7 +7,7 @@ import org.broadinstitute.dsde.firecloud.test.Tags
 import org.broadinstitute.dsde.workbench.auth.AuthToken
 import org.broadinstitute.dsde.workbench.config.{Config, Credentials, UserPool}
 import org.broadinstitute.dsde.workbench.fixture.{BillingFixtures, GroupFixtures, WorkspaceFixtures}
-import org.broadinstitute.dsde.workbench.service.{AclEntry, RestException, WorkspaceAccessLevel}
+import org.broadinstitute.dsde.workbench.service.{AclEntry, Orchestration, RestException, WorkspaceAccessLevel}
 import org.broadinstitute.dsde.workbench.service.Orchestration.billing.BillingProjectRole
 import org.broadinstitute.dsde.workbench.service.Orchestration.groups.GroupRole
 import org.broadinstitute.dsde.workbench.service.test.{CleanUp, WebBrowserSpec}
@@ -30,7 +30,6 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
   with BillingFixtures with GroupFixtures
   with UserFixtures {
 
-  val projectName: String = Config.Projects.common
   /*
    * Unless otherwise declared, this auth token will be used for API calls.
    * We are using a curator to prevent collisions with users in tests (who are Students and AuthDomainUsers), not
@@ -38,13 +37,6 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
    */
   val defaultUser: Credentials = UserPool.chooseCurator
   val authTokenDefault: AuthToken = defaultUser.makeAuthToken()
-
-  private def checkWorkspaceFailure(workspaceSummaryPage: WorkspaceSummaryPage, workspaceName: String): Unit = {
-    val error = workspaceSummaryPage.readError()
-    error should include(projectName)
-    error should include(workspaceName)
-    error should include("does not exist")
-  }
 
   private def checkWorkspaceFailure(workspaceSummaryPage: WorkspaceSummaryPage, projectName: String, workspaceName: String): Unit = {
     val error = workspaceSummaryPage.readError()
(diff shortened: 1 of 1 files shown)


breilly2:
My only concern is that, since this tests data access, false positives in these tests can have very bad consequences. Therefore, we need to be _VERY_ confident in `withCleanBillingProject`. What do you think @jmthibault79 @dvoet @helgridly ?

helgridly:
> Therefore, we need to be VERY confident in `withCleanBillingProject`

Or then you have sufficient non-`withCleanBillingProject` testery elsewhere. Which I thought we did?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -76,20 +69,22 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
...
         implicit val authToken: AuthToken = authTokenDefault
         withGroup("AuthDomain", List(user.email)) { authDomainName =>
-          withWorkspace(projectName, "AuthDomainSpec_share", Set(authDomainName), List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
-            withCleanUp {
-              withSignIn(user) { listPage =>
-                val summaryPage = listPage.enterWorkspace(projectName, workspaceName)
+          withCleanBillingProject(defaultUser) { projectName =>
+            withWorkspace(projectName, "AuthDomainSpec_share", Set(authDomainName), List(AclEntry(user.email, WorkspaceAccessLevel.Reader))) { workspaceName =>
+              withCleanUp {
+                withSignIn(defaultUser) { listPage =>
```
breilly2: If I remember correctly, the original reason for the additional user in this test was to verify specifically that a user other than the workspace owner can not widen the auth domain by cloning the workspace. While the same is true for the workspace owner, I think it's a better test if done by another user. Is there anything preventing us from doing:
```
withCleanBillingProject(defaultUser, List(user.email)) { ...
```
above and keeping the users as they were? Same applies to the 2 cases below related to multi-group auth domains.

If there is good reason for this change, the test should be simplified by removing the extra user altogether.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
```diff
@@ -459,11 +490,13 @@ class AuthDomainSpec extends FreeSpec /*with ParallelTestExecution*/ with Matche
               implicit val authToken: AuthToken = authTokenDefault
               withGroup("AuthDomain", List(user.email)) { groupOneName =>
                 withGroup("AuthDomain", List(user.email)) { groupTwoName =>
-                  withWorkspace(projectName, "AuthDomainSpec_share", Set(groupOneName, groupTwoName)) { workspaceName =>
-                    withSignIn(user) { listPage =>
-                      val summaryPage = listPage.enterWorkspace(projectName, workspaceName)
-                      summaryPage.readAuthDomainGroups should include(groupOneName)
-                      summaryPage.readAuthDomainGroups should include(groupTwoName)
+                  withCleanBillingProject(defaultUser) { projectName =>
+                    withWorkspace(projectName, "AuthDomainSpec_share", Set(groupOneName, groupTwoName)) { workspaceName =>
```
breilly2: This workspace needs to be created by the local `user` in order to test the billing project owner's special access (as long as they're in all of the groups making up the auth domain).

breilly2 (changes requested):
One important change and one that may be open for discussion.

dvoet:
@breilly2 I disagree. The actual project has no bearing on what is tested by this spec which is why the default project was ok to begin with. Nothing in this spec actually tests google permissions. There are some tests that care about project owner permissions but I would argue this is a step up. There was little control over project owners before, now there is.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
jmthibault79: Something I was surprised to learn yesterday: `withCleanBillingProject()` actually adds all members as project *OWNERS* - so it would be a good idea to audit all uses of this method for correctness.  But that should be a separate exercise.

For this correction I will do what you're looking for by adding `user` to the project separately.

helgridly:
You might want to consider getting/releasing a single GPAlloc project in before/afterAll if your tests play nicely together. every call to get a GPAlloc project pays a few-second penalty to register the project with rawls+sam. you can do this with `[claim|release]GPAllocProject`

helgridly (approved):

jmthibault79:
@helgridly Good idea.  I'll add that to the list of "auto-tests housekeeping" I'm collecting for Nicole after we are properly thawed.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/security/AuthDomainSpec.scala
jmthibault79: Updated tests with your corrections.  Thanks!

jmthibault79:
8 Test Errors at https://fc-jenkins.dsp-techops.broadinstitute.org/job/firecloud-ui-fiab-test-runner/1522/consoleFull :
* PreviewSpec `"Google Bucket:[]" was not equal to "Google Bucket:[ commons-dss-commons]"`
* AuthDomainSpec: 503 from gpalloc-qa.dsp-techops.broadinstitute.org
* AuthDomainSpec "can be seen but is not accessible": 403 "You must have the permission \"Billing Account User\" on billingAccounts/00708C-45D19D-27AAFA to create a project with it." 
* AuthDomainSpec "cannot be seen and is not accessible": 403 "You must have the permission \"Billing Account User\" on billingAccounts/00708C-45D19D-27AAFA to create a project with it.
* AuthDomainSpec "can be seen but is not accessible": 403 "You must have the permission \"Billing Account User\" on billingAccounts/00708C-45D19D-27AAFA to create a project with it.
* AuthDomainSpec "can be seen and is accessible": 403 "You must have the permission \"Billing Account User\" on billingAccounts/00708C-45D19D-27AAFA to create a project with it.
* AuthDomainSpec "the user has correct permissions": 403 "You must have the permission \"Billing Account User\" on billingAccounts/00708C-45D19D-27AAFA to create a project with it.
* DataSpec: workflow submission failed

5 of the AuthDomainSpecs are plausibly a result of these changes so I'll look into those.

jmthibault79:
jenkins retest

PR #1267, opened 2018-04-05
Fix GCS links showing pieces in wrong order [GAWB-2495, GAWB-1912]

aednichols:
I can't believe this actually works, but it does. (At least in Chrome.)

---

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/project/Settings.scala
+++ b/automation/project/Settings.scala
@@ -26,8 +26,14 @@ object Settings {
     "-Xmax-classfile-name", "100"
   )
 
+  // test parameters explanation:
+  // `-o` - causes test results to be written to the standard output
+  //     `F` - Display full stack traces
+  //     `D` - Display test duration after test name
+  //     `G` - show reminder of failed and canceled tests with full stack traces at the end of log file
+  // `-fW` - causes test results to be written to the summary.log without colored text
   val testSettings = List(
-    testOptions in Test += Tests.Argument("-oF", "-u", "test-reports")
+    testOptions in Test += Tests.Argument("-oFDG", "-u", "test-reports", "-fW", "test-reports/summary.log")
   )
 
   //common settings for all sbt subprojects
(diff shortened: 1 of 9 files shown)


MattPutnam:
Does this cover all cases? Looks like we should need it on the next line too. And also for DOS files?

aednichols:
The problem only seems to happen when the text fits the pattern `<number>.<something`.

`gs://asdf/4.jpg` is fine, so is `a1.jpg` - and DOS files seem to always display the full URI.

I'll test it some more tomorrow.

aednichols:
Tests failed because we added an extra (non-printable) character and Selenium looks for an exact thing

MattPutnam:
Would this work via `[:span {:style {:textDirection "ltr"}} ...]` instead?

jmthibault79:
Does this also cover GAWB-1912 ?

aednichols:
I think so. I'll check out the A/C and repro steps on 1912 and close it if possible.

MattPutnam (approved):
I'm skeptical of this being 100% correct, but it seems more correct than before at least.

aednichols:
@MattPutnam tried that, seemed to have no effect; also tried `[:div {:style {:direction "ltr"}} ...]` which fixed the order but broke the ellipsis

aednichols:
jenkins retest because my last test run was expunged from Jenkins

aednichols:
@MattPutnam: the weirdness continues.

The above commit fixes the bug from [GAWB-1912](https://broadinstitute.atlassian.net/browse/GAWB-1912).

The leading `&lrm;` I put in originally is _not_ sufficient to fix the quote being teleported to the start of the string. Maybe at some stage of parsing - possibly influenced by quotes - we are reading from right-to-left, so that's where the control character has to be in order to apply to the whole line?

I tested that ellipsis still works, `5.png` still works, and obviously that the quote is no longer teleported. Removing the leading `&lrm;` re-breaks `5.png`, so both are necessary.

jmthibault79:
Fancy!  How about making this a function?  Something that can be invoked like `(lrm-escape mystring)`

jmthibault79 (approved):

rushtong (approved):
👍 Very nice!

davidangb (approved):
:+1: did not test myself, but looks like a good hack. Nice job!

PR #1285, opened 2018-04-27
remove orch service test; it now lives in orch

davidangb:
This is the last piece of DataBiosphere/firecloud-app#10.

Delete OrchestrationApiSpec.scala; it now lives at https://github.com/broadinstitute/firecloud-orchestration/tree/develop/automation/src/test/scala/org/broadinstitute/dsde/test/api/orch and is running as part of orch-level service tests.


davidangb:
Unit tests fail with a segfault on slimerjs. This is a known issue; safe to merge with this failure.

jmthibault79 (approved):

PR #1287, opened 2018-05-02
downgrade to node-sass 4.6.1 to get upgraded transitive dependencies

davidangb:
This attempts to fix the CVE-2018-3728 vulnerability that github identified and is yelling about on the front page of this repo.

The vulnerability is in `hoek 2.16.3`, which is pulled in as a transitive dependency a few layers down: https://github.com/broadinstitute/firecloud-ui/network/dependencies

The direct dependency `node-sass` is the culprit; see https://github.com/sass/node-sass/issues/2355. The fix I'm proposing is to *downgrade* node-sass. Version 4.6.1 of node-sass is the latest that depends on `"request": "^2.79.0"`: https://github.com/sass/node-sass/blob/v4.6.1/package.json#L71, while version 4.9.0 of node-sass depends on `"request": "~2.79.0"`: https://github.com/sass/node-sass/blob/v4.9.0/package.json#L71.

This means that by using an earlier version of `node-sass`, we get a later version of `request`. The later version of `request` in turn pulls in (a couple layers down) a later version of `hoek` that doesn't have the vulnerability.

To get a clean slate and rid ourselves of the older versions of node-sass, request, hawk, boom, sntp, and hoek - all the packages in the vulnerable dependency chain - I completely rebuilt package-lock.json. It's going to show a huge diff; remember this is an auto-generated file (my process was to rm package-lock.json; rm -rf node_modules; npm install).

Finally - one of our transitive dependencies moved from using `ajv` as a dependency to using `ajv` as a peerDependency. Since peer dependencies aren't automatically installed, I added it to our direct dependency list in package.json.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7,12 +7,13 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
-      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q=="
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
+      "dev": true
     },
     "acorn": {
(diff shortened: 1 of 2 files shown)


davidangb:
jenkins retest to see if we get past fiab-start ...

zarsky-broad (approved):
It should be noted that this peer dependency is installed correctly, but there's an npm bug we're running into here: https://github.com/npm/npm/issues/15708.

Tested and working correctly!

ghost (approved):
Tested 👍

PR #1298, opened 2018-05-16
skip tests: billing accounts require user, not  admin

davidangb:
see https://software.broadinstitute.org/firecloud/documentation/article?id=9762

noticed while reviewing related functionality with PO. No ticket; faster to fix than to file and triage and get through the backlog.

as a text-only change, skipping tests.

billing


ghost (approved):
👍

PR #1322, opened 2018-06-15
GAWB-3625 Validate method configurations properly

helgridly:
Friends with https://github.com/broadinstitute/rawls/pull/945/

Updated in line with some backend changes that correctly check the validity of method config expressions when inputs on the method are optional. Upshot of this is that the endpoint returns more info about precisely _why_ your conf is wrong, so we handle that here.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodLaunchSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodLaunchSpec.scala
@@ -18,7 +18,7 @@ class MethodLaunchSpec extends FreeSpec /*with ParallelTestExecution*/ with Matc
   val methodConfigName: String = SimpleMethodConfig.configName + "_" + UUID.randomUUID().toString
   val wrongRootEntityErrorText: String = "Error: Method configuration expects an entity of type sample, but you gave us an entity of type participant."
   val noExpressionErrorText: String = "Error: Method configuration expects an entity of type sample, but you gave us an entity of type sample_set."
-  val missingInputsErrorText: String = "is missing definitions for these inputs:"
+  val missingInputsErrorText: String = "Missing inputs:"
   val testData = TestData()
 
   "launch workflow and delete a workflow" in {
--- a/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceOwnerSpec.scala
+++ b/automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/workspace/WorkspaceOwnerSpec.scala
@@ -220,8 +220,6 @@ class WorkspaceOwnerSpec extends FreeSpec with ParallelTestExecution with Matche
             withWebDriver { implicit driver =>
               withSignIn(user1) { listPage =>
                 api.importMetaData(billingProject, workspaceName, "entities", TestData().participantEntity)
-                api.methodConfigurations.createMethodConfigInWorkspace(billingProject, workspaceName, SimpleMethod, SimpleMethodConfig.configNamespace, s"$methodConfigName Config", 1,
-                  SimpleMethodConfig.inputs, SimpleMethodConfig.outputs, "participant")
                 val detailPage = listPage.enterWorkspace(billingProject, workspaceName)
                 val aclEditor = detailPage.openShareDialog(user2.email, "OWNER")
                 eventually { aclEditor.canComputeBox.isEnabled shouldBe false }
@@ -242,8 +240,6 @@ class WorkspaceOwnerSpec extends FreeSpec with ParallelTestExecution with Matche
             withWebDriver { implicit driver =>
               withSignIn(user1) { listPage =>
                 api.importMetaData(billingProject, workspaceName, "entities", TestData().participantEntity)
-                api.methodConfigurations.createMethodConfigInWorkspace(billingProject, workspaceName, SimpleMethod, SimpleMethodConfig.configNamespace, s"$methodConfigName Config", 1,
-                  SimpleMethodConfig.inputs, SimpleMethodConfig.outputs, "participant")
                 val detailPage = listPage.enterWorkspace(billingProject, workspaceName)
                 val aclEditor = detailPage.openShareDialog(user2.email, "READER")
                 eventually { aclEditor.canComputeBox.isEnabled shouldBe false }


src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -36,6 +36,12 @@
         entity-datums (map (partial str "this.") (get-in entity-types [(keyword selected-entity-type) :attributeNames]))]
     (concat entity-datums workspace-datums)))
 
+(defn- fix-validated-method-config [config]
+  (let
+    [munged-missing (map (fn [missing-input] {(keyword missing-input) "This input is declared in the method and missing in your method configuration."}) (:missingInputs config))
+     munged-extras (map (fn [extra-input] {(keyword extra-input) "This input is declared in your method configuration but not listed in the method."}) (:extraInputs config))
+     new-inputs (into (:invalidInputs config) (concat munged-missing munged-extras))]
+    (assoc config :invalidInputs new-inputs)))
```
helgridly: Being fairly lazy here, and folding the missing and extras lists into the list of invalid inputs with default messages, and pretends they were there all along. That way I don't have to update the MC render code.

MattPutnam (approved):

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
MattPutnam: Seems like a reasonable approach.

You can use `update` to modify a value in a map, instead of manually extracting->updating->reinserting. Something like:
```clojure
(update config :invalidInputs into munged-missing munged-extras)
```

Which means `munged-missing` and `munged-extras` can be left as lists-of-pairs instead of maps:
```clojure
(let [munged-missing (map (fn [missing-input] [(keyword missing-input) "text"]) (:missingInputs config))
      ...])
```

And then for advanced fun,
```clojure
(fn [missing-input] [(keyword missing-input) "text"])
```
can be
```clojure
(juxt keyword (constantly "text"))
```

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
```diff
@@ -355,13 +361,14 @@
...
    (fn [{:keys [state props]}]
      (endpoints/call-ajax-orch
       {:endpoint (endpoints/get-validated-workspace-method-config (:workspace-id props) (:config-id props))
        :on-done (fn [{:keys [success? get-parsed-response status-text]}]
                   (if success?
-                    (let [response (get-parsed-response)
+                    (let [response (fix-validated-method-config (get-parsed-response))
                           fake-inputs-outputs (fn [data]
                                                 (let [method-config (:methodConfiguration data)]
                                                   {:inputs (mapv (fn [k] {:name (name k)}) (keys (:inputs method-config)))
```
ansingh7115: I would remove the all to get-inputs-outputs if we are now validating against the WDL in the validate endpoint. 
I think this would be accomplished by replacing `(endpoints/call-ajax-orch {...})` with the contents of `fake-inputs-outputs`(which, now, are no longer fake).

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
ansingh7115: This is more philosophy, but I think `(juxt keyword (constantly "text"))` would be a little confusing to people who aren't ui grown-ups.

Agreed with everything else though :)

helgridly: > `(update config :invalidInputs into munged-missing munged-extras)`

I think one of my in-between attempts looked like:

```
(update config :invalidInputs munged-missing)
```

which only returned the new value of `invalidInputs`, not the new value of `config`. I'll doublecheck this works, but for my understanding: this works because `into` plays the role of `f` [here](https://clojuredocs.org/clojure.core/update), right?

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
helgridly: there's a bit of a behaviour change here that i want to think through.

if the underlying method doesn't exist, the call to `get-validated-workspace-method-config` will now return 404 and not the contents of the MC. so the code as-is will fall through to displaying simply that error and not the MC page at all.

this makes me think that we should first call the _unvalidated_ version of the MC so we have anything to display at all, and then use `get-validated-workspace-method-config` to  
a) determine if the underlying method exists (and if not we set `:redacted? true`)  
b) fill in the `:inputs-outputs` for the method itself  
c) get any validation error text to display.  

does this make sense to you?

ansingh7115: That does make sense - but if `get-validated-workspace-method-config` succeeds, we don't need to make the extra call. So, maybe instead of calling the unvalidated version first, what if we call `get-validated-workspace-method-config` first, and then only if that call fails do we make a call to get the unvalidated version?

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
ansingh7115: yes

src/cljs/main/broadfcui/page/workspace/method_configs/method_config_editor.cljs
helgridly: I thought about it some more.

If we call get-unvalidated-MC and get a 404, the MC doesn't exist and we display a page 404 like normal.

If we get an MC, we can then call get-validated-MC. If _that_ returns a 404, the underlying method doesn't exist, so we should set `redacted = true`. We also have to fake the underlying inputs and outputs, much like `fake-inputs-outputs`.

If it returns successful, we still need to infer the underlying method's inputs and outputs, which is a function _like_ `fake-inputs-outputs` but is a bit more involved, because it uses the `missingInputs` field (and others) from the ValidatedMethodConfiguration response.

I think this is overcomplicated compared to just calling `get-inputs-outputs`. Which I think means the call order should be:

1. get-unvalidated-MC (404 = no MC)
2. get-inputs-outputs (404 = no method in which case stop here, use fake-inputs-outputs, add a message saying "we're not validating your expressions until you give us a valid MC")
3. get-validated-MC (use to get real validation errors)

The existence of that message at step 2 is probably why the "evaluate my expressions without knowing anything about the underlying method" endpoint existed in the first place.

breilly2:
Temporarily closing while I push a change for safe keeping, rebase the branch locally, then force-push the branch again.

breilly2:
Can't reopen this PR after force-pushing the branch... even though force-pushing the branch is perfectly fine while the PR is open! 😡New PR: #1353

PR #1333, opened 2018-06-25
GAWB-3660: Infinite spinner due to error in Dockstore "Launch With Firecloud"

ansingh7115:
Talked to Dockstore people about this and created an issue here for getting better error messages:
https://github.com/ga4gh/dockstore/issues/1533 

Because this ticket was scoped to getting rid of the endless spinner, I propose this very simple solution in which we just show a generic error message. Currently, Dockstore error messages themselves are very generic (ticket above address that somewhat). 

Eventually, we want to make it so that we can display html error messages. That will take more work so I propose using this PR as an intermediate solution that at least gets rid of the infinite spinner and then we can figure out how to display html error messages. 

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/external_importer.cljs
+++ b/src/cljs/main/broadfcui/page/external_importer.cljs
@@ -68,15 +68,16 @@
      (let [{:keys [id version]} props]
        (endpoints/dockstore-get-wdl
         id version
-        (fn [{:keys [success? get-parsed-response]}]
+        (fn [{:keys [success? get-parsed-response status-code status-text]}]
           (if success?
             (swap! state assoc :wdl (:descriptor (get-parsed-response)))
-            (swap! state assoc :load-error true))))))
+            (swap! state assoc :load-error (str "(" status-code ") " status-text)))))))
    :-render-error
    (fn [{:keys [state]}]
      [:div {}
       [:div {:style {:margin "0.5rem 0"}}
-       "Error loading WDL. Please verify the workflow path and version and ensure this workflow supports WDL."]])
+       "Error loading WDL. Please verify the workflow path and version and ensure this workflow supports WDL."]
+      [:div {} "Message from Dockstore: " (:load-error @state)]])
    :-render-export
    (fn [{:keys [props state locals this]}]
      (let [{:keys [id]} props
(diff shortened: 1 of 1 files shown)


davidangb:
I think you could look for `raw-response` or `xhr` instead of `get-parsed-response` if you want the actual error message (but then you'd have to figure out how to display it). Nice simple solution.

ahaessly (approved):
👍
can you create a ticket for this follow on work?

ansingh7115:
Here's the ticket I made: https://broadinstitute.atlassian.net/browse/GAWB-3670

ansingh7115:
@davidangb 
The html _does_ show up in`raw-response` but, according to a conversation with Matt P., it is a little more complicated to implement rendering of html errors in FC than it was in Saturn (they've looked into doing it before). I created another ticket for that b/c GAWB-3660 was only scoped to figuring out why that particular method version was returning a 404 and I didn't want scope explosion on a 2 point ticket :P

ansingh7115:
jenkins retest

src/cljs/main/broadfcui/page/external_importer.cljs
```diff
@@ -71,13 +71,12 @@
         (fn [{:keys [success? get-parsed-response]}]
           (if success?
             (swap! state assoc :wdl (:descriptor (get-parsed-response)))
-            (swap! state assoc :load-error (get-parsed-response false)))))))
+            (swap! state assoc :load-error true))))))
```
breilly2: I think we should show a little bit of easy-to-access context: `status-code` and (when available) `status-text`. Something like:
```
(fn [{:keys [success? get-parsed-response status-code status-text]}]
  (if success?
    (swap! state assoc :wdl (:descriptor (get-parsed-response)))
    (swap! state assoc :load-error (str "(" status-code ") " status-text)))))))
```
And then restore the `ErrorViewer` or some other rendering of `:load-error`.

It might not mean much to everyone but some users might be able to do something with it or at least give it to us to make our support triage easier.

breilly2 (changes requested):
Small change but I think it's worth making.

src/cljs/main/broadfcui/page/external_importer.cljs
ansingh7115: That's better, I agree

breilly2 (approved):
👍 Looks good!

PR #1344, opened 2018-07-02
GAWB-3671: Fix quoting when populating from JSON

breilly2:
- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: If you changed a URL that is used elsewhere (e.g. in an email), comment about where it is used and ensure the dependent code is updated.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [ ] **Submitter**: If you're adding new automated UI tests, review the test plan with QA
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
  * Submitter makes further commits. DO NOT SQUASH
  * Submitter updates documentation as needed
  * Submitter **reassigns to LR** for further feedback
- [ ] **TL** sign off
- [ ] **LR** sign off
- [ ] **Product Owner** sign off
- [ ] **Submitter**: Verify all tests go green, including CI tests and automated UI tests.
- [ ] **Submitter**: Squash commits and merge to develop. If adding test code, merge application code and test code at the same time.
- [ ] **Submitter**: Delete branch after merge
- [ ] **Submitter**: **Test this change works on dev environment after deployment**. YOU own getting it fixed if dev isn't working for ANY reason!
- [ ] **Submitter**: Mark JIRA issue as resolved once this checklist is completed

PR #1350, opened 2018-07-08
GAWB-3551: Create Download link for input json

ansingh7115:
Depends a bit on Brian's PR https://github.com/broadinstitute/firecloud-ui/pull/1344/files

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/cljs/main/broadfcui/common/links.cljs
```diff
@@ -13,22 +13,22 @@
...
-        :download filename}
-    label
-    icons/download-icon]))
+  ([label url] (create-download label url nil false)) ; nil :download is fine with React
+  ([label url filename] (create-download label url filename false))
+  ([label url filename create-internal?]
+   (let [attributes {:href url :download filename} ]
+     (if create-internal?
+       (create-internal attributes label)
+       [:a attributes label icons/download-icon]))))
```
ansingh7115: This is so we can use `create-internal` for the Download link so that it's visually the same as the Populate Json link

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -78,18 +97,26 @@
                        {:on-upload
                         (fn [{:keys [file-contents]}]
                           (let [uploaded-inputs (utils/parse-json-string file-contents true)
-                                new-inputs (merge (:inputs @state) uploaded-inputs)]
+                                new-inputs (merge (:inputs @state) (into {} (map (fn [[k v]] [k (utils/->json-string v)]) uploaded-inputs)))]
                             (when-not editing? (begin-editing))
-                            (after-update #(doseq [[k v] (vec new-inputs)] ((k @locals) :set-value v)))
+                            (after-update #(doseq [[k v] (vec new-inputs)] (when (contains? @locals k) ((k @locals) :set-value v))))
```
ansingh7115: Added these two lines for now, these are in Brian's PR, will rebase off of his when it's merged. This PR should wait on that one.

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -22,6 +22,25 @@
 
 (def clip (partial merge table-style/clip-text))
 
+; ask what other types it could be. Maps?
```
breilly2: `Float`, `Boolean`, `Map`. Maybe `Object`?

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -22,6 +22,25 @@
...
+(defn get-typed [string-input input-type]
+  (if (or (= input-type "int")
```
breilly2: What types do we not want to parse as JSON?

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -22,6 +22,25 @@
...
+          (string/includes? input-type "array")
+          (= input-type "string"))
+    ; if conversion to json is unsuccessful (as a result of invalid json), just return string
+    (try (utils/parse-json-string string-input)
+         (catch js/Error e string-input))
```
helgridly: I think what you're after is behaviour looks like this (thing in box -> json value):

`42` -> `42` (ints are json)
`this.foo` -> `"this.foo"` (expressions aren't in quotes and are thus invalid json, so you have to represent them as strings)
`"bees"` -> `"\"bees\""` (things in quotes are valid json, but you have to differentiate them from expressions, so you have to escape the quotes)
`[1,"2"]` ->`[1,"2"]` (arrays are valid json (even with mixed element types))
`[1,"2"` -> `"[1,\"2\""` (not a valid array, represent as string and escape quotes)

So that's:

* represent json as json where possible
* represent expressions as strings
* string literals end up with escaped quotes (to differentiate from expressions)
* invalid json ends up in a string with escaped quotes

helgridly: In Scala I'd do this (I think):

```
//input is the full context of the text box, including any quotes the user typed
input.parseJson match {
	case Success(JsString(str)) => JsString(input.escapeme())
	case Success(JsValue(other)) => JsValue(other)
	case Failure(_) => JsString(input.escapeme())
}
```

src/cljs/main/broadfcui/common/method/config_io.cljs
```diff
@@ -22,6 +22,25 @@
 
 (def clip (partial merge table-style/clip-text))
 
+; ask what other types it could be. Maps?
+(defn get-typed [string-input input-type]
+  (if (or (= input-type "int")
+          (string/includes? input-type "array")
```
breilly2: `string/starts-with?` should work and not match things that it shouldn't (e.g. `Map[String,Array]`).

src/cljs/main/broadfcui/common/method/config_io.cljs
breilly2: I just did some manual testing and `String`, `Int` and `Array[String]` work. Here are examples of types that weren't rendered as I expected they should:
```
{
  "w.t.inFloat": "1.5",
  "w.t.inStringMap": "{\"foo\":\"bar\"}",
  "w.t.inFile": "\"gs://foo/bar\"",
  "w.t.inBoolean": "true"
}
```
All of these have extra outer quotes that shouldn't be there.

ansingh7115: @helgridly I was under the impression that someone downloading the inputs json would potentially be using it to submit directly to Cromwell, in which case we'd want `"bee"` -> `"bee"` and this._ and workspace._ would be illegal. But, reading the ticket, it says for use in other workspaces, so you're right :)

@breilly2 Yes, I will add those types as well

breilly2:
I think there should be a selenium test or two for this. I don't know why I hesitated to suggest this earlier, but I was thinking about it since yesterday and think it would be worth doing.

In addition, now that we have both upload and download, maybe we should have a test that fills in some values, downloads the JSON, then uploads the JSON and verifies that the values are the same as what was entered. This is a stated use case for these features. I'm happy to help write that test.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,96 @@
...
+      case Left(clickable) => clickable.doClick()
+      case Right(query) => {
+        logger.info(s"form: ${query.queryString}")
+        find(query).get.underlying.submit()
+      }
+    }
+
+    for {
+      hello <- Option("why does this work???")
+    } yield archiveDownloadedFile(s"$downloadPath/$fileName")
```
ansingh7115: So `archiveDownloadedFile(s"$downloadPath/$fileName")` without a for/yield, by itself causes the following error: `Expression of type String doesn't conform to expected type TypeParameterForSyntheticFunction_`
However, when its returned as part of a for comprehension, like it is now, it works fine.

I don't understand why this is!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,96 @@
+package org.broadinstitute.dsde.firecloud.fixture
```
ansingh7115: Added a test - the code to download from a link was in WorkspaceDataPage so I broke it out into `DownloadFixtures`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
helgridly: The signature for `def downloadFile` says it should return an `Option[String]`. `archiveDownloadedFile` returns a `String`, so that doesn't match.

For-comps are syntactic sugar for a series of map and flatMap, so what you're doing here is equivalent to this:

```
Option("why does this work???") map { _ =>
    archiveDownloadedFile(s"$downloadPath/$fileName")
}
```

i.e. the thing in the `yield` ends up getting "boxed" into a monad of the first type the for-comp unpacks. In this case, an `Option`, which is why it works.

In any case, if you return `Option(archiveDownloadedFile(s"$downloadPath/$fileName"))` instead of the for-comp it should work fine.

I don't know why it says things about `TypeParameterForSyntheticFunction_`; that bit looks like [some IntelliJ specific thing](https://github.com/scalatest/intellij-scala-old/blob/master/src/org/jetbrains/plugins/scala/util/ScalaUtils.java#L53). It's _possible_ you might get a different (and more comprehensible) compile error if you run scalac directly here.

ansingh7115: Thanks! I'm removing Option altogether and just returning String

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -12,10 +12,56 @@ import org.broadinstitute.dsde.workbench.service.test.WebBrowserSpec
...
+
+  val wdlInputs = ListMap(
+    "w.t.inString" -> "\"test\"",
+    "w.t.inFloat"-> "1.5",
+    "w.t.inInt" -> "2",
+    "w.t.inBoolean" -> "true",
+    "w.t.inFile" -> "\"gs://foo/bar\"",
+    "w.t.inStringArray" -> """["foo","bar"]""",
+    "w.t.inStringArray2" -> """["say \"hi\"!"]""",
+    "w.t.inStringMap" -> """{"foo":"bar"}"""
```
ansingh7115, line 66: whoops, forgot to test `this._` and `workspace._`. I'll add that

ansingh7115:
jenkins retest

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,94 @@
...
+  def makeTempDownloadDirectory(): String = {
+    /*
+     * This might work some day if docker permissions get straightened out... or it might not be
+     * needed. For now, we instead `chmod 777` the directory in run-tests.sh.
+    new File("chrome").mkdirs()
+    val downloadPath = Files.createTempDirectory(Paths.get("chrome"), "downloads")
+    val permissions = Set(PosixFilePermission.OWNER_WRITE, PosixFilePermission.GROUP_WRITE, PosixFilePermission.OTHERS_WRITE)
+    Files.setPosixFilePermissions(downloadPath, permissions.asJava)
+    downloadPath.toString
+     */
```
helgridly: does this comment block still need to be here?

helgridly (approved):
you should get a real UI person to review this because honestly i don't have much of a clue

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
breilly2: This was a huge pain to figure out and the way that it currently works is an ugly hack that is in fiab-start... far away from here where it matters... so I'd much rather be able to do something here. There might be a way to get it to work but I couldn't afford the time at the time.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,94 @@
...
+
+import org.broadinstitute.dsde.firecloud.component.Clickable
+import org.broadinstitute.dsde.workbench.service.util.Util
+import org.scalatest.concurrent.Eventually
+import java.util.UUID
+
+import org.broadinstitute.dsde.workbench.service.test.WebBrowserUtil
+import org.openqa.selenium.WebDriver
+
+trait DownloadFixtures extends Eventually with LazyLogging with WebBrowserUtil {
```
breilly2: I think this should be called something other than `DownloadFixtures`. Now that I think about it, our other "fixture" classes should maybe have a different naming scheme. They contain functions for creating fixtures, not fixtures themselves. Still, the functions here seem more like utility functions. I'm not sure I love `DownloadUtil` as an alternative but it's what I can think of right now. Feel free to offer other suggestions.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,94 @@
...
+    * into the "downloads" directory off the current working directory. This serves two purposes:
+    *
+    * 1. Archiving the file for later inspection when tests fail
+    * 2. Keeping the browser download directory clean so that it doesn't auto-rename subsequent
+    * downloads with the same filename
+    *
+    * @param downloadPath the directory where the browser saves downloaded files
+    * @return the relative path to the moved download file, or None if downloadPath was not given
+    */
+  def downloadFile(downloadPath: String, fileName: String, downloadThing: Either[Clickable, CssSelectorQuery])(implicit webDriver: WebDriver): String = synchronized {
```
breilly2: Use of `Either` feels a little odd here. Probably not worth establishing an ADT for this... maybe a method overload based on type? That way callers don't have to worry about saying `Left` or `Right`.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,94 @@
...
+
+    downloadThing match {
+      case Left(clickable) => clickable.doClick()
+      case Right(query) => {
+        logger.info(s"form: ${query.queryString}")
+        find(query).get.underlying.submit()
+      }
+    }
+
+    archiveDownloadedFile(s"$downloadPath/$fileName")
```
breilly2: Yes, I like this better because the caller, who cares about the filename, is now in control of the filename. 👍

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -12,10 +12,73 @@ import org.broadinstitute.dsde.workbench.service.test.WebBrowserSpec
...
+  )
+
+  val refInputs = ListMap(
+    "w.t.inWorkspaceRef" -> """workspace.hello""",
```
breilly2: Don't need to use `"""` here. It's used above to avoid having to escape embedded `"` (or, in the case of `"""["say \"hi\"!"]"""`, to make it easier to see that `\"` represents the literal string `\"`, not an escaped `"`).

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
```diff
@@ -132,34 +227,30 @@ class MethodConfigSpec extends FreeSpec with Matchers with WebBrowserSpec with W
...
+
+              // Populate input fields from a json containing all the field values and some fields that don't exist
+              configPage.populateInputsFromJson(generateInputsJson(inputs ++ unmatchedVariables))
 
+              // We should have been automatically switched to edit mode
               configPage.isEditing shouldBe true
-              variables foreach {
+
+              // for each input, (not included the fields that didn't exist), we should have the correct value
+              // for referential inputs ($this._ or $workspace._), it should have removed the "$"
```
breilly2, line 248: Except it looks like you modified `generateInputsJson` to remove the leading "$" so there's nothing left for FC to remove.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/DownloadFixtures.scala
```diff
@@ -0,0 +1,94 @@
...
+    *
+    * If downloadPath is given, the file is given a timestamped name and moved from that location
+    * into the "downloads" directory off the current working directory. This serves two purposes:
+    *
+    * 1. Archiving the file for later inspection when tests fail
+    * 2. Keeping the browser download directory clean so that it doesn't auto-rename subsequent
+    * downloads with the same filename
+    *
+    * @param downloadPath the directory where the browser saves downloaded files
+    * @return the relative path to the moved download file, or None if downloadPath was not given
```
breilly2: Remove "or None if downloadPath was not given" now that `downloadPath` is not an `Option`. (I don't remember why it used to be an `Option`.)

breilly2 (commented):
Done reviewing for now. A couple of minor things and some questions.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/analysis/MethodConfigSpec.scala
ansingh7115: whoops

ansingh7115:
jenkins retest

ansingh7115:
jenkins retest

breilly2 (approved):
:+1: Looks good!

ansingh7115:
Tests pass except for 1 which is related to signing in and out through google. This is an unrelated failure that is also happening on other PRs and I received the go-ahead from QA to merge regardless of that test failure.

PR #1362, opened 2018-07-17
use CSS pseudo-elements instead of text for LTR marks

davidangb:
See also:
* #1267
* https://broadinstitute.atlassian.net/browse/GAWB-2495
* https://broadinstitute.atlassian.net/browse/GAWB-1912
* https://broadinstitute.atlassian.net/browse/GAWB-1216
and more ... 

The solution to the combination of (left-ellipses AND (trailing punctuation OR numeric filenames)) in #1267 was to pad the string value at both start and end with the LTR mark (unicode [200E](https://www.compart.com/en/unicode/U+200E)). However, this broke copying, because the LTR mark was included when selecting the text and then copying it.

In this PR, I change how we pad the string value - I'm using the CSS `::before` and `::after` pseudo-elements instead of directly changing the text. Since pseudo-elements are not selectable/copyable, this fixes the copy issue.

databiosphere/firecloud-app#110 has a list of test cases to check.

### DIFF
--- a/src/static/index.html
+++ b/src/static/index.html
@@ -26,6 +26,9 @@
       *:disabled {
         cursor: not-allowed;
       }
+      .ltr-marked::before, .ltr-marked::after {
+        content: "\200E";
+      }
     </style>
     <link rel="icon" type="image/x-icon" href="assets/favicon.ico">
 </head>
--- a/src/cljs/main/broadfcui/common/gcs_file_preview.cljs
+++ b/src/cljs/main/broadfcui/common/gcs_file_preview.cljs
@@ -201,12 +201,6 @@
                                :workspace-namespace (:workspace-namespace props)
                                :dismiss (:dismiss props))])))]))})
 
-;; Sometimes we apply an RTL rule so that long links overflow and show ellipses on the left-hand side.
-;; Go back to LTR here so we do not reorder the object name. Both the leading and trailing instances
-;; are necessary to cover all cases. (GAWB-2495, GAWB-1912)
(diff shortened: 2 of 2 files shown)


src/static/index.html
```diff
@@ -26,6 +26,9 @@
       *:disabled {
         cursor: not-allowed;
       }
+      .ltr-marked::before, .ltr-marked::after {
+        content: "\200E";
+      }
```
davidangb, line 31: CSS psuedo-elements can't be specified inline, so I'm adding them here.

davidangb:
one unrelated failure in https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/7830/. I will re-run this spec manually.
```
[info] *** 1 TEST FAILED ***
[error] Failed: Total 124, Failed 1, Errors 0, Passed 123, Ignored 4
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.billing.BillingSpec

[info]     - should be able add a new user, create a workspace, and run a method, change billing account *** FAILED *** (11 minutes, 42 seconds)
[info]       Creating billing project: billing-spec-create-mfxsqfv "[failure]" did not equal "[success]" (BillingSpec.scala:146)
```

update: `BillingSpec` passed in the manual run: https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/7836/

ghost (approved):
Very nice! Successfully tested all cases described in firecloud-app#110.

PR #1394, opened 2018-08-28
TOS gate during login

davidangb:
~DO NOT MERGE. This is incomplete; I'm making the PR as a means of handing off work in progress.~

I have plugged in a call to query the TOS API during the login process. If the current user has already accepted the TOS, all is well and login continues. If the user hasn't responded yet or has actively declined the TOS, login halts and the user should not be able to access the app.

~Still a lot left to do. Here's my list, but there may be more I haven't thought of:~

- [x] move the `appid` and `tosversion` values to config instead of hardcoding in `TermsOfService` component.
- [x] move the API's URL to config instead of hardcoding in `TermsOfService` component.
- [x] move the ajax call into `endpoints.cljs` instead of hardcoding in `TermsOfService` component. 
- [x] potentially, eliminate the duplication between the `UserStatus` and `TermsOfService` components. The `TermsOfService` component may change enough that the duplication isn't worth fixing.
- [x] display the actual terms of service content if the user hasn't already accepted
- [x] require the user to scroll to the bottom of the terms of service content before allowing the user to accept (check with PO: is this required?) (@zarsky-broad sez: checked with Ilyana, this is required. Going to get a markdown version of the ToS and put it in the alert bucket, generally copying the pattern from the trial.)
- [x] provide an action (button, link, etc) to allow the user to accept the terms of service. On acceptance, fire a POST call to the TOS API, handling errors.
- [x] update selenium tests. I fully expect massive test failures on this code, since all our test users will be blocked from logging in.

Twinned with ~https://github.com/broadinstitute/firecloud-develop/pull/1375~ https://github.com/broadinstitute/firecloud-develop/pull/1372


zarsky-broad:
(Just documenting my thought process here.)
### Unfulfilled requirements
The full text should be on the page, so users have to scroll to the bottom to find the accept/decline buttons.

### Proposed next steps
ToS in markdown format, in a JSON file, keyed by version number in the alerts bucket so multiple versions could be live simultaneously, & it's super specific.
- JSON file URL should be in config
- Would this mean needing to change where the existing Terms of Service link in the footer goes?

### Tests
@MattPutnam is going to take a look once I have everything in place.

zarsky-broad:
👍ed by @bradtaylor

zarsky-broad:
Successful test run here: https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Pipelines/job/swatomation-pipeline/10975/

src/cljs/main/broadfcui/auth.cljs
```diff
@@ -206,40 +209,96 @@
...
+          [:div {}
+           [:div {:style {:color (:state-exception style/colors) :paddingBottom "1rem"}}
+            "Error loading Terms of Service information. Please try again later."]
+           [:table {:style {:color (:text-lighter style/colors)}}
+            [:tbody {:style {}}
+             [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "What went wrong:"] [:td {} (:message error)]]
+             [:tr {} [:td {:style {:fontStyle "italic" :textAlign "right" :paddingRight "0.3rem"}} "Status code:"] [:td {} (:statusCode error)]]]]])]))
+   :component-did-mount
+   (fn [{:keys [state this]}]
+     (ajax/get-google-bucket-file "tos" #(swap! state assoc :tos (% (-> (config/tos-version) str keyword))))
```
davidangb, line 285: in which environments have you populated `tos.json`?

secondarily - did you discuss with PO the need for the tos to be content-managed in a bucket? We could save it as a resource right in the codebase, if we don't expect to change this often or have a need to change it quickly.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/UserFixtures.scala
```diff
@@ -46,11 +46,22 @@ trait UserFixtures extends CleanUp with ScaledTimeSpans with Eventually { self:
...
-    * user is not registered and returns a ready RegistrationPage.
+    * user is not registered and returns a ready TermsOfServicePage.
     */
   def withSignInNewUser(user: Credentials)
-                           (testCode: RegistrationPage => Any)(implicit webDriver: WebDriver): Unit = {
-    withSignIn(user, new RegistrationPage)(testCode)
+                       (testCode: RegistrationPage => Any)(implicit webDriver: WebDriver): Unit = {
+    withCleanUp {
+      withSignIn(user, new TermsOfServicePage) { tosPage =>
+        register cleanUp executeAsyncScript("window.rejectToS(arguments[arguments.length - 1])")
```
davidangb, line 54: for persistent environments like alpha and staging, I think this will create a constant stream of accept/reject/accept/reject responses for the user. I don't have an alternate solution right now - but do you agree this will be the result?

davidangb (commented):
2 comments inline ...

src/cljs/main/broadfcui/auth.cljs
panentheos: From @zarsky-broad : It's populated in all the environments he has access to, definitely dev and qa. If there are others, he doesn't have access so would need someone else to do it.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/UserFixtures.scala
panentheos: Correct, but from the current API this seems to be the best we can do for now. It would probably be a good idea at some point to develop a separate mechanism for bulk clearing state, for use in testing environments.

davidangb: yeah, I can add this as a TODO for follow-on work - just wanted to confirm.

src/cljs/main/broadfcui/auth.cljs
davidangb: there definitely are others - for instance, prod; we wouldn't want this to go live without populating in prod. I can do this later this week; must hold off merging (well, releasing) until it's done.

davidangb: update: tos.json now exists in all envs.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/UserFixtures.scala
ahaessly: i don't think we need withCleanUp here. this trait extends CleanUp so we should be able to just call register cleanup...

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/user/RegistrationSpec.scala
```diff
@@ -44,30 +44,33 @@ class RegistrationSpec extends FreeSpec with BeforeAndAfter with Matchers with W
...
-
-          new DataLibraryPage().validateLocation()
-
-          val profilePage = new ProfilePage().open
-          val username = testUser.email.split("@").head
-          /* Re-enable this code and remove the temporary code below after fixing rawls for GAWB-2933
-          profilePage.readProxyGroupEmail should (startWith (username) and endWith ("firecloud.org"))
-          */
-          profilePage.readProxyGroupEmail should endWith("firecloud.org")
+        withCleanUp {
```
ahaessly, line 47: same comment as above - we should not need the withCleanUp here

ahaessly (dismissed):
looks good. just some extra withCleanUp blocks

katevoss:
Jenkins retest

katevoss:
Jenkins retest

katevoss:
Jenkins retest

katevoss:
Jenkins retest

PR #1403, opened 2018-09-07
third-party dependency updates + accompanying test updates

davidangb:
This PR has a few parts:

1. Update `webpack` and `css-loader` third-party dependencies to latest versions. These have transitive dependencies that our automated scans proactively identified as having security vulns.

2. The migration from webpack 3 to 4 required other dependency changes: a `node-sass` update and the inclusion of `webpack-cli`. Additionally, the webpack 3->4 update required changes to the webpack config (see https://webpack.js.org/migrate/4/).

3. Post-upgrade, the code had mysterious test failures in `DataUseAwareSearchSpec` and `MethodConfigSpec`. These failures happened only in Jenkins; they didn't happen when running tests locally from IntelliJ using local Chrome. I assume this is a interaction between headless tests and some library upgrade. I tracked this down to the autosuggest Javascript using the `aria-controls` attribute instead of `aria-owns` - though local IntelliJ/Chrome runs still used `aria-owns`. I updated the test code to be resilient about using either attribute, and centralized the test code so it's shared between the two failing tests. (I didn't try running headless locally. Since we want to support both headless and headful, I felt good about just adding the resilient code.)

Don't try to review commit-by-commit. You'll see all my interim commits when I was debugging what was going on in Jenkins.

Finally, reviewer, remember that `package-lock.json` is an automatically-generated file. You probably don't want to review that one.

-----

- [ ] **Submitter**: Include the JIRA issue number in the PR description
- [ ] **Submitter**: Check documentation and code comments. Add explanatory PR comments if helpful.
- [ ] **Submitter**: If you changed a URL that is used elsewhere (e.g. in an email), comment about where it is used and ensure the dependent code is updated.
- [ ] **Submitter**: JIRA ticket checks:
  * Acceptance criteria exists and is met
  * Note any changes to implementation from the description
  * To Demo flag is set
  * Release Summary is filled out, if applicable
  * Add notes on how to QA
- [ ] **Submitter**: Update RC_XXX release ticket with any config or environment changes necessary
- [ ] **Submitter**: Update FISMA documentation if changes to:
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
- [ ] **Submitter**: If you're adding new libraries, sign us up to security updates for them
- [ ] **Submitter**: If you're adding new automated UI tests, review the test plan with QA
* Review cycle:
  * LR reviews
  * Rest of team may comment on PR at will
  * **LR assigns to submitter** for feedback fixes
  * Submitter rebases to develop again if necessary
[...]

PR #1407, opened 2018-09-11
Revert "third-party dependency updates + accompanying test updates"

davidangb:
Reverts broadinstitute/firecloud-ui#1403 due to problems with IGV.


jmthibault79:
Solves the immediate IGV problem.

PR #1409, opened 2018-09-12
add git revision to UI html source [risk: medium]

davidangb:
I got frustrated always wondering which version of the app was running in which environment.

This PR adds git revision/hash to the HTML source as a meta tag. Developers/admins/other interested parties can just view-source on the FireCloud UI to know which specific commit is in use.

The `try/catch` in webpack.config.js should make this code resilient to environments where git information isn't available - notably, running the UI via `./config/docker-rsync-local-ui.sh`. In that local environment, we rsync codebases into a docker volume and don't sync the `.git` directory. But in Jenkins builds, this information should be available; I verified by starting a custom fiab and saw source of:
```

<html>
  <head profile="http://www.w3.org/2005/10/profile">
  <meta charset="utf-8">
  <meta http-equiv="Content-type" content="text/html;charset=utf-8">
  <meta name="viewport" content="initial-scale=1,width=device-width">
  <meta name="application-name" content="FireCloud"
    data-revision="fisma-r1-1455-g734c37a0"
    data-commithash="734c37a03a7752cc5494dab2acd7e7f149e1ad58">
  <title>FireCloud | Broad Institute</title>
```

I'm very conservatively labeling this `[risk:medium]` instead of low. Potentially, an attacker could use the revision information to see which bugs/vulnerabilities have been fixed since the commit in use. However, since our github repo is open-source anyway, this doesn't actually give the attacker any new information - it just makes it slightly easier to know which code is in use when.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2665,6 +2665,12 @@
         "assert-plus": "1.0.0"
       }
     },
+    "git-revision-webpack-plugin": {
+      "version": "3.0.3",
+      "resolved": "https://registry.npmjs.org/git-revision-webpack-plugin/-/git-revision-webpack-plugin-3.0.3.tgz",
+      "integrity": "sha512-B2ixM0fY7VgR61ZRSXYrh0R57Er7RY+CZb+fja5OFe21Y5o9GgzQanMgdlcBwWZ+LoOVqxBogbDutTTYMXQDWw==",
+      "dev": true
+    },
     "github-markdown-css": {
       "version": "2.10.0",
       "resolved": "https://registry.npmjs.org/github-markdown-css/-/github-markdown-css-2.10.0.tgz",
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "css-loader": "^0.28.9",
     "extract-text-webpack-plugin": "^3.0.2",
     "file-loader": "^1.1.6",
+    "git-revision-webpack-plugin": "^3.0.3",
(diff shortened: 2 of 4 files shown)


adrazhi (commented):
Hi @davidangb, your logic makes a lot of sense in that even if this introduces an attack surface at some level, it does not add much due to the fact that our repo is open to the world.  

Is there anything else associated with this that's likely to change in the future?

davidangb:
@adrazhi I don't have any immediate plans to make changes to this feature, though of course something could change in the future. If you're concerned about something specific, could you explain?

adrazhi:
@davidangb Not thinking of anything specific, but I guess whatever comes up in the future is worth another discussion. Trying to make sure that the info presented to a potential attacker doesn't go beyond the attacker knowing which bug/vulnerability has been fixed. (vs knowing how to exploit for example).

davidangb:
@adrazhi sounds good, and I'll assume the same answer for broadinstitute/firecloud-orchestration#673 too. If I were to make changes to this feature, I'd submit them for another security review.

If you are satisfied, could I get a thumb from you on the PR? (I will seek out other reviewer for code/syntax)

dmohs (approved):
Love this. It's super easy to find when all you have to do is inspect the source.

jmthibault79 (approved):
Looks reasonable and does not break Local UI

adrazhi (approved):

PR #1411, opened 2018-09-13
dependency updates + IGV 2.x + compatibility/test updates [risk: low]

davidangb:
followup to #1403. That PR addressed upgrades of transitive dependencies, automation test updates and other compatibility updates, but it broke IGV. IGV broke because, for some reason, it couldn't find its jquery-ui/draggable dependency. I reverted #1403, so this PR includes all of #1403 as well as additional work to fix IGV. Specifically, this PR includes:

1. Update `webpack` and `css-loader` third-party dependencies to latest versions. These have transitive dependencies that our automated scans proactively identified as having security vulns.

2. The migration from webpack 3 to 4 required other dependency changes: a `node-sass` update and the inclusion of `webpack-cli`. Additionally, the webpack 3->4 update required changes to the webpack config (see https://webpack.js.org/migrate/4/).

3. Post-upgrade, the code had mysterious test failures in `DataUseAwareSearchSpec` and `MethodConfigSpec`. These failures happened only in Jenkins; they didn't happen when running tests locally from IntelliJ using local Chrome. I assume this is a interaction between headless tests and some library upgrade. I tracked this down to the autosuggest Javascript using the `aria-controls` attribute instead of `aria-owns` - though local IntelliJ/Chrome runs still used `aria-owns`. I updated the test code to be resilient about using either attribute, and centralized the test code so it's shared between the two failing tests. (I didn't try running headless locally. Since we want to support both headless and headful, I felt good about just adding the resilient code.)

4. Upgrade IGV from 1.0.6 to 2.0.0-rc5. The big advantage of IGV 2.x for this PR is that IGV 2.x is dependency-free. This works around the jquery-ui/draggable problem as well as allowing us to remove jquery-ui from our dependency tree entirely, allowing us to simplify how we load IGV, and removing our IGV patches. And of course we get the UX/user benefit of the latest IGV!

Don't try to review commit-by-commit. You'll see all my interim commits when I was debugging what was going on in Jenkins.

Finally, reviewer, remember that `package-lock.json` is an automatically-generated file. You probably don't want to review that one.


package.json
```diff
@@ -22,16 +21,15 @@
...
-    "node-sass": "4.6.1",
+    "node-sass": "^4.9.3",
     "rebuild-node-sass": "^1.1.0",
     "sass-loader": "^6.0.6",
     "style-loader": "^0.19.1",
-    "uglifyjs-webpack-plugin": "^1.1.6",
     "url-loader": "^0.6.2",
-    "webpack": "^3.10.0"
+    "webpack": "^4.18.0",
+    "webpack-cli": "^3.1.0"
```
davidangb, line 33: remove `jquery-ui` entirely; IGV was the only thing depending on it.
remove `extract-text-webpack-plugin`. This was unused, and webpack 4 threw warnings about it being old.
remove `uglifyjs-webpack-plugin`. This is now a default inside of webpack 4.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
```diff
@@ -10,61 +11,58 @@
    [broadfcui.utils.user :as user]
    ))
 
-(defonce ^:private igv-styles-loaded? (atom false))
-
-(defn- load-igv-styles []
-  (.. (js/$ "head") ; IGV uses jQuery, so I'm using it here too.
-      (append "<link rel=\"stylesheet\" type=\"text/css\" href=\"https://igv.org/web/release/1.0.1/igv-1.0.1.css\">"))
-  (reset! igv-styles-loaded? true))
-
```
davidangb, line 19: IGV 2.x is dependency-free; no need to load CSS.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
```diff
@@ -10,61 +11,58 @@
...
 (defn- options [tracks token]
   (clj->js
-   {:genome "hg19"
-    :trackDefaults {:palette ["#00A0B0" "#6A4A3C" "#CC333F" "#EB6841"]
-                    :bam {:coverageThreshold 0.2
-                          :coverageQualityWeight true}}
```
davidangb, line 25: I've removed as many of our overrides as possible, so we are using all the new/improved IGV defaults.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
```diff
@@ -10,61 +11,58 @@
...
+   {
+    ;; TODO: should we set hg38 as the default?
+    :genome "hg19"
+    :oauthToken token
```
davidangb, line 19: IGV 2.x supports setting an oauth token at the root of IGV instead of per-track. This would even allow us to use tracks pointing at GA4GH variant stores or htsget to google servers in the future.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
```diff
@@ -10,61 +11,58 @@
...
-                 {:key "igv-deps"
-                  :on-error #(swap! state assoc :error? true)
-                  :on-load #(swap! state assoc :deps-loaded? true)
-                  :path "igv-deps.bundle.js"}])]))
+                  {:key "igv"
+                   :on-error #(swap! state assoc :error? true)
+                   :on-load #(do
+                               (swap! state assoc :igv-loaded? true)
+                               (this :refresh))
+                   :path "https://igv.org/web/release/2.0.0-rc5/dist/igv.min.js"}])]))
```
davidangb, line 56: IGV 2.x does support installation via npm. We could do that instead of loading via `ScriptLoader`. I didn't attempt this because:
1) I wanted to keep this PR as small as possible (it's already quite large)
2) I still like only loading the igv js once a user navigates to the relevant tab.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
```diff
@@ -10,61 +11,58 @@
...
    :component-did-update
    (fn [{:keys [props state prev-props this]}]
      (when (and (not= (:tracks props) (:tracks prev-props)) (:igv-loaded? @state))
        (this :refresh)))
    :refresh
    (fn [{:keys [props refs]}]
+     ;; .createBrowser returns a Promise that we should be using to add/remove tracks. For now, we do it brute-force:
+     ;; empty the container, then create a new IGV browser each time we change tracks.
+     (set! (.-innerHTML (@refs "container")) "")
```
davidangb, line 65: I'd consider fixing this in a future PR; I don't want to do it here, to keep changes small.

davidangb:
@aweng98 - you had concerns about the automation test approach for autosuggest. Could you review at minimum that part?
@zarsky-broad - you reviewed the first webpack change in #1403 and you looked into the IGV stuff. Could you give this a review?
@jmthibault79 - hello fellow FCApp team member! Could you also give this a review?

aweng98:
Would you consider something retry to handle `StaleElementException` like something here:
```
def getSuggestions(implicit webDriver: WebDriver): Seq[String] = {
    val wait = new WebDriverWait(webDriver, 10)
    wait until new java.util.function.Function[WebDriver, Seq[String]] {
      override def apply(d: WebDriver): Seq[String] = {
        try {
          texts // this is a func that gets all text in dropdown. you can remove handling of stale element exception from this func. 
        } catch {
          case e: StaleElementReferenceException =>
            logger.warn("getSuggestions",e) // log stacktrace to help troubleshooting in case wait time exhausted
            throw e
        }
      }
    }
  }
```

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Suggests.scala
```diff
@@ -0,0 +1,68 @@
...
+import com.typesafe.scalalogging.LazyLogging
+import org.openqa.selenium.{StaleElementReferenceException, WebDriver}
+
+import scala.util.{Failure, Success, Try}
+
+/**
+  * Mix in for Components (ex. SearchField or TextField) which supply dropdown autosuggestions
+  * based on user input.
+  */
+trait Suggests extends LazyLogging { this: Component =>
```
aweng98, line 12: I really like how you extracted this code and made it a `trait`. 👍

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Suggests.scala
```diff
@@ -0,0 +1,68 @@
...
+    // wait for dropdown to contain at least one option
+    val listOptionXpath = s"//div[@id='$dropdownId']/ul[@role='listbox']/li[@role='option']"
+    await condition {
+      find(xpath(s"//div[@id='$dropdownId']")).exists(_.isDisplayed) &&
+      findAll(xpath(listOptionXpath)).map(_.text).toSeq.nonEmpty // getting Element's text force screen scroll if item is outside of viewport
+    }
+
+    // return the value of the options text
+    // retry on StaleElementReferenceException
+    Try[Seq[String]] {
```
aweng98, line 58: this wasn't working as I found out few days ago. still finding stale exception. thus I suggested the retry with wait.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Suggests.scala
```diff
@@ -0,0 +1,68 @@
...
+    // the  autocomplete JavaScript will add either an aria-owns or an aria-controls attribute to the input element.
+    // It's unclear which attribute is used under which conditions by the third-party autocomplete JavaScript,
+    // so we look for either here.
+    //
+    // the value of this attribute will be the id of the DOM element (a div) that contains the suggestion elements.
+    // see https://www.w3.org/WAI/PF/aria-1.1/states_and_properties#attrs_relationships for info on aria-owns/aria-controls.
+    //
+    // wait for the aria-owns/aria-controls attribute to exist:
+    await condition {
+      Option(uel.getAttribute("aria-owns")).nonEmpty ||
```
aweng98, line 35: this is good addition!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/analysis/WorkspaceAnalysisPage.scala
```diff
@@ -0,0 +1,36 @@
...
+import org.broadinstitute.dsde.workbench.model.google.GcsPath
+import org.broadinstitute.dsde.workbench.service.test.WebBrowserUtil
+import org.openqa.selenium.WebDriver
+import org.scalatest.selenium.{Page, WebBrowser}
+
+class WorkspaceAnalysisPage(namespace: String, name: String)(implicit webDriver: WebDriver)
+  extends WorkspacePage(namespace, name) with Page with PageUtil[WorkspaceAnalysisPage] with LazyLogging {
+
+  override def awaitReady(): Unit = {
+    await condition igvContainerSelector.findElement.exists(_.isDisplayed)
```
aweng98, line 18: this might be problematic because `.isDisplayed` is not a reliable method to use to determine page finished rendering. I don't know this page well. Is there any additional page objects that should be checked in `awaitReady`? It's okay if there isn't anything else. we can wait and find out in time.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
zarsky-broad: Why not install it via npm and load it through the igvDeps bundle? Also, I think that was the only place timingDiagram.js was loaded...?

src/cljs/main/broadfcui/page/workspace/analysis/track_selector.cljs
```diff
@@ -17,6 +17,13 @@
 
 
 (def ^:private supported-file-types [".bam" ".vcf" ".bed"])
+;; IGV.js supports:
```
jmthibault79: Why does this list differ from the above line and the `case file-extension` block from igv.cljs?

jmthibault79:
IGV works when testing locally.  One question about supported file types.

zarsky-broad (approved):
One thought on a simplification. I'm happy otherwise!

src/cljs/main/broadfcui/page/workspace/analysis/track_selector.cljs
davidangb: This PR maintains the list of supported file types - it neither adds nor removes any. Way back when, we limited the supported file types to reduce scope of testing and troubleshooting. This comment is my - I guess passive-aggressive - way of reminding ourselves that we could improve on this. However, any change to supported file types would require code changes in `find-index-file` and potentially elsewhere - as well as testing - so I'd rather not do it as part of this PR.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
davidangb: timingDiagram.js is loaded via base-deps.js: https://github.com/broadinstitute/firecloud-ui/blob/da_dependencyUpdatesAndIGV2/src/js/base-deps.js#L13, which makes WAY more sense to me - since timingDiagram.js handles display of workflow timings, I don't want it bundled with IGV.

As I said above, I'm open to trying to install IGV via npm - I just don't want to scope-creep this PR any further. Would you consider that for a follow-on PR?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/page/workspaces/analysis/WorkspaceAnalysisPage.scala
davidangb: What specifically is problematic about `.isDisplayed`? Are there specific use cases that cause problems? The only other element we could test for is the "Select Tracks..." button.

src/cljs/main/broadfcui/page/workspace/analysis/igv.cljs
zarsky-broad: > base-deps.js

No argument here, I just wanted to make sure it was somewhere, I'm not sure why it was in igv to start with.

> follow-on PR

Sure!

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Suggests.scala
davidangb: @aweng98 do you feel the retry/wait needs to be a part of this PR? This PR is already super big and complicated - I hesitate to add more to it, especially something that could take a few iterations to stabilize. Would you be ok with this PR as-is, which (hopefully) neither improves nor hampers the tests, and have a follow-on PR for retries?

src/cljs/main/broadfcui/page/workspace/analysis/track_selector.cljs
jmthibault79: Makes sense.  Then I would like a small comment update.  Something like "FireCloud currently supports the above subset of IGV file types.  The full list is..."

automation/src/test/scala/org/broadinstitute/dsde/firecloud/component/Suggests.scala
aweng98: not a problem if retry isn't in this PR. A new PR afterward is fine.

aweng98 (approved):

davidangb:
Rebased to develop, so it looks like there are recent commits. The only actual new change is 48cc757 - a comment update for @jmthibault79 .

src/cljs/main/broadfcui/page/workspace/analysis/track_selector.cljs
jmthibault79: 👍to update

jmthibault79 (approved):

davidangb:
after rebase, one failure in automation tests:
```
[info] *** 1 TEST FAILED ***
[error] Failed: Total 123, Failed 1, Errors 0, Passed 122, Ignored 7
[error] Failed tests:
[error] 	org.broadinstitute.dsde.firecloud.test.library.DataLibrarySpec
[error] (Test / testOnly) sbt.TestsFailedException: Tests unsuccessful

[info] - Input multiple tags in filter field *** FAILED *** (1 minute, 28 seconds)
[info]   org.broadinstitute.dsde.workbench.service.RestException: {
[info]   "statusCode": 504,
[info]   "source": "FireCloud",
[info]   "timestamp": 1536955007224,
[info]   "causes": [],
[info]   "stackTrace": [],
[info]   "message": "Request Timed Out"
[info] }
```

I will re-run this single test manually ...

aweng98:
fyi. ran PR against all services. It passed. 
https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/9286/

One failed test in https://fc-jenkins.dsp-techops.broadinstitute.org/job/swatomation-pipeline/9285/ is due to hitting Google quote limit.

davidangb:
passed in my manual re-run too: https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Pipelines/job/swatomation-pipeline/9291/

PR #1495, opened 2019-01-08
NPS survey,  redux

davidangb:
DataBiosphere/firecloud-app#239

Requirements from PO:
* run a brand new NPS survey, asking for feedback from everyone, even if they already responded/declined the first one
* show the NPS survey as an overlay/popup, instead of the gentler icon in bottom-right, in an attempt to get greater participation

My implementation:
*  update `npssurvey.js` based on the new survey created in SurveyMonkey
* create a conditional in the code to show the NPS survey as a popup vs.  as a comment icon, based on a variable
* set the variable to show NPS as popup
* add a version flag to our local-storage persistence for NPS survey, so that if we switch back to the comment icon, we'll show the icon to users who previously answered/dismissed the first survey

-----

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/cljs/main/broadfcui/main.cljs
```diff
@@ -43,8 +43,19 @@
 
 (injections/setup)
 
+;; - when true, will show NPS survey as a popup. Popup visibility is controlled by SurveyMonkey via cookies.
+;; - when false, will show a comment icon in the bottom right of the page; clicking on the icon shows the NPS popup.
+;;   Popup visibility is controlled by our own persistence in local storage.
+;;
+;; Switching this var from true to false can lead to undesirable user behavior. If the user has already dismissed the
```
breilly2, line 50: Would it help to merge the concepts of persistence version and popup trigger into a single "NPS campaign" concept? My thought is that then switching from popup to icon would be a new campaign with a different set of cookie/local storage tracking data. I don't know if that would mean a new SurveyMonkey campaign and whether that would be good or bad.

davidangb: it *would* require a new campaign on the SM side - with corresponding ID change in `npssurvey.js`. They feel orthogonal to me - we could switch campaign+version without changing the  `popup?` value,  and we could the `popup?` value from false to true  without changing the campaign+version. The only potentially tricky change is moving `popup?` from true to  false.

breilly2: I think I get it now. We're leaving it up to SM to manage whether or not to show the popup rather than keeping anything in local storage... probably because we don't have any hooks into whether or not they responded to or dismissed the popup.

It's a little confusing, though, because it seems like even in the icon case we're still relying on SM making the decision about whether or not to show the survey... so there's duplicated state between SM and our local storage tracking. Perhaps this is unavoidable.

davidangb: the duplicated state in icon mode is exactly the icky part that causes potential problem. Our local state controls whether or not we show the icon. Once the user clicks on the icon, we defer to surveymonkey, who may decide to not show  the popup ... thus the undesirable UX of clicking on something and nothing happening. (not quite nothing: the icon goes away).

This is not a problem in popup mode. In popup mode, you either see the popup or you don't; there's no extra layer of state.

breilly2 (approved):
👍 Nicely done! I do wish that we had better insight into the surveymonkey cookies to avoid duplicated state, but I can't think there's any solution we could put in place now that we couldn't put in place later if we come up with some clever solution.

PR #1521, opened 2019-04-01
Terra redirect for workspace summary page [risk: low]

davidangb:
* add logic to redirect to Terra, based on the current location in FireCloud
* add config setting to turn redirects on/off - defaulting to off, since redirects would break tests right now
* add a `?fcredir=1` query param to the redirect, to allow for future analytics tracking
* add a Terra redirect for workspace summary, as a proof of the logic

This PR wants - but does not require -  the corresponding `firecloud-develop` PR at broadinstitute/firecloud-develop#1591

work remaining:
* all additional contextual redirects. I'd like to do these in future PRs to keep things small.

-----

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/main.cljs
+++ b/src/cljs/main/broadfcui/main.cljs
@@ -200,6 +200,15 @@
                (aget e "filename")]]
     :show-cancel? false :dismiss dismiss :ok-button "OK"}])
 
+(react/defc- TerraRedirect
+  {:render
+   (fn []
+     [:div {} "Redirecting to Terra..."])
+   :component-did-mount
+    (fn [{:keys [props]}]
+      (let [{:keys [terra-redirect make-props]} props]
+        (js-invoke (aget js/window "location") "replace" (str (config/terra-url) "/?fcredir=1#" (terra-redirect (make-props))))))})
+
 (react/defc- App
   {:handle-hash-change
    (fn [{:keys [state]}]
@@ -222,15 +231,11 @@
    (fn [{:keys [state]}]
      (let [{:keys [auth2 user-status window-hash config-loaded?]} @state
            {:keys [component make-props public? terra-redirect]} (nav/find-path-handler window-hash)
+           terra-redirect? (and config-loaded? (config/terra-redirects-enabled) terra-redirect)
            sign-in-hidden? (or (nil? component)
                                public?
                                (contains? (:user-status @state) :signed-in))]
        [:div {}
-        ;; if Terra redirects are globally enabled via config;
-        ;;   and the current path handler has a Terra redirect,
-        ;; then redirect to Terra.
-        (when (and config-loaded? (config/terra-redirects-enabled) terra-redirect)
-          (js-invoke (aget js/window "location") "replace" (str (config/terra-url) "/?fcredir=1#" (terra-redirect (make-props)))))
         (when (contains? user-status :signed-in)
           [:div {}
             [notifications/TerraBanner]
@@ -267,7 +272,6 @@
                                                       :refresh-token-saved))))}]
            (when (and config-loaded? (not auth2))
              [auth/GoogleAuthLibLoader {:on-loaded #(swap! state assoc :auth2 %)}])
-
            (cond
              (not config-loaded?)
              [config-loader/Component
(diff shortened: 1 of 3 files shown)


src/cljs/main/broadfcui/main.cljs
```diff
@@ -204,6 +204,15 @@
...
      (let [window-hash (aget js/window "location" "hash")]
+       ;; if Terra redirects are globally enabled via config;
+       ;;   and the current hash has a path handler;
+       ;;   and that path handler has a Terra redirect,
+       ;; then redirect to Terra.
+       (when (config/terra-redirects-enabled)
+         (when-let [handler (nav/find-path-handler window-hash)]
+           (let [{:keys [make-props terra-redirect]} handler]
+             (when terra-redirect
+               (js-invoke (aget js/window "location") "replace" (str (config/terra-url) (terra-redirect (make-props))))))))
```
breilly2: I like the approach.

Looks like my cljs is rusty. How does `make-props` get its arguments? In this case, it needs `namespace` and `name` in order to construct the URL.

davidangb: `make-props` gets its arguments based on the regular expression capture matches, as defined by the `regex` in each path definition, e.g. https://github.com/broadinstitute/firecloud-ui/blob/develop/src/cljs/main/broadfcui/page/workspace/details.cljs#L190. The code to calculate the regex matches is at https://github.com/broadinstitute/firecloud-ui/blob/develop/src/cljs/main/broadfcui/nav.cljs#L34.

breilly2: Ah, thanks. What I was missing is that the `make-props` attribute returned from `find-path-handler` captures the arguments; it's not the same parameterized `make-props` from the call to `nav/defpath`.

src/cljs/main/broadfcui/main.cljs
```diff
@@ -204,6 +204,15 @@
...
      (let [window-hash (aget js/window "location" "hash")]
+       ;; if Terra redirects are globally enabled via config;
+       ;;   and the current hash has a path handler;
+       ;;   and that path handler has a Terra redirect,
+       ;; then redirect to Terra.
+       (when (config/terra-redirects-enabled)
+         (when-let [handler (nav/find-path-handler window-hash)]
+           (let [{:keys [make-props terra-redirect]} handler]
+             (when terra-redirect
+               (js-invoke (aget js/window "location") "replace" (str (config/terra-url) "/?fcredir=1#" (terra-redirect (make-props))))))))
```
breilly2: Could we also potentially use `fcredir` as a switch to show a "Welcome to FireCloud powered by Terra" message in Terra?

davidangb: sure! sounds good. That would also be good place in Terra UI to *strip* the `?fcredir=1` from the url so it isn't persistent across "pages".

breilly2 (changes requested):
A couple of things.
1. This is working for in-app navigation but not for direct links, such as bookmarks or opening links in a new tab/window (e.g. http://local.broadinstitute.org/#workspaces/broad-dsde-dev/real_bams)
2. The view starts to render in FC before the redirect happens. Is there any way to interrupt the render flow when we know we're going to redirect to avoid that while the browser is connecting to Terra?

src/cljs/main/broadfcui/main.cljs
breilly2: Agreed!

davidangb:
@breilly2 :
1. direct/bookmark links should  now be working.  The problem was that in `component-will-mount`, we hadn't yet loaded values from `config.json`, so the check to see if redirects are enabled would always fail.  Navigating through the app works because once you've loaded any component in the app, config was loaded, and we could rely on the previously-loaded values.
2. related to the previous bullet ... without a large refactor, we have to load at least one component before redirects will work, so *something* is going to render. How important do you think this is? (the other solution is to eliminate the config check to see if redirects are enabled; I'd  prefer to keep that)

breilly2:
The flicker alone probably isn't a big deal. Combined with the extra API calls that will happen while the view is rendering might be enough to tip in favor of doing this. Those API calls would show up in dashboards as FireCloud usage as well as put unnecessary extra load on the backend services.

I'm thinking a simple redirect component:
```
(react/defc- Redirect
  {:render
   (fn []
     [:div {} "Redirecting..."])})
```
Then extract a `terra-redirect?` boolean and include this somewhere in the `cond` further down:
```
terra-redirect?
[Redirect]
```

davidangb:
Since the redirect code is already in the top-level `App` component, it means refactoring that component entirely  - I'll take a shot and see what kind of effort/risk is involved.

breilly2 (approved):
👍 Simple and effective. Very nice!

PR #1522, opened 2019-04-02
Suppress rendering page when redirecting

breilly2:


davidangb (changes requested):
thanks for this - no longer necessary!

PR #1523, opened 2019-04-03
example test for terra redirects [risk: nope]

davidangb:
REVIEWERS PLEASE READ:

This PR sets up a test example for the contextual FireCloud->Terra redirects (see #1521 for code and DataBiosphere/terra-ui#1346 for requirements). In this PR:
* the test is ignored
* the redirect is not enabled

If you unignore the test without enabling the redirect, the test will fail.
If you enable the redirect, approximately 65 other tests will fail.

There is a big effort elsewhere to migrate those failing tests, and we don't want to enable the redirect itself (in prod) until May 1. The purpose of this PR is to set up the test framework so I can  keep implementing redirects in a local/dev env and have confidence in my work.

See code comments that explain how to enable the test or redirect, if you feel like doing so. You can  also look at the test results for the commit right before the final commit in this PR (convenience link: https://fc-jenkins.dsp-techops.broadinstitute.org/job/firecloud-ui-fiab-test-runner/13029/) , which shows that the new `RedirectSpec` passes, along with all the other expected failures.

Finally, there are two `TODO` comments in this PR on which I would really like reviewers' opinions - please read them while reviewing!


automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+  // negative test - ensure that Library/Methods Repo/whatever does NOT redirect
+  "A user who visits FireCloud should NOT be redirected to the same location in Terra for" - {
+
+    "methods repo" in {
+      implicit val authToken = user.makeAuthToken()
+      withWebDriver { implicit driver =>
+        withSignIn(user) { _ =>
+          val methodRepoPage = new MethodRepoPage().open
+          // give the page a chance to issue a redirect, if a redirect mistakenly existed on this page
+          Thread.sleep(1000)
```
aweng98, line 48: use a longer sleep to be certain redirect is not happening

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+    (url.contains("saturn") && url.contains("appspot.com"))
+  }
+
+  // negative test - ensure that Library/Methods Repo/whatever does NOT redirect
+  "A user who visits FireCloud should NOT be redirected to the same location in Terra for" - {
+
+    "methods repo" in {
+      implicit val authToken = user.makeAuthToken()
+      withWebDriver { implicit driver =>
+        withSignIn(user) { _ =>
```
aweng98, line 45: does auto redirect only happen if opened page is a workspace (like Summary tab) page? if not, auto-redirect could happen on listPage before action of opening methodRepoPage after signed in. there is a short delay after sign in.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
davidangb: why do you think redirect would take longer than this?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
davidangb: true ... but then `MethodRepoPage().open` would fail, and the test would fail, which is the desired result

aweng98:
I'm not familiar with this. when does the auto-redirect happen?

davidangb:
@aweng98 apologies, I should have linked to the relevant  PR: #1521

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    }
+
+  }
+
+
+  "A user who visits FireCloud should be redirected to the same location in Terra for" - {
+
+    "workspace summary page" ignore {
```
aweng98, line 60: it's ignored. but is this a negative test? ScalaTest provides two methods works with expected exception: assertThrows and intercept

davidangb: no, it's not a negative test ... it's testing that the redirect happens.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
```diff
@@ -0,0 +1,86 @@
...
+      implicit val authToken = user.makeAuthToken()
+      withCleanBillingProject(user) { billingProject =>
+        withWorkspace(billingProject, "Terra_redirect_spec") { workspaceName =>
+          withWebDriver { implicit driver =>
+            // TODO: withSignIn will throw an error at the end of its closure, when it tries to sign out the user.
```
breilly2, line 65: I'm pretty sure the main purpose of signing-out from `withSignIn` (other than the very few tests that need to sign in multiple times) was a shot-in-the-dark attempt to help us from being recognized as a robot by Google. If we want to eliminate the error from trying to sign out, we could extract a `signIn` function from https://github.com/broadinstitute/firecloud-ui/blob/develop/automation/src/test/scala/org/broadinstitute/dsde/firecloud/fixture/UserFixtures.scala#L73 and use that instead of `withSignIn`.

davidangb: agreed on extracting the function - I was just hoping to avoid doing that :(

ndbolliger (approved):
Looks good.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
aweng98: I thought speed of network could affect re-directs

davidangb: speed of network could certainly affect how fast the *content* of the Terra UI loads. However, this test is checking `driver.getCurrentUrl`, which is under the control of FCUI's JavaScript. We know that the test has already loaded FCUI due to `withSignIn`, etc. So what this test is really checking is "did FCUI JavaScript tell the browser to go to a Terra URL?" and not "did the Terra UI load". That's why I chose one second - that's a long time in terms of JavaScript execution. How does that sit with you?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
breilly2: What's the consequence of not extracting a function? Does the test fail because of this or is there just an ugly error logged somewhere?

aweng98: this may helps `try page.signOut() catch nonFatalAndLog(s"ERROR logging out user: ${user.email}")`

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
aweng98: sorry for I'm asking stupid questions. I'm not clear on what should happens when:
1. user log in FC
2. on workspace list page, user doesn't immediately click on any tab.
3. scroll page up/down. Does auto-redirect happen?
4. user in Terra, user wants to go Method repo page. how does user get back to Method page?

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
aweng98: I see now clearly. I'm good with it

aweng98:
this comment is not about this PR. generally speaking about redirect, if it happen on some pages but not all pages, it can cause confusion which may lead to bad experience. In my experience, page redirect happens on LogIn page, not after log in.

davidangb:
@aweng98 DataBiosphere/terra-ui#1346 is the relevant requirements. In short:
* if a user hits the FC homepage, user will be redirected to Terra UI homepage.
* Because Terra UI hasn't implemented everything yet, it still relies on FCUI for Methods Repo and Data  Library. Those pages in FC *must not*  redirect. This means we cannot have a global redirect.
* if a user clicks a link or bookmark to a specific workspace, method config, or other subsection of  FCUI, we will redirect to the equivalent location in Terra UI. This means we cannot blindly send all FC traffic to Terra UI homepage; we have to be smart and contextual.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
davidangb: the test doesn't fail, as evidenced by it passing in the second-to-last commit. The test log does contain a stack trace - that's the only consequence I can see - I'm tossing it out here to ask if anyone else can predict another consequence, or if folks object strongly to the additional stacktrace. (for context, remember that these tests are being sunset along with the UI and we'll be paying less attention to them)

aweng98 (approved):
Thank you for taking time to explain how it should works.

automation/src/test/scala/org/broadinstitute/dsde/firecloud/test/terraredirect/RedirectSpec.scala
davidangb: @aweng98 see my other comment for explanation of requirements. Certainly no redirect happens on *scroll*.

I don't understand your fourth question. If the user is in methods repo and clicks on a method, user will go to that method. Is that what you're asking?

breilly2 (approved):
👍

PR #1546, opened 2019-06-04
[SATURN-176] handle arbitrary return destinations

zarsky-broad:
twinned with https://github.com/broadinstitute/firecloud-develop/pull/1672

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off


src/cljs/main/broadfcui/main.cljs
```diff
@@ -200,7 +200,7 @@
            ;; are redirects both enabled and the current page has a redirect?
            terra-redirect? (and terra-redirects-enabled? terra-redirect)
            ;; to what url should we redirect, if redirects are enabled?
-           terra-redirect-url (when terra-redirect (make-terra-redirect-url terra-redirect make-props))
+           terra-redirect-url (when terra-redirect? (make-terra-redirect-url terra-redirect make-props))
```
davidangb: this is a functional change, and affects the url shown in the banner, even when redirects are not enabled. Did you mean to change this intentionally?

To test: set your preference to disable redirects. Navigate inside FC to some page that would have redirected you. Is the URL in the top banner correct?

src/cljs/main/broadfcui/common.cljs
```diff
@@ -246,11 +246,11 @@
...
+(def return-app (js* "new URLSearchParams(document.location.search).get('return')"))
 
-(defn has-firecloud-return? []
-  (string/includes? js/window.location.search "return=firecloud"))
+(defn make-return-url [app]
+  (string/replace (config/terra-base-url) "*" app))
 
-(defn has-return? []
-  (or (has-terra-return?) (has-firecloud-return?)))
+(defn get-return-url [] (make-return-url (if (= return-app "terra") "app" return-app)))
```
davidangb: two comments:

* if this is ever called outside of a `(when common/has-return? ...)` clause, it will return something like "null.terra.bio", right? Should probably handle the `null` case,  something like:
```
(defn get-return-url [] (make-return-url (if (or (nil? return-app) (= return-app "terra")) "app" return-app)))
```
* More fundamentally, this function encodes the business logic of how we use hostnames - the "terra"="app"  logic is hardcoded. I don't know how many more of these we plan to release, but how about something a little more future-flexible, logic that does:
  1.  if `@config` contains a value for `(str return-app "TerraUrl")`, use the value from config
  2. else, replace `*` in the terra-base-url with the value of `return-app`

This way, we can handle the business-logic exceptions by adding config values, instead of changing this code.

src/cljs/main/broadfcui/main.cljs
zarsky-broad: Ah, good catch. I missed that usage, I was trying to chase down a loading error from the value being loaded from config before config was loaded. I'll figure out an alternate solution.

src/cljs/main/broadfcui/common.cljs
zarsky-broad: I think, by design, base terra is very much the only exception. On the terra-ui side, it's coded as: `export const returnParam = () => getAppName().toLowerCase()`, and for anything else, we're intentionally keeping the name and the url the same. If anything, I'd rather codify that pairing further.

davidangb (approved):
:+1: looks fine; I did not test.

I'm still a little worried about potential for non-standard new "return" domains but if they arrive, we'll just have to deal with them later.

zarsky-broad:
UI tests passed: https://fc-jenkins.dsp-techops.broadinstitute.org/view/Testing/view/Pipelines/job/swatomation-pipeline/20639/

PR #1550, opened 2019-06-27
filter library project from url [SATURN-881]

zarsky-broad:
This allows any number of project names to be specified in the query string in the url when visiting the data library. Upon loading the page the filters are applied when the data loads, overriding any that were persisted, ~~and then removed from the url~~. If there are any other query parameters, they are preserved.

  * To Demo flag is set
  * Add notes on how to QA
  * Authentication
  * Authorization
  * Encryption
  * Audit trails
* Review cycle:
  * LR reviews
- [ ] **TL** sign off
- [ ] **LR** sign off

### DIFF
--- a/src/cljs/main/broadfcui/page/library/library_page.cljs
+++ b/src/cljs/main/broadfcui/page/library/library_page.cljs
@@ -387,10 +387,10 @@
     :component-did-mount
     (fn [{:keys [state]}]
       (let [js-search-params (js* "new URLSearchParams(document.location.search)")
-            search-projects (.get js-search-params "projects") ; have to preserve these if we have to mutate the object later
-            has-projects? (.has js-search-params "projects")]
+            search-projects (js->clj (.getAll js-search-params "project")) ; have to preserve these if we have to mutate the object later
+            has-projects? (.has js-search-params "project")]
         (when has-projects?
-          (.delete js-search-params "projects")
+          (.delete js-search-params "project")
           (let [projectless-query (.toString js-search-params)]
             (js/window.history.replaceState
              nil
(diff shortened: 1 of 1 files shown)


davidangb (changes requested):
this relies on a comma-delimited list of projects in a single query param:
```
https://portal.firecloud.org/?projects=TCGA,TOPMed#library
```
While technically feasible, we will run into nasty complexities if a project name itself contains a comma  -  and in general, working with delimited strings is less clean than working with e.g. arrays.

I suggest instead moving to multiple query params:
```
https://portal.firecloud.org/?projects=TCGA&projects=TOPMed#library
```
(rename "projects" to "project" as desired)
and then use `URLSearchParams.getAll` instead of and then use `URLSearchParams.get`, and loop through the array instead of splitting on a comma.

davidangb (approved):
:+1:

this *does* allow a user to specify a project that doesn't exist, e.g. http://local.broadinstitute.org/?project=nonexistent+L337+H4x0r#library

that's a bigger fix and raises UX questions about what we *should* do in this feature ... I'm fine merging as-is.

zarsky-broad:
@davidangb Yeah, for the time being I'm calling that out of scope, pending product.

zarsky-broad:
Ok @davidangb, last time. Talked with product; to avoid clobbering a user's saved filter, we won't remove the query string from the URL. When it's present, we won't persist filters, either.

davidangb (approved):
:+1: I'm fine with this if PO is. There's still some weird UX - which admittedly is tricky to resolve - for instance: 
* nav to a specific project, i.e. /?project=TCGA&return=terra#library
* manually uncheck TCGA in the left nav
* reload page; TCGA is now re-selected

but again, I'm ok with this

PR #1576, opened 2020-06-11
remove new relic instrumentation [no ticket; risk: no]

davidangb:
This removes New Relic instrumentation from firecloud-ui. It does not delete the NR JavaScript stub; that will happen in broadinstitute/firecloud-develop#2162 at some point after this one merges.

Terra UI and SCP will need parallel PRs.


breilly2 (approved):
👍 Code looks right. I was not able to test locally. There are several other references to newrelic in the code that we could also remove, but this appears to be the critical piece to stop using it in the app.

PR #1577, opened 2020-06-11
skip tests: remove newrelic.js from more places [no ticket;  risk: no]

davidangb:
#1576 was incomplete; this removes newrelic.js from more places.

broadinstitute/firecloud-develop#2162 removed newrelic.js from config, so this PR is necessary to prevent override.sh from looking for that file.


zarsky-broad (approved):
Missed 2: they're referenced further down in both `.conf` files in the `src/docker` folder; we'll also need to make sure the docker rsync script is tweaked to remove the reference to the js file when the fc-dev pr happens. Looks good pending those removals.

PR #1592, opened 2021-04-15
Integration test to check if Cromwell is uploading archived metadata files to GCS successfully [BW-572]

salonishah11:
Script output from Jenkins job runs can be found [here](https://github.com/broadinstitute/dsp-jenkins/pull/518#issue-616338181).

Closes https://broadworkbench.atlassian.net/browse/BW-572


automation/alpha-archived-metadata-file-upload-check.sh
```diff
@@ -0,0 +1,31 @@
...
+VAULT_TOKEN=$(cat /etc/vault-token-dsde)
+
+# Get a workflow which completed at least 40 minutes ago and whose metadata is archived
+# (archive-delay in alpha is set to 30 minutes. Hence adding 10 more minutes should give Cromwell enough time to
+# archive such a workflow)
+ARCHIVED_WORKFLOW_RESULT=$(
+  docker run --rm \
+     -e VAULT_TOKEN=$VAULT_TOKEN \
+     broadinstitute/dsde-toolbox mysql-connect.sh -p firecloud -a cromwell1 -e alpha \
+     "SELECT WORKFLOW_EXECUTION_UUID, ROOT_WORKFLOW_EXECUTION_UUID FROM cromwell.WORKFLOW_METADATA_SUMMARY_ENTRY WHERE METADATA_ARCHIVE_STATUS = 'Archived' AND END_TIMESTAMP <= DATE_SUB(NOW(), INTERVAL 40 minute) ORDER BY END_TIMESTAMP DESC LIMIT 1;"\
```
aednichols: Stylistic only, the `-` operator is overloaded to work as one expects
```suggestion
     "SELECT WORKFLOW_EXECUTION_UUID, ROOT_WORKFLOW_EXECUTION_UUID FROM cromwell.WORKFLOW_METADATA_SUMMARY_ENTRY WHERE METADATA_ARCHIVE_STATUS = 'Archived' AND END_TIMESTAMP <= ( NOW() - INTERVAL 40 minute ) ORDER BY END_TIMESTAMP DESC LIMIT 1;"\
```

aednichols (approved):

automation/alpha-archived-metadata-file-upload-check.sh
```diff
@@ -0,0 +1,31 @@
...
+if [[ $ROOT_WORKFLOW_ID == 'NULL' ]]
+then
+  FILE_LOCATION=${WORKFLOW_ID}/${WORKFLOW_ID}.csv
+else
+  FILE_LOCATION=${ROOT_WORKFLOW_ID}/${WORKFLOW_ID}.csv
+fi
+
+echo "Checking if CSV file exists for workflow ${WORKFLOW_ID} in bucket 'cromwell-carbonited-workflows-alpha'..."
+
+# If file exists, 'gsutil stat' will print it's metadata, else it returns 'No URLs matched' and exits with code 1.
```
cjllanwarne:
```suggestion
# If file exists, 'gsutil stat' will print its metadata, else it returns 'No URLs matched' and exits with code 1.
```

automation/alpha-archived-metadata-file-upload-check.sh
```diff
@@ -0,0 +1,31 @@
...
+VAULT_TOKEN=$(cat /etc/vault-token-dsde)
+
+# Get a workflow which completed at least 40 minutes ago and whose metadata is archived
+# (archive-delay in alpha is set to 30 minutes. Hence adding 10 more minutes should give Cromwell enough time to
+# archive such a workflow)
+ARCHIVED_WORKFLOW_RESULT=$(
+  docker run --rm \
+     -e VAULT_TOKEN=$VAULT_TOKEN \
+     broadinstitute/dsde-toolbox mysql-connect.sh -p firecloud -a cromwell1 -e alpha \
+     "SELECT WORKFLOW_EXECUTION_UUID, ROOT_WORKFLOW_EXECUTION_UUID FROM cromwell.WORKFLOW_METADATA_SUMMARY_ENTRY WHERE METADATA_ARCHIVE_STATUS = 'Archived' AND END_TIMESTAMP <= (NOW() - INTERVAL 40 minute) ORDER BY END_TIMESTAMP DESC LIMIT 1;"\
```
cjllanwarne, line 16: Once the archiver catches up, it might be a good idea to check the time stamp is actually **recent** enough (eg within the last day) rather than **old** enough. Eg if the archiver stalls, this test will keep passing because it will just keep finding the most recent successfully archived entry

automation/alpha-archived-metadata-file-upload-check.sh
```diff
@@ -0,0 +1,31 @@
...
+     | tail -n1
+)
+
+WORKFLOW_ID=$(echo $ARCHIVED_WORKFLOW_RESULT | cut -d' ' -f1)
+ROOT_WORKFLOW_ID=$(echo $ARCHIVED_WORKFLOW_RESULT | cut -d' ' -f2)
```
cjllanwarne, line 21: Is it worth having a second instance of this test which deliberately picks an archived subworkflow to guarantee that it was put into its root workflow's directory?

automation/alpha-archived-metadata-file-upload-check.sh
salonishah11: That makes sense. I can create a ticket to change this query logic to search for workflows whose `END_TIMESTAMP <= (NOW() - INTERVAL 40 minute) AND END_TIMESTAMP > (NOW() - INTERVAL 1 DAY)` once the archiver has caught up and after we change the `archive-delay` to 30 minutes. How does that sound?

automation/alpha-archived-metadata-file-upload-check.sh
salonishah11: I think there's no harm in having additional check for that. The only thing I am concerned about is whether we have workflows run in alpha on daily basis that have subworkflows (I say daily because of above comment where we eventually want to get a workflow id that was run within the last day). The 35k scatter workflow doesn't have any subworkflow. I will check if there are any such workflows run in alpha at certain intervals, and if they are, I can add additional check for the subworkflows.

salonishah11: So I checked in alpha db, and it looks like we don't have any Jenkins tests that have workflows with subworkflows in it. If you think we should check that archived subworkflow is put in its root workflow's directory, then we would have to add a Jenkins test that runs a workflow with subworkflow in it.

automation/alpha-archived-metadata-file-upload-check.sh
cjllanwarne: Sounds good to me

automation/alpha-archived-metadata-file-upload-check.sh
cjllanwarne: Maybe we should write this up as a separate ticket and decide in a refinement whether it's worth doing or not? Then for now, it can not a blocker to completing this ticket.

cjllanwarne (approved):

automation/alpha-archived-metadata-file-upload-check.sh
salonishah11: Created https://broadworkbench.atlassian.net/browse/BW-658

automation/alpha-archived-metadata-file-upload-check.sh
salonishah11: Created https://broadworkbench.atlassian.net/browse/BW-657

salonishah11:
@cjllanwarne I requested a re-review since I made few small changes to the script after you reviewed it. The main change is addition of [this](https://github.com/broadinstitute/firecloud-ui/blob/b1b54336bf70c8fca61426055e764bdb35a9d406/automation/alpha-archived-metadata-file-upload-check.sh#L23-L29) condition that exits with code 1 if no workflow id was found.

cjllanwarne (approved):

PR #1594, opened 2021-05-19
De-duplicate ACCESS_TOKEN generation and log submission details responses [BW-687]

cjllanwarne:
Makes two improvements to the Cromwell CI scripts: 

- Added resilience should result in not having to re-fetch tokens every time one is needed. 
- When we fetch submission details, we now long what came back from curl so that we can debug it later if necessary.

Currently manually testing as: https://fcprod-jenkins.dsp-techops.broadinstitute.org/job/Complex-Workflow-testV2-against-prod/6601/

### DIFF
--- a/automation/prod-workflow-inc.sh
+++ b/automation/prod-workflow-inc.sh
@@ -24,6 +24,23 @@ checkToken () {
     fi
 }
 
+getAccessToken() {
+  user=$1
+
+  [ "${ACCESS_TOKEN_USER}" = "${user}" -a -n "${ACCESS_TOKEN}" ] || ACCESS_TOKEN=$(
+    docker \
+        run \
+        --rm \
+        -v "${WORKING_DIR}:/app/populate" \
+        -w /app/populate \
+        broadinstitute/dsp-toolbox \
+        python get_bearer_token.py "${user}" "${JSON_CREDS}"
+  )
+
+  export ACCESS_TOKEN
+  export ACCESS_TOKEN_USER="${user}"
+}
+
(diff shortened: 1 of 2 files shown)


automation/prod-workflow-inc.sh
```diff
@@ -108,33 +125,18 @@ monitorSubmission() {
...
-    workflowsStatus=$(
-        curl \
+    submissionDetails=$(curl \
             -X GET \
             --header 'Accept: application/json' \
             --header "Authorization: Bearer ${ACCESS_TOKEN}" \
-            "https://api.firecloud.org/api/workspaces/$namespace/$name/submissions/$submissionId" \
-        | jq -r '.workflows[] | .status'
-    )
+            "https://api.firecloud.org/api/workspaces/$namespace/$name/submissions/$submissionId")
```
salonishah11, line 134: Not sure if this is already being done somewhere, but do we log the response of this curl? It would be helpful to log it in case we get an error apart from 401 from Rawls

aednichols (approved):

automation/prod-workflow-inc.sh
cjllanwarne: I originally had it explicitly `echo`ing these values out, but the bash script starts with `-x` which turns on debug logging for every line anyway. So it IS printed out, just not explicitly on every line in the script

salonishah11: Ah I see. Thank you for explaining 👍

salonishah11 (approved):

PR #52, opened 2015-07-28
added button and modal for UI for importing method configurations

eddiebroad:
this time, all UI modifications/changes lie in the context of the "Method Configurations" tab.

PR #1596, opened 2021-06-03
Fix 'ACCESS_TOKEN_USER: unbound variable' error [BW-723]

salonishah11:
Apart from fixing this error, this PR also 

- adds few println statements to make debugging easier  
- for `monitorSubmission()` in `automation/submission-perf-inc.sh`, it reverts the code to retrieve submissionStatus, workflowsStatus and workflowFailures as the reverted code works. Otherwise during the test runs I was seeing error `jq: error (at <stdin>:8): Cannot iterate over null (null)`

### DIFF
--- a/automation/prod-workflow-inc.sh
+++ b/automation/prod-workflow-inc.sh
@@ -10,7 +10,7 @@ checkToken () {
 
     # Get access token if it hasn't been initialized. This usually happens during the start of a test.
     # For more details see https://broadworkbench.atlassian.net/browse/BW-721
-    if [ -z "${ACCESS_TOKEN}" ]
+    if [ -z "${ACCESS_TOKEN-}" ]
     then
       getAccessToken "$user"
     fi
@@ -33,7 +33,7 @@ checkToken () {
 getAccessToken() {
   user=$1
 
-  if [ -z "${ACCESS_TOKEN}" ]
+  if [ -z "${ACCESS_TOKEN-}" ]
   then
     NEED_TOKEN=true
   else
--- a/automation/prod-workflow-inc.sh
+++ b/automation/prod-workflow-inc.sh
@@ -33,7 +33,7 @@ checkToken () {
 getAccessToken() {
   user=$1
 
-  if [ "${ACCESS_TOKEN_USER}" = "${user}" -a -n "${ACCESS_TOKEN}" ]
+  if [ -n "${ACCESS_TOKEN}" ]
   then
     checkToken "$user"
   else
@@ -54,7 +54,6 @@ getAccessToken() {
   fi
 
   export ACCESS_TOKEN
-  export ACCESS_TOKEN_USER="${user}"
   export NEED_TOKEN=false
 }
 
--- a/automation/prod-workflow-inc.sh
+++ b/automation/prod-workflow-inc.sh
@@ -44,7 +44,6 @@ getAccessToken() {
 
   if [ "${NEED_TOKEN}" = "true" ]
   then
-    printf "\nRetrieving new ACCESS_TOKEN for user '%s'" "${user}"
     ACCESS_TOKEN=$(
       docker \
         run \
--- a/automation/prod-workflow-inc.sh
+++ b/automation/prod-workflow-inc.sh
@@ -33,11 +33,11 @@ checkToken () {
 getAccessToken() {
   user=$1
 
-  if [ -z "${ACCESS_TOKEN-}" ]
+  if [ "${ACCESS_TOKEN_USER-}" = "${user}" -a -n "${ACCESS_TOKEN-}" ]
   then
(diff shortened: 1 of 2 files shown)


automation/prod-workflow-inc.sh
```diff
@@ -33,7 +33,7 @@ checkToken () {
 getAccessToken() {
   user=$1
 
-  if [ "${ACCESS_TOKEN_USER}" = "${user}" -a -n "${ACCESS_TOKEN}" ]
+  if [ "${ACCESS_TOKEN_USER-}" = "${user}" -a -n "${ACCESS_TOKEN-}" ]
   then
```
salonishah11, line 39: TODO: Add a comment stating the purpose of this check

aednichols (approved):

automation/prod-workflow-inc.sh
```diff
@@ -42,6 +44,7 @@ getAccessToken() {
 
   if [ "${NEED_TOKEN}" = "true" ]
   then
+    printf "\nRetrieving new ACCESS_TOKEN for user '%s'" "${user}"
```
cjllanwarne: in `automation/prod-workflow-inc.sh`-based scripts this is all already printed due to having `set -x` set (ie "debug printing"). Might be another alternative to set the same option in `automation/submission-perf-inc.sh`?

automation/submission-perf-inc.sh
```diff
@@ -244,16 +255,33 @@ monitorSubmission() {
...
-        -X GET \
-        --header 'Accept: application/json' \
-        --header "Authorization: Bearer ${ACCESS_TOKEN}" \
-        "https://api.firecloud.org/api/workspaces/$namespace/$name/submissions/$submissionId")
+    printf "\nFetching status for submission ID '%s':" "${submissionId}"
 
-    submissionStatus=$(jq -r '.status' <<< "${submissionDetails}" )
-    workflowsStatus=$(jq -r '.workflows[] | .status' <<< "${submissionDetails}")
-    workflowFailures=$(jq -r '[.workflows[] | select(.status == "Failed")] | length' <<< "${submissionDetails}")
+    # curl -s -X GET --header 'Accept: application/json' --header "Authorization: Bearer $ACCESS_TOKEN" "https://firecloud-orchestration.dsde-alpha.broadinstitute.org/api/submissions/queueStatus" | jq -r '"\(now),\(.workflowCountsByStatus.Queued),\(.workflowCountsByStatus.Running),\(.workflowCountsByStatus.Submitted)"' | tee -a workflow-progress-$BUILD_NUMBER.csv
```
cjllanwarne: Since it's unused (and has no context), now might be a good time to just remove this comment?

automation/submission-perf-inc.sh
```diff
@@ -244,16 +255,33 @@ monitorSubmission() {
...
+    submissionStatus=$(
+        curl \
+            -X GET \
+            --header 'Accept: application/json' \
+            --header "Authorization: Bearer ${ACCESS_TOKEN}" \
+            "https://firecloud-orchestration.dsde-alpha.broadinstitute.org/api/workspaces/${namespace}/${name}/submissions/${submissionId}" \
+        | jq -r '.status'
+    )
+    workflowsStatus=$(
```
cjllanwarne, line 268: This seems like a backwards step? Instead of fetching this value once and then interpreting it locally, we're now fetching it multiple times?

automation/prod-workflow-inc.sh
salonishah11: Maybe add `set -x` in `submission-perf-inc.sh` as well? But I am a little hesitant considering it is mainly used in perf tests and not by us. I can then remove this println from `prod-workflow-inc.sh` but keep it in `submission-perf-inc.sh`. How does that sound?

automation/submission-perf-inc.sh
salonishah11: Yes. I am not sure what is going on, but getting details once and then interpreting it locally was causing the tests to fail with `jq: error (at <stdin>:8): Cannot iterate over null (null)` ([#52](https://fc-jenkins.dsp-techops.broadinstitute.org/job/alpha-impossible-to-read-archived-metadata-from-within-workflow-check/52/console), [#1199](https://fc-jenkins.dsp-techops.broadinstitute.org/job/PerformanceTest-against-Alpha/1199/console)). Not sure why it was getting workflows[] as empty.  But reverting it back to this worked.

cjllanwarne: Interesting! 👍  for now, and maybe I'll test locally to see what's going wrong!

automation/prod-workflow-inc.sh
cjllanwarne: Sounds good to me

cjllanwarne (approved):

PR #1603, opened 2021-08-03
30m token timeout and more logging in 35k scatter test [BW-762]

cjllanwarne:
Note: Saloni mentioned problems in the following tests by switching to a "fetch submission details once" approach" the last time, but the benefits of "fetch once" are huge:

- The details are logged, so it's easier to debug what's going on
- The failures often happen on the second or third attempt to access the APIs, perhaps due to race conditions when the liveness check being several seconds or minutes behind the actual request.

Therefore, I am bringing it back, but:
- With a slightly different implementation
- In only `submission-perf-inc.sh`
- With manual testing of https://fc-jenkins.dsp-techops.broadinstitute.org/job/PerformanceTest-against-Alpha/ and https://fc-jenkins.dsp-techops.broadinstitute.org/job/alpha-impossible-to-read-archived-metadata-from-within-workflow-check/ against the branch as part of the testing for this change.

### DIFF
--- a/automation/submission-perf-inc.sh
+++ b/automation/submission-perf-inc.sh
@@ -26,8 +26,11 @@ checkToken () {
     # Verify that user is authorized to make the API call and does not need to refresh their token
     if [[ "${tokenStatus}" =~ '"requiresRefresh":true' ]] || [[ ${tokenStatus} =~ "401 Unauthorized" ]]
     then
-      NEED_TOKEN=true
-      export NEED_TOKEN
+        export NEED_TOKEN=true
+    else if [ "$(( $(date +%s) - ${TOKEN_CREATION_TIME-0} ))" -gt "1800" ]
+        export NEED_TOKEN=true
+    else
+        export NEED_TOKEN=false
     fi
 }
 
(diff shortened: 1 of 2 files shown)


automation/submission-perf-inc.sh
```diff
@@ -36,15 +41,15 @@ getAccessToken() {
...
+  if [ "${ACCESS_TOKEN_USER-}" = "${user}" ] && [ -n "${ACCESS_TOKEN-}" ]
   then
     checkToken "$user"
   else
     NEED_TOKEN=true
   fi
 
-  if [ "${NEED_TOKEN}" = "true" ]
-  then
+  while [ "${NEED_TOKEN}" = "true" ]
```
cjllanwarne: Switch back to if/then

aednichols:
It looks like there is a pending change to if/then so will hold off reviewing until those commits pop up

cjllanwarne:
@aednichols now with the while=>if change made, should be good to review

aednichols (approved):

cahrens (approved):

PR #1626, opened 2023-06-09
WX-1021 Improve Performance Test Reliability

THWiseman:
This branch is intended to make our Jenkins "Performance Test Against Alpha" less flaky. It allows the `curl` commands to have a few retries, and also echos the output of the bigger one to the console for easier debugging in the future.

### DIFF
--- a/automation/submission-perf-inc.sh
+++ b/automation/submission-perf-inc.sh
@@ -202,8 +202,9 @@ findSubmissionID() {
             --header 'Accept: application/json' \
             --header "Authorization: Bearer ${ACCESS_TOKEN}" \
             "https://firecloud-orchestration.dsde-alpha.broadinstitute.org/api/workspaces/$namespace/$name/submissions" \
+        | tee submissionIDResponse.json \
         | jq -r "[.[] | select($selectorString)] | sort_by(.submissionDate) | reverse[0] | .submissionId")
-
+    cat submissionIDResponse.json
     export submissionID
 }
 
--- a/automation/submission-perf-inc.sh
+++ b/automation/submission-perf-inc.sh
@@ -168,7 +168,7 @@ launchSubmission() {
     echo "Response code is:"
     echo "$response"
     response_code=$(echo "$response" | tr -d '\n')
-    echo("response: ${response}")
+    echo "response: ${response}"
 }
 
 findSubmissionID() {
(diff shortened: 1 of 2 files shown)


automation/submission-perf-inc.sh
```diff
@@ -143,7 +147,13 @@ launchSubmission() {
         optionalEntityNameField="\"entityName\":\"${entityName}\","
     fi
 
-    curl \
+    response=$(curl \
```
THWiseman, line 150: This curl command is the most liable to fail, from what I've seen. It's also the only one submitting a non-trivial amount of data. Giving it some bonus retry attempt/time

automation/submission-perf-inc.sh
```diff
@@ -184,12 +196,15 @@ findSubmissionID() {
...
+            --retry 3 \
+            --retry-max-time 20 \
             -X GET \
             --header 'Accept: application/json' \
             --header "Authorization: Bearer ${ACCESS_TOKEN}" \
             "https://firecloud-orchestration.dsde-alpha.broadinstitute.org/api/workspaces/$namespace/$name/submissions" \
+        | tee submissionIDResponse.json \
         | jq -r "[.[] | select($selectorString)] | sort_by(.submissionDate) | reverse[0] | .submissionId")
-
+    cat submissionIDResponse.json
```
THWiseman, line 207: When this test has flaked/failed in the past, this HTTP request was the culprit, but the error was passed into `jq` and lost. This change should dump the failed response into the logs so we have a lead to chase when troubleshooting.

automation/submission-perf-inc.sh
```diff
@@ -143,7 +148,13 @@ launchSubmission() {
         optionalEntityNameField="\"entityName\":\"${entityName}\","
     fi
 
-    curl \
+    launchSubmissionResponse=$(curl \
+        -s \
+        -o /dev/null \
+        -w "%{http_code}" \
+        --connect-timeout 60 \
+        --retry 8 \
```
kraefrei: Have we found we actually want 8 retries here? Likely isn't going to be a huge contributor to slowing what is already pretty slow CI, but seems high to me.

jgainerdewar (approved):
I agree with Christian that 8 retries is high, other than that LGTM!

aednichols (approved):

